PR #42, opened 2022-10-05
Add `microceph init`

stgraber:


microceph/cmd/microceph/init.go
```diff
@@ -24,5 +33,146 @@ func (c *cmdInit) Command() *cobra.Command {
...
+	}
+
+	lc, err := m.LocalClient()
+	if err != nil {
+		return err
+	}
+
+	// Check if already initialized.
+	_, err = lc.GetClusterMembers(context.Background())
+	isUninitialized := err != nil && err.Error() == "Daemon not yet initialized"
```
tomponline, line 49: Can we add a sentinel error value for uninitialized cluster?

stgraber: This is a bit of a workaround for now. The goal should be to have a proper boolean API value we can query instead.

PR #54, opened 2022-11-15
Add a unit test for ceph bootstrapping

sabaini:
Perform some refactoring to be able to use mocks

Signed-off-by: Peter Sabaini <peter.sabaini@canonical.com>

PR #58, opened 2022-11-24
Create a unit test for microceph bootstrapping

sabaini:
Need to perform some refactoring to be able to use mocks. This has been split out from PR #54

Signed-off-by: Peter Sabaini peter.sabaini@canonical.com


microceph/ceph/bootstrap.go
```diff
@@ -5,17 +5,17 @@ import (
...
 	"database/sql"
 	"fmt"
+	"github.com/canonical/microceph/microceph/common"
 	"os"
 	"path/filepath"
 
 	"github.com/canonical/microceph/microceph/database"
 
-	"github.com/canonical/microcluster/state"
 	"github.com/pborman/uuid"
```
stgraber, line 14: Hmm, we've been using the LXD import convention thus far.

Basically, splitting the imports into up to 3 groups.

```
import (
    "built-in/packages"

    "external packages (other git repo)"

    "internal packages (same git repo)"
)
```

So in this case it should be something like:
```
import (
	"context"
	"database/sql"
	"fmt"
	"os"
	"path/filepath"

	"github.com/pborman/uuid"

	"github.com/canonical/microceph/microceph/common"
```

(I realize that this particular file was already not fully compliant)

stgraber:
On top of the import issue, this still needs a lot more splitting I suspect.

Looking at this, I see a few different type of changes in here:
 - Addition and move from struct to interface as argument to various functions (`common.StateInterface`)
 - Extracting some code into shared functions
 - Test code itself

Ideally, I'd have each newly split out function be its own commit which both introduces the new function and updates the existing code to use it. Then a commit to introduce the new interfaces, then another commit to port the code to the new interfaces.
Lastly, a commit to put the tests in place and wrap things up with a commit for the gomod update.

sabaini:
Closing this one in favor of splitting out PRs

PR #69, opened 2022-12-13
Fix OSD startup script liveness check

sabaini:
In addition to checking for admin socket presence, also check there is a process listening at the other end in order to prevent false positives in checking ceph-osd startup due to a stale socket file

Signed-off-by: Peter Sabaini <peter.sabaini@canonical.com>


stgraber:
Oh, good catch!

PR #77, opened 2023-01-19
Add encryption support for OSDs

### LINKED ISSUE #34: Add support for LUKS encryption
It would be great to have a story for encryption at rest.

Ideally, we'd want the ability to provide a single passphrase or key file to then unlock all the OSDs.
OSDs probably still ought to be individually encrypted with individual keys.

I'd suggest using one small LUKS volume that contains key files for each of the OSDs.
After system reboot, the user would need to do `microceph unlock` which would prompt for the passphrase (or key file) used to unlock that small LUKS volume. At that point, the OSD startup script will be able to read the keys for each of the OSDs from that volume and unlock them individually.

This makes it easy to change the key without having to alter each of the OSDs.

A `microceph lock` command could be added which would unmount and close the small LUKS volume.
Doing so would prevent any access to the individual OSD keys, which would in turn prevent the addition of additional OSDs but also prevent crashed OSD daemons from respawning (so a bit of a double edged sword).

ChrisMacNaughton:
OSDs can optionally be encrypted via LUKS and dm-crypt on initialization

Closes #34

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -38,6 +38,7 @@ jobs:
           sudo snap install --dangerous microceph_*.snap
           sudo snap connect microceph:block-devices
           sudo snap connect microceph:hardware-observe
+          sudo snap connect microceph:dm-crypt
 
           sudo microceph cluster bootstrap
 
@@ -62,7 +63,7 @@ jobs:
             # names (/dev/sdiY) that are not used inside GitHub Action runners
             minor="${loop_dev##/dev/loop}"
             sudo mknod -m 0660 "/dev/sdi${l}" b 7 "${minor}"
-            sudo microceph disk add --wipe "/dev/sdi${l}"
+            sudo microceph disk add --wipe "/dev/sdi${l}" --encrypt
           done
 
           # Wait for OSDs to become up
--- a/microceph/ceph/osd.go
+++ b/microceph/ceph/osd.go
@@ -2,9 +2,12 @@ package ceph
 
 import (
 	"context"
+	"crypto/rand"
 	"database/sql"
+	"encoding/base64"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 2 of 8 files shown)


microceph/ceph/osd.go
```diff
@@ -68,12 +71,141 @@ func nextOSD(s *state.State) (int64, error) {
...
 
+// setupEncryptedOSD sets up an encrypted OSD on the given disk.
+//
+// Takes a path to the disk device as well as the osd data path and the osd id.
+// Returns the path to the encrypted device and an error if any.
+func setupEncryptedOSD(devicePath string, osdDataPath string, osdID int64) (string, error) {
+	if err := os.Symlink(devicePath, filepath.Join(osdDataPath, "unencrypted")); err != nil {
+		return "", fmt.Errorf("Failed to add unencrypted block symlink: %w", err)
+	}
+	// Create a key for the encrypted device
```
stgraber, line 82: Code style nit, put you'll want a line break before that comment.
Same for the chunks below.

.github/workflows/tests.yml
```diff
@@ -80,6 +81,10 @@ jobs:
           sudo microceph.ceph health | grep -xF "HEALTH_OK"
           pgrep ceph-osd || { echo "No ceph-osd process found" ; exit 1; }
 
+      - name: consider debugging
+        uses: lhotari/action-upterm@v1
+        if: failure()
+
```
stgraber, line 87: Do we want to always have that?
In theory that may cause some Github Actions to fail and hit the block for the shell when the user isn't paying attention, causing for long running Github Actions which could end up being costly depending on what plan Canonical is running on these days.

snapcraft.yaml
```diff
@@ -34,6 +34,7 @@ apps:
     daemon: simple
     plugs:
       - block-devices
+      - dm-crypt
```
stgraber, line 37: Is that a new interface to snapd?

If so, we'll probably want to introduce a `assumes` section so we get snapd to fail ahead of time with an error that's more user friendly.

snapcraft/commands/osd.start
```diff
@@ -3,6 +3,26 @@ export SNAP_CURRENT="$(realpath "${SNAP_DATA}/..")/current"
 echo $$ > "${SNAP_CURRENT}/run/ceph-osd.pid"
 cd "${SNAP}"
 
+
+maybe_unlock() {
```
stgraber, line 7: Is `shellcheck` happy with this change?

I somehow would expect it to be complaining at least about the lack of quoting around `$dev` and `$luksname` here.

snapcraft.yaml
ChrisMacNaughton: Frustratingly, it exists (not sure since when) but isn't complete pending a snapd release

stgraber: Okay, so will probably want to still do an `assumes` for the release where it actually works.
Will also need to file an auto-connect request for that interface as I don't expect the store to grant it automatically.

ChrisMacNaughton: the code for the encryption should actually handle the case where the interface is present but not functional, it just doesn't allow encryption then

stgraber: Ah, okay, that's good.

ChrisMacNaughton: specifically, this bit: https://github.com/canonical/microceph/pull/77/files#diff-3418d4d1a1b4a82b23ce7f4b04ec10f2fff558f4e61ef25e2f4f93834d33738cR192

snapcraft/commands/osd.start
sabaini: Its not complaining as the make target actually doesn't check it :-] 
I've added the quoting you mentioned, making a note for myself to get the shell scripts checked as well

PR #84, opened 2023-02-08
Enable basic RadosGW backend

sabaini:
Backend part of the RGW service

Signed-off-by: Peter Sabaini <peter.sabaini@canonical.com>

PR #85, opened 2023-02-08
Add RGW frontend and test changes

sabaini:
Hook up client frontend, stacked on top of PR #84 

Signed-off-by: Peter Sabaini <peter.sabaini@canonical.com>


ChrisMacNaughton:
I'd be super excited if this PR included new functional test coverage of writing to an object via RGW and reading it back out!

sabaini:
> I'd be super excited if this PR included new functional test coverage of writing to an object via RGW and reading it back out!

Good idea! I've added some functesting however as I need radosgw-admin (for setting up auth) I've created a separate PR for this -- PR #87

.github/workflows/tests.yml
```diff
@@ -68,11 +75,25 @@ jobs:
           # Wait for OSDs to become up
           sleep 30
 
+      - name: Enable RGW
+        run: |
+          set -eux
+          sudo microceph enable rgw
+          # Wait for RGW to settle
+          sleep 30
```
lmlg, line 83: I'm not a big fan of these sleeps to assert that a service is up. Would it be too hard to loop until the RGW has settled (up or down) ?

sabaini: I just tried to stay consistent with the existing tests :-D 
But yea I kinda feel the same way, I'll try to do some retry loop

microceph/api/endpoints.go
```diff
@@ -10,4 +10,5 @@ var Endpoints = []rest.Endpoint{
 	disksCmd,
 	resourcesCmd,
 	servicesCmd,
+	enableRGWCmd,
```
stgraber, line 13: That seems odd. I was expecting enabling services to be done through an update to the existing services endpoint.

Right now, we only have `GET /1.0/services` which gives us a pretty basic list of services and their location.

I think if I was to do this change, I'd instead change the services API to offer something like:
 - GET /1.0/services (same as we do today probably?)
 - GET /1.0/services/SERVICE (where service is osd, mon, rgw, mgr, mds) which provides us with service state and a list of locations serving it
 - PUT /1.0/services/SERVICE to alter the configuration of a service, letting you enable it and indicate where it should be enabled (setting the location)

.github/workflows/tests.yml
```diff
@@ -68,11 +75,34 @@ jobs:
...
           # Show ceph's status
           sudo microceph.ceph status
 
-          # Confirm ceph is healthy
+          # Confirm ceph is healthy and services started
           sudo microceph.ceph health | grep -xF "HEALTH_OK"
+          sudo microceph.ceph status | grep -F "mon: 1 daemons"
+          sudo microceph.ceph status | grep -E "mgr: .*active, "
+          sudo microceph.ceph status | grep -F "osd: 3 osds"
+          sudo microceph.ceph status | grep -F "rgw: 1 daemon"
```
stgraber, line 105: Would be even better if you could use s3cmd to push and pull a file to S3, making sure the whole thing works as expected.

stgraber: Nevermind, just saw the PR that adds exactly that, all good!

microceph/api/endpoints.go
sabaini: Thanks @stgraber , this is valuable feedback.

I have now revamped the API, adding a `PUT /1.0/services/rgw` endpoint which takes a flag to enable/disable a service. I would like to defer choosing a location for the service, as well as listing services and locations, to a separate PR if this works for you.

stgraber:
The git history is a bit of a mess in this PR, could you do a rebase to get rid of those merge commits?

sabaini:
Hey @stgraber I've rebased onto rgw-backend, hope this is what you had in mind?

stgraber:
Not quite, your PR shows it includes 6 commits which is incorrect, it only includes 2.
You need to rebase against the main repo's main branch to get that down to just the two commits you're actually adding.

sabaini:
Thx @stgraber hope this works better now

.github/workflows/tests.yml
stgraber, line 105: Can you split the changes to this file into a separate commit (same PR is fine)?

microceph/ceph/rgw.go
```diff
@@ -49,14 +49,45 @@ func EnableRGW(s common.StateInterface, port int) error {
...
 }
 
-// rgwUpdateDatabase records changes to the database.
-func rgwUpdateDatabase(s common.StateInterface) error {
+// DisableRGW disables the RGW service on the cluster.
+func DisableRGW(s common.StateInterface) error {
+	confPath := filepath.Join(os.Getenv("SNAP_DATA"), "conf")
+	dataPath := filepath.Join(os.Getenv("SNAP_COMMON"), "data")
+
+	if err := stopRGW(); err != nil {
```
stgraber, line 57: Another code style thing here which you may or may not want to continue, up to you.

In LXD's codebase, we don't allow the define+check if syntax, instead requiring the function call and assignment be done on the line prior to the if statement. The reason for that is mostly to avoid accidental, issue with variable scoping and to make it a bit easier when cherry-picking fixes and logic changes.

A lot of Go projects do use the more compact syntax just fine though, so it's definitely up to you, if you want to keep using it, let me know and I'll stop flagging it in reviews :)

microceph/cmd/microceph/enable_rgw.go
```diff
@@ -0,0 +1,52 @@
...
+
+type cmdEnableRGW struct {
+	common   *CmdControl
+	flagPort int
+}
+
+func (c *cmdEnableRGW) Command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "rgw [--port <port>]",
+		Short: "Enable the RGW service on this node",
```
stgraber, line 21: `node` is a term which hasn't been used in this codebase yet, maybe you should use `server` instead?

Can you also make it so `enable` and `disable` can be used through `UseTarget` with a `--target` flag?

So running `microceph enable rgw --target foo` will call it with `UseTarget("foo")` which will then cause microceph to forward the request to the correct server. This enables much easier management of the cluster by not having to SSH into the right system.

PR #95, opened 2023-02-23
Record the new OSD id before formatting

### LINKED ISSUE #94: UNIQUE constraint failed: disks.osd
While adding several OSDs on two cluster members concurrently I got these errors:

```
Error: Failed adding new disk: Failed to record disk: Failed to create "disks" entry: UNIQUE constraint failed: disks.osd
Error: Failed adding new disk: Failed to record disk: Failed to create "disks" entry: UNIQUE constraint failed: disks.osd
```

Think we run into races starting here where we're [computing an OSD id](https://github.com/canonical/microceph/blob/main/microceph/ceph/osd.go#L122) but only [much later record it](https://github.com/canonical/microceph/blob/main/microceph/ceph/osd.go#L170)

As the whole OSD formatting is between these two steps there's ample time for nodes to race each other. 

Probably could fix this by doing the id computation and INSERT in a single transaction

ChrisMacNaughton:
When we defer recording an OSD ID until after
the OSD is prepared, there is a long race that
another OSD can get setup with a duplicate ID.

Closes #94

Signed-off-by: Chris MacNaughton <chris.macnaughton@canonical.com>

### DIFF
--- a/microceph/ceph/osd.go
+++ b/microceph/ceph/osd.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/canonical/microcluster/state"
 	"github.com/lxc/lxd/lxd/resources"
+	"github.com/lxc/lxd/lxd/revert"
 	"github.com/lxc/lxd/shared"
 	"github.com/pborman/uuid"
 
@@ -70,6 +71,9 @@ func nextOSD(s *state.State) (int64, error) {
 
 // AddOSD adds an OSD to the cluster, given a device path and a flag for wiping
 func AddOSD(s *state.State, path string, wipe bool) error {
+	revert := revert.New()
+	defer revert.Fail()
+
 	// Validate the path.
 	if !shared.IsBlockdevPath(path) {
 		return fmt.Errorf("Invalid disk path: %s", path)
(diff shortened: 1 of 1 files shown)


microceph/ceph/osd.go
```diff
@@ -124,24 +124,48 @@ func AddOSD(s *state.State, path string, wipe bool) error {
...
+		})
+	}
+
 	// Create directory.
 	dataPath := filepath.Join(os.Getenv("SNAP_COMMON"), "data")
 	osdDataPath := filepath.Join(dataPath, "osd", fmt.Sprintf("ceph-%d", nr))
 
 	err = os.MkdirAll(osdDataPath, 0700)
 	if err != nil {
+		cleanup()
```
lmlg, line 154: why not `defer`ing the cleanup callback? This way is a bit too error prone imo.

ChrisMacNaughton: yeah, I had the same thought; except defer is the thing that always runs at the end, AIUI

ChrisMacNaughton: there does also come a point where different cleanup is needed, rather than just removing the entry from our database

lmlg: > yeah, I had the same thought; except defer is the thing that always runs at the end, AIUI

From my limited Go experience, you typically set a flag that the callback reads before performing cleanup (i.e: "we've reached the happy path, no need to clean the mess").

lmlg: > there does also come a point where different cleanup is needed, rather than just removing the entry from our database

yeah, this is harder and probably requires splitting this function into smaller ones. I don't think it's worth it for this PR fwiw, just thought of a suggestion :)

ChrisMacNaughton: > > yeah, I had the same thought; except defer is the thing that always runs at the end, AIUI
> 
> From my limited Go experience, you typically set a flag that the callback reads before performing cleanup (i.e: "we've reached the happy path, no need to clean the mess").

I can probably get away with `if err == nil` as the check

stgraber:
@ChrisMacNaughton you may want to look at https://github.com/lxc/lxd/tree/master/lxd/revert

It's what we developed for LXD and it allows easily piling up reversion function calls as you go through a function, then at the end of a successful function run, you just tell it that all is fine and nothing needs to be reverted. If you return for any other reason, then the defer kicks in and reverts everything.

It also allows effectively merging reverters together so you can have a function that performs some action on its own and gives you back a reverter for inclusion in the caller.

stgraber:
Its tests show the basic usage https://github.com/lxc/lxd/blob/master/lxd/revert/revert_test.go

microceph/ceph/osd.go
tomponline: We use the revert package in lxd to get this conditional behavior that you may find beneficial. 

See https://github.com/lxc/lxd/blob/e1ed3d148f03a0ad32b8975d8dcef76515e4d896/lxd/revert/revert_test.go#L9

tomponline: This also allows to return the reverters Fail function as a revert.Hook to allow a function to return its cleanup steps to the calling function, which can sometimes be useful in more complex call chains.

tomponline: Sorry didn't see @stgraber had said the same thing earlier :)

lmlg (approved):

PR #100, opened 2023-03-15
Use port 7443 for init to bring inline with bootstrap/join

iherdgoats:
I'm guessing this was missed during refactoring.


stgraber:
Hello,

Contributions to projects in the Canonical org require signing the Canonical Licensing Agreement (CLA) which is at https://ubuntu.com/legal/contributors/agreement

Your commits also need to be signed-off using `git commit --amend -s`.

Once both those things are done we should see the CLA and DCO tests pass which will then let us review the content of your pull request.

Thanks!

iherdgoats:
CLA should now be signed.

PR #101, opened 2023-03-15
Update install docs with upgrade management section

lmlg:
This PR adds a section in the installation docs that specifies the recommended way to manage upgrades.

### DIFF
--- a/docs/tutorial/install.rst
+++ b/docs/tutorial/install.rst
@@ -38,15 +38,3 @@ Install Microceph
     for i in $(seq 1 3); do
         lxc exec microceph-$i -- sh -c 'snap install microceph'
     done
-
-Update management
-~~~~~~~~~~~~~~~~~
-
-Updates are not managed from within microceph itself. As such, users are the ones who will manage them. Thus, automatic snap updates should be postponed indefinitely.
-This can be achieved with the following command (Supported with snapd from version 2.58 onwards):
-
-.. code-block:: shell
-
-    for i in $(seq 1 3); do
-        lxc exec microceph-$i -- sh -c 'snap refresh --hold microceph'
-    done
--- a/docs/tutorial/install.rst
+++ b/docs/tutorial/install.rst
@@ -42,7 +42,7 @@ Install Microceph
 Update management
 ~~~~~~~~~~~~~~~~~
 
-For the time being, it's recommended that updates be postponed indefinitely.
+Updates are not managed from within microceph itself. As such, users are the ones who will manage them. Thus, automatic snap updates should be postponed indefinitely.
 This can be achieved with the following command (Supported with snapd from version 2.58 onwards):
 
 .. code-block:: shell
--- a/docs/tutorial/install.rst
+++ b/docs/tutorial/install.rst
@@ -36,5 +36,5 @@ Install Microceph
 .. code-block:: shell
 
     for i in $(seq 1 3); do
-        lxc exec microceph-$i -- sh -c 'snap install microceph'
+        lxc exec microceph-$i -- sh -c 'snap install microceph && snap refresh --hold microceph'
     done
--- a/docs/tutorial/install.rst
+++ b/docs/tutorial/install.rst
@@ -38,3 +38,15 @@ Install Microceph
     for i in $(seq 1 3); do
         lxc exec microceph-$i -- sh -c 'snap install microceph'
     done
+
+Update management
+~~~~~~~~~~~~~~~~~
+
+For the time being, it's recommended that updates be postponed indefinitely.
+This can be achieved with the following command (Supported with snapd from version 2.58 onwards):
+
+.. code-block:: shell
+
+    for i in $(seq 1 3); do
+        lxc exec microceph-$i -- sh -c 'snap refresh --hold microceph'
+    done
--- a/docs/tutorial/update.rst
+++ b/docs/tutorial/update.rst
@@ -1,7 +1,8 @@
 Update management
 ~~~~~~~~~~~~~~~~~
 
-Microceph will not manages its own updates - users are required to do so. In order to set up microceph to manage updates correctly, the following steps need to be taken.
+Updates in Microceph are managed by the usual snap mechanisms. However, because snaps automatically deploy updates, this can be problematic in Microceph clusters that need an ordered rollout of updates.
+As such, we strongly advise that automatic updates be disabled for Microceph. Documentation about snaps updates can be consulted `here <https://snapcraft.io/docs/keeping-snaps-up-to-date>`_.
 
 Disable automatic updates
 ~~~~~~~~~~~~~~~~~~~~~~~~~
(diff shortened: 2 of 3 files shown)


docs/tutorial/install.rst
```diff
@@ -38,3 +38,15 @@ Install Microceph
     for i in $(seq 1 3); do
         lxc exec microceph-$i -- sh -c 'snap install microceph'
     done
+
+Update management
+~~~~~~~~~~~~~~~~~
+
+For the time being, it's recommended that updates be postponed indefinitely.
```
ChrisMacNaughton, line 45: This isn't a 'for the time being' thing, it's a 'this is how we recommend you install and manage the lifecycle of microceph. Fundamentally, we're never going to be able to manage microceph updates from within microceph so we have to recommend the correct workflows for our users to manage it themselves and that includes holding the version, as well as how to update when they're ready!

ChrisMacNaughton (changes requested):

docs/tutorial/install.rst
```diff
@@ -38,3 +38,15 @@ Install Microceph
...
+Update management
+~~~~~~~~~~~~~~~~~
+
+Updates are not managed from within microceph itself. As such, users are the ones who will manage them. Thus, automatic snap updates should be postponed indefinitely.
+This can be achieved with the following command (Supported with snapd from version 2.58 onwards):
+
+.. code-block:: shell
+
+    for i in $(seq 1 3); do
+        lxc exec microceph-$i -- sh -c 'snap refresh --hold microceph'
```
wolsen, line 51: This command is specific to running micro ceph in a lxd container. You should simply state that the snap refresh --hold command is the correct command. The example can reference to running within a lxd container, but a baremetal machine wouldn't use the lxc part of the command.

wolsen (changes requested):

docs/tutorial/update.rst
```diff
@@ -0,0 +1,54 @@
...
+        pools:   1 pools, 1 pgs
+        objects: 2 objects, 577 KiB
+        usage:   65 MiB used, 45 GiB / 45 GiB avail
+        pgs:     1 active+clean
+
+Within a cluster healthy, we need to run the following command on each node:
+
+.. code-block:: shell
+
+    snap install microceph
```
ChrisMacNaughton, line 51: I believe that this will do nothing?

I suspect we want something like:

```
snap refresh microceph
snap refresh --hold microceph
```

ChrisMacNaughton (changes requested):

docs/tutorial/update.rst
```diff
@@ -0,0 +1,64 @@
+Update management
+~~~~~~~~~~~~~~~~~
+
+Microceph will not manages its own updates - users are required to do so. In order to set up microceph to manage updates correctly, the following steps need to be taken.
```
sabaini, line 4: I feel like users who are not familiar with snaps would benefit from some context around snap updates. Also, I don't think we should make it an imperative -- a single node MicroCeph cluster should be fine with auto-upgrades (if the upgrade works at all). Maybe something like:

----

MicroCeph updates are managed by the snap mechanisms. By default snaps will automatically deploy updates, this however can be problematic for MicroCeph clusters due to ordering of and concurrent rollout of updates. It's therefore strongly advised to disable automatic updates for MicroCeph snaps. For details on the snap update mechanism look [here](https://snapcraft.io/docs/keeping-snaps-up-to-date)

----

docs/tutorial/update.rst
```diff
@@ -0,0 +1,64 @@
...
+
+Disable automatic updates
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+On each node of the cluster where microceph has been installed, we must run the following:
+
+.. code-block:: shell
+
+    snap refresh --hold microceph
```
sabaini, line 13: Will need sudo
`sudo snap refresh --hold microceph`

docs/tutorial/update.rst
```diff
@@ -0,0 +1,64 @@
...
+        objects: 2 objects, 577 KiB
+        usage:   65 MiB used, 45 GiB / 45 GiB avail
+        pgs:     1 active+clean
+
+With a healthy cluster, we need to run the following command on each node:
+
+.. code-block:: shell
+
+    snap refresh microceph
+    snap refresh --hold microceph
```
sabaini, line 52: Will need sudo

docs/tutorial/update.rst
```diff
@@ -0,0 +1,64 @@
...
+
+1. Managers
+2. Monitors
+3. All other entities (i.e: OSDs)
+
+The output of the 'microceph.ceph status' command should provide us with the hostnames of the mons and managers ('microceph-1' et al in this example).
+
+At present time, managers and monitors reside on the same nodes, but that may not always be the case.
+
+With that done, we should verify the cluster together to make it's settled and healthy once again.
```
sabaini, line 64: Maybe add, for version verification:

```
.. code-block:: shell
    snap list microceph
```


Maybe this again for health verification
```
.. code-block:: shell
    microceph.ceph status
```

sabaini (changes requested):
Hey, thanks for the instructions, added some comments inline

docs/tutorial/update.rst
```diff
@@ -0,0 +1,73 @@
+Update management
+~~~~~~~~~~~~~~~~~
+
+Updates in Microceph are managed by the usual snap mechanisms. However, because snaps automatically deploy updates, this can be problematic in Microceph clusters that need an ordered rollout of updates.
+As such, we strongly advise that automatic updates be disabled for Microceph. Documentation about snaps updates can be consulted `here <https://snapcraft.io/docs/keeping-snaps-up-to-date>`_.
```
sabaini, line 5: s/snaps updates/snap updates/

docs/tutorial/update.rst
```diff
@@ -0,0 +1,73 @@
...
+        pgs:     1 active+clean
+
+With a healthy cluster, we need to run the following command on each node:
+
+.. code-block:: shell
+
+    sudo snap refresh microceph
+    sudo snap refresh --hold microceph
+
+The order on which we run the commands is important. It should be as follows:
```
sabaini, line 55: s/order on/order in/

docs/tutorial/update.rst
```diff
@@ -0,0 +1,73 @@
...
+
+1. Managers
```
sabaini, line 57: Just for extra foolproof-ness maybe:

1. Managers (mgr)
2. Monitors (mon)
3. All other entities (OSDs, rgw, etc.)

sabaini (changes requested):
Looking good! 
Noted two typos and a very minor suggestion

ChrisMacNaughton:
One thing that I don't see in this proposal is that we should include the `snap refresh --hold microceph` in the install documentation as well, or people will only discover this outage=inducing behaviour at upgrade time (or outage time!), rather than at install time!

sabaini:
One of the checks shows a warning -- update.rst is not being referenced by any TOC. Which is a fair point I guess, we'll want it to be included in the index probably?

PR #102, opened 2023-03-20
Enable Format Flags for List

kyzrfranz:
Since automation is one of the things I'm currently working on, I realised that the `list` commands don't make use of the formatting. I realise that's due to the fact that the output combines two sets of data essentially, that's not quite straightforward to put into one set.

So the suggestion would be to:
- enable the format flags on `list`
- introduce a new flag `-a`| `--available` that shows only locally available, unconfigured disks
- make listing microceph configured devices the default behavior

hope this is feasible.


microceph/cmd/microceph/cluster_list.go
```diff
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"github.com/lxc/lxd/shared/i18n"
```
wolsen, line 5: The order of the imports should be revisited. The convention for the repository is:

```
built-in

external imports

internal imports (github.com/canonical/microceph/XYZ in this case)
```

wolsen (changes requested):
Thanks for the PR @kyzrfranz! This PR feels like it should be multiple commits as there's at least 2 distinct changes in here (list available disks, add format flags).

In addition it'd be great to see some test coverage for new functionality.

kyzrfranz:
So I went ahead and did as mentioned before. 
I am not very much in love with how the asserts work on the table tbh but that's due to the fact that I would like to propose some changes to  `utils.RenderTable` first. This should really set up differently tbh only returning the model. You can run some decent assertions on that one, without digging into os.stdout and parsing strings...
So pending that discussion I went for a "works for now" (hopefully) approach.

Let me know how that works for you - I added tests for all commands I can stick the newly proposed `ApiClient` in.

PR #105, opened 2023-03-21
Enable Format Flags for List

kyzrfranz:
Sorry for the confusion - had to resolve some conflicts and did something weird, so #102 got closed :D
Changes as described in https://github.com/canonical/microceph/pull/102

So I went ahead and did as mentioned before.
I am not very much in love with how the asserts work on the table tbh but that's due to the fact that I would like to propose some changes to utils.RenderTable first. This should really set up differently tbh only returning the model. You can run some decent assertions on that one, without digging into os.stdout and parsing strings...
So pending that discussion I went for a "works for now" (hopefully) approach.

Let me know how that works for you - I added tests for all commands I can stick the newly proposed ApiClient in.

### DIFF
--- a/microceph/client/client.go
+++ b/microceph/client/client.go
@@ -12,12 +12,56 @@ import (
 	"github.com/canonical/microceph/microceph/api/types"
 )
 
+type ApiDisksGetter interface {
+	GetDisks(ctx context.Context) (types.Disks, error)
+}
+
+type ApiResourcesGetter interface {
+	GetResources(ctx context.Context) (*api.ResourcesStorage, error)
+}
+
+type ApiServicesGetter interface {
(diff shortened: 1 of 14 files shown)


kyzrfranz:
Hope i got 'em all - realized that theres actually a `mocks`folder so put the apiclient mock there :)

sabaini (approved):

sabaini:
I'm +1 on this; @stgraber @ChrisMacNaughton would you be able to cast an eye?

UtkarshBhatthere:
@kyzrfranz Could you please take a look at this PR ? We need another rebase.

UtkarshBhatthere:
Closed due to long inactivity.

PR #106, opened 2023-03-24
Add OSD encryption

sabaini:
Add optional OSD disk encryption via LUKS and dm-crypt

### DIFF
--- a/microceph/ceph/osd.go
+++ b/microceph/ceph/osd.go
@@ -2,9 +2,12 @@ package ceph
 
 import (
 	"context"
+	"crypto/rand"
 	"database/sql"
+	"encoding/base64"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strconv"
 	"strings"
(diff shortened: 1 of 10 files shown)


sabaini:
Note this depends on https://github.com/snapcore/snapd/pull/12653 landing

microceph/ceph/osd.go
```diff
@@ -69,15 +72,155 @@ func nextOSD(s *state.State) (int64, error) {
...
+
+	// Check if we have the dm_crypt module
+	inf, err := os.Stat("/sys/module/dm_crypt")
+	if err != nil || inf == nil || !inf.IsDir() {
+		return fmt.Errorf("Missing dm_crypt module: %w", err)
+	}
+
+	// Check if we can list the /run directory; older snapd had an issue with this, https://github.com/snapcore/snapd/pull/12445
+	if _, err = os.ReadDir("/run"); err != nil {
+		return fmt.Errorf("Can't access /run, might need to update snapd to >2.58: %w", err)
```
ChrisMacNaughton, line 204: Looks like this should be updated to 2.59, matching the line above

ChrisMacNaughton (changes requested):
Generally approve, just one complaint!

microceph/ceph/osd.go
sabaini: Heh, the /run bug was really fixed earlier than the other one above (in 2.58.1 or so) so this line technically is correct. But guess this might be confusing and doesn't really help... I'll push a change

ChrisMacNaughton (approved):
LGTM, thanks for all of the back-and-forth needed to get this ready to use in both this project as well as more broadly in snapd!

PR #116, opened 2023-03-29
Add Cluster config management to microceph

UtkarshBhatthere:
Added command
```microceph cluster config set/get/reset/list```
for managing configs (currently limited to cluster_network and public_network)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -130,6 +130,36 @@ jobs:
           s3cmd --host localhost --host-bucket="localhost/%(bucket)" --access_key=fooAccessKey --secret_key=fooSecretKey --no-ssl put -P ~/test.txt s3://testbucket
           curl -s http://localhost/testbucket/test.txt | grep -F hello-radosgw
 
+      - name: Test Cluster Config
+        run: |
+          set -eux
+          cip=$(ip -4 -j route | jq -r '.[] | select(.dst | contains("default")) | .prefsrc' | tr -d '[:space:]')
+          
+          # pre config set timestamp for service age
+          ts=$(sudo systemctl show --property ActiveEnterTimestampMonotonic snap.microceph.osd.service | cut -d= -f2)
+
+          # set config
+          sudo microceph cluster config set cluster_network $cip/8 --wait
+
+          # post config set timestamp for service age
+          ts2=$(sudo systemctl show --property ActiveEnterTimestampMonotonic snap.microceph.osd.service | cut -d= -f2)
+
+          # Check config output
+          output=$(sudo microceph cluster config get cluster_network | grep -cim1 'cluster_network')
+          if [[ $output -lt 1 ]] ; then echo "config check failed: $output"; exit 1; fi
+
+          # Check service restarted
+          if [ $ts2 -lt $ts ]; then echo "config check failed: TS1: $ts2 TS2: $ts3"; exit 1; fi
+
+          # reset config
+          sudo microceph cluster config reset cluster_network --wait
+
+          # post config reset timestamp for service age
+          ts3=$(sudo systemctl show --property ActiveEnterTimestampMonotonic snap.microceph.osd.service | cut -d= -f2)
+
+          # Check service restarted
+          if [ $ts3 -lt $ts2 ]; then echo "config check failed: TS2: $ts2 TS3: $ts3"; exit 1; fi
+
       - name: Upload artifacts
         if: always()
         uses: actions/upload-artifact@v3
--- a/microceph/api/configs.go
+++ b/microceph/api/configs.go
@@ -23,17 +23,21 @@ var configsCmd = rest.Endpoint{
 }
 
 func cmdConfigsGet(s *state.State, r *http.Request) response.Response {
+	var err error
 	var req types.Config
+	var configs types.Configs
 
-	err := json.NewDecoder(r.Body).Decode(&req)
+	err = json.NewDecoder(r.Body).Decode(&req)
 	if err != nil {
 		return response.InternalError(err)
 	}
 
-	// Fetch configs.
-	configs, err := ceph.ListConfigs()
-	if err != nil {
-		return response.InternalError(err)
+	// If a valid key string is passed, fetch that key.
+	if len(req.Key) > 0 {
+		configs, err = ceph.GetConfigItem(req)
+	} else {
+		// Fetch all configs.
+		configs, err = ceph.ListConfigs()
 	}
 
 	return response.SyncResponse(true, configs)
@@ -55,12 +59,8 @@ func cmdConfigsPut(s *state.State, r *http.Request) response.Response {
 	}
 
 	services := configTable[req.Key].Daemons
-	go func() {
-		client.SendRestartRequestToClusterMembers(s, r, services)
-		// Restart Daemons on host.
-		ceph.RestartCephServices(services)
-	}()
-
+	client.ConfigChangeRefresh(s, services, req.Wait)
+	
 	return response.EmptySyncResponse
 }
 
@@ -80,11 +80,7 @@ func cmdConfigsDelete(s *state.State, r *http.Request) response.Response {
 	}
 
 	services := configTable[req.Key].Daemons
-	go func() {
-		client.SendRestartRequestToClusterMembers(s, r, services)
-		// Restart Daemons on host.
-		ceph.RestartCephServices(services)
-	}()
+	client.ConfigChangeRefresh(s, services, req.Wait)
 
 	return response.EmptySyncResponse
 }
--- a/microceph/api/configs.go
+++ b/microceph/api/configs.go
@@ -11,11 +11,7 @@ import (
 	"github.com/canonical/microceph/microceph/api/types"
 	"github.com/canonical/microceph/microceph/ceph"
 	"github.com/canonical/microceph/microceph/client"
-	"github.com/canonical/microceph/microceph/common"
 )
-
-// Config table for ConfigExtras
-var configTable = ceph.GetConfigTable()
 
 // /1.0/configs endpoint.
 var configsCmd = rest.Endpoint{
@@ -35,7 +31,7 @@ func cmdConfigsGet(s *state.State, r *http.Request) response.Response {
 	}
 
 	// Fetch configs.
-	configs, err := ceph.ListConfigs(common.CephState{State: s}, req.Key)
+	configs, err := ceph.ListConfigs()
 	if err != nil {
 		return response.InternalError(err)
 	}
@@ -45,62 +41,50 @@ func cmdConfigsGet(s *state.State, r *http.Request) response.Response {
 
 func cmdConfigsPut(s *state.State, r *http.Request) response.Response {
 	var req types.Config
+	configTable := ceph.GetConfigTable()
 
 	err := json.NewDecoder(r.Body).Decode(&req)
 	if err != nil {
 		return response.InternalError(err)
 	}
 
-	if !client.IsForwardedRequest(r) {
-		// Call set on the original Request but not on Forwarded Requests.
-		err = ceph.SetConfigItem(common.CephState{State: s}, req)
-		if err != nil {
-			return response.SmartError(err)
-		}
-
-		// Forward request to cluster members
-		err = client.ForwardConfigRequestToClusterMembers(s, r, &req, client.SetConfig)
+	// Configure the key/value
+	err = ceph.SetConfigItem(req)
+	if err != nil {
+		return response.SmartError(err)
 	}
 
-	// Restart Daemons on host.
-	daemons := configTable[req.Key].Daemons
-	for i := range daemons {
-		err = ceph.RestartCephService(daemons[i])
-		if err != nil {
-			return response.SmartError(err)
-		}
-	}
+	services := configTable[req.Key].Daemons
+	go func() {
+		client.SendRestartRequestToClusterMembers(s, r, services)
+		// Restart Daemons on host.
+		ceph.RestartCephServices(services)
+	}()
 
 	return response.EmptySyncResponse
 }
 
 func cmdConfigsDelete(s *state.State, r *http.Request) response.Response {
 	var req types.Config
+	configTable := ceph.GetConfigTable()
 
 	err := json.NewDecoder(r.Body).Decode(&req)
 	if err != nil {
 		return response.InternalError(err)
 	}
 
-	if !client.IsForwardedRequest(r) {
-		// Call set on the original Request but not on Forwarded Requests.
-		err = ceph.RemoveConfigItem(common.CephState{State: s}, req)
-		if err != nil {
-			return response.SmartError(err)
-		}
-
-		// Forward request to cluster members
-		err = client.ForwardConfigRequestToClusterMembers(s, r, &req, client.ClearConfig)
+	// Clean the key/value
+	err = ceph.RemoveConfigItem(req)
+	if err != nil {
+		return response.SmartError(err)
 	}
 
-	// Restart Daemons on host.
-	daemons := configTable[req.Key].Daemons
-	for i := range daemons {
-		err = ceph.RestartCephService(daemons[i])
-		if err != nil {
-			return response.SmartError(err)
-		}
-	}
+	services := configTable[req.Key].Daemons
+	go func() {
+		client.SendRestartRequestToClusterMembers(s, r, services)
+		// Restart Daemons on host.
+		ceph.RestartCephServices(services)
+	}()
 
 	return response.EmptySyncResponse
 }
--- a/microceph/api/configs.go
+++ b/microceph/api/configs.go
@@ -65,7 +65,7 @@ func cmdConfigsPut(s *state.State, r *http.Request) response.Response {
 	// Restart Daemons on host.
 	daemons := configTable[req.Key].Daemons
 	for i := range daemons {
-		err = ceph.RestartCephDaemon(daemons[i])
+		err = ceph.RestartCephService(daemons[i])
 		if err != nil {
 			return response.SmartError(err)
 		}
@@ -96,7 +96,7 @@ func cmdConfigsDelete(s *state.State, r *http.Request) response.Response {
 	// Restart Daemons on host.
 	daemons := configTable[req.Key].Daemons
 	for i := range daemons {
-		err = ceph.RestartCephDaemon(daemons[i])
+		err = ceph.RestartCephService(daemons[i])
 		if err != nil {
 			return response.SmartError(err)
 		}
--- a/microceph/api/configs.go
+++ b/microceph/api/configs.go
@@ -39,13 +39,16 @@ func cmdConfigsGet(s *state.State, r *http.Request) response.Response {
 		// Fetch all configs.
 		configs, err = ceph.ListConfigs()
 	}
+	if err != nil {
+		return response.SmartError(err)
+	}
 
 	return response.SyncResponse(true, configs)
 }
 
 func cmdConfigsPut(s *state.State, r *http.Request) response.Response {
 	var req types.Config
-	configTable := ceph.GetConfigTable()
+	configTable := ceph.GetConstConfigTable()
 
 	err := json.NewDecoder(r.Body).Decode(&req)
 	if err != nil {
@@ -66,7 +69,7 @@ func cmdConfigsPut(s *state.State, r *http.Request) response.Response {
 
 func cmdConfigsDelete(s *state.State, r *http.Request) response.Response {
 	var req types.Config
-	configTable := ceph.GetConfigTable()
+	configTable := ceph.GetConstConfigTable()
 
 	err := json.NewDecoder(r.Body).Decode(&req)
 	if err != nil {
--- a/microceph/api/configs.go
+++ b/microceph/api/configs.go
@@ -1,0 +1,106 @@
+package api
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/canonical/microcluster/rest"
+	"github.com/canonical/microcluster/state"
+	"github.com/lxc/lxd/lxd/response"
+
+	"github.com/canonical/microceph/microceph/api/types"
+	"github.com/canonical/microceph/microceph/ceph"
+	"github.com/canonical/microceph/microceph/client"
+	"github.com/canonical/microceph/microceph/common"
+)
+
+// Config table for ConfigExtras
+var configTable = ceph.GetConfigTable()
+
+// /1.0/configs endpoint.
+var configsCmd = rest.Endpoint{
+	Path: "configs",
+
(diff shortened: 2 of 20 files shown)

### CODE REVIEW

**Comment by stgraber** [comment]
Am I missing something or this currently doesn't actually cause any changes to the generated config?

I also just wanted to check, I'm assuming the reason why this is being done is because those two config keys can't simply be set through `microceph.ceph config set global KEY VALUE`? My understanding is that for 99% of config keys, we just want to configure them through `ceph config` these days instead of having to manually keep track of them and putting them into ceph.conf (which then requires restarting the daemons).

**Comment by UtkarshBhatthere** [comment]
> Am I missing something or this currently doesn't actually cause any changes to the generated config?

The first commit is around the command plane, there is another commit in pipe which deals with updating the config. 

> I also just wanted to check, I'm assuming the reason why this is being done is because those two config keys can't simply be set through `microceph.ceph config set global KEY VALUE`? My understanding is that for 99% of config keys, we just want to configure them through `ceph config` these days instead of having to manually keep track of them and putting them into ceph.conf (which then requires restarting the daemons).

It really does make more sense to use ceph config rather than managing a ceph.conf file but the motivation for microceph cluster config is more around providing a opinionated abstraction over the breadth of ceph config. The idea is to allow a limited list of frequently required configs (currently only global but can be extended to mon-only, client-only configs etc). The actual config update can take place through ceph config itself. This is also under discussion in a spec (CE0011).

**Comment by ChrisMacNaughton** [comment]
It is certainly possible to set `cluster_network` and `public_network` via (eg:) `ceph config set global cluster_network 192.168.1.0/24`; however, changing those configurations still requires restarting daemons to pick up new network bindings.

File: microceph/client/client.go [hunk 1]
```diff
@@ -50,6 +54,40 @@ func GetConfig(ctx context.Context, c *client.Client, data *types.Config) (types
 	return configs, nil
 }
 
+// IsForwardedRequest determines if this request has been forwarded from another cluster member.
+func IsForwardedRequest(r *http.Request) bool {
+	return r.Header.Get("User-Agent") == request.UserAgentNotifier
+}
+
+func ForwardConfigRequestToClusterMembers(s *state.State, r *http.Request, data *types.Config, handle func(ctx context.Context, c *client.Client, data *types.Config) error) (error) {
```
ChrisMacNaughton, line 62: I don't think we want to call this a config forwarding thing; it's more of a generic "please restart this kind of deamon" thing

microceph/client/client.go
```diff
@@ -50,6 +54,40 @@ func GetConfig(ctx context.Context, c *client.Client, data *types.Config) (types
...
+	// Get a collection of clients every other cluster member, with the notification user-agent set.
+	cluster, err := s.Cluster(r)
+	if err != nil {
+		return fmt.Errorf("Failed to get a client for every cluster member: %w", err)
+	}
+
+	err = cluster.Query(s.Context, true, func(ctx context.Context, c *client.Client) error {
+		addrPort, err := restTypes.ParseAddrPort(s.Address().URL.Host)
```
ChrisMacNaughton, line 70: Why are you parsing this port? it looks like it's only used for logging?

microceph/ceph/config.go
```diff
@@ -3,15 +3,144 @@ package ceph
...
+
+// Struct to get Config Items from config dump json output.
+type ConfigDump []struct{
+	Section string
+	Name    string
+	Value   string
+}
+
+func RestartCephDaemon(daemon string) error {
+	return snapReload(daemon)
```
ChrisMacNaughton, line 60: As is, I don't think that just wrapping `snapReload` is that useful; however, this might be a good place to restart and then wait for the daemon to be fully functional again

microceph/ceph/config.go
```diff
@@ -3,15 +3,144 @@ package ceph
...
+}
+
+func GetConfigItem(s common.StateInterface, c types.Config) (types.Config, error) {
+	var err error
+	args := []string{
+		"config",
+		"get",
+		configTable[c.Key].Who,
+		c.Key,
+	}
```
ChrisMacNaughton, line 87: the other calls of `ceph config` in here seem to be using the `json-pretty` format

ChrisMacNaughton (changes requested):
The commits should also be cleaned up so that we're not adding a migration and then removing that migration

microceph/ceph/config.go
```diff
@@ -3,15 +3,140 @@ package ceph
...
+	"github.com/canonical/microceph/microceph/api/types"
 	"github.com/canonical/microceph/microceph/common"
 	"github.com/canonical/microceph/microceph/database"
 )
 
+type ConfigExtras struct{
+	Who     string // Ceph Config internal <who> against each key
+	Daemons []string // Daemons that need to be restarted post config change.
+	Regexp  string // Regular Expression to check value against key
+	Cb      *func(s common.StateInterface, c types.Config)
```
sabaini, line 21: I believe neither Regexp nor Cb are currently in use, right? Suggest removing them if so

microceph/ceph/config.go
```diff
@@ -3,15 +3,140 @@ package ceph
...
 	"os"
 	"path/filepath"
 	"strings"
 
+	"github.com/canonical/microceph/microceph/api/types"
 	"github.com/canonical/microceph/microceph/common"
 	"github.com/canonical/microceph/microceph/database"
 )
 
+type ConfigExtras struct{
```
sabaini, line 17: Some docs on this type would be good, same for ConfigTable below.

I feel like the name of these two type could be clarified as well. Afaict the purpose of this type is to find services to restart when configs are set, so maybe `ServiceConfigs` and  `ServiceRestartMap`  or such? Possibly the types could be unified too

microceph/ceph/config.go
```diff
@@ -3,15 +3,140 @@ package ceph
...
+
+type ConfigTable map[string]ConfigExtras
+
+// Check if certain key is present in the map.
+func (c ConfigTable) isKeyPresent(key string) bool {
+	if _, ok := c[key]; !ok {
+        return false
+    }
+
+	return true
```
sabaini, line 32: Needs go formatting

microceph/ceph/config.go
```diff
@@ -3,15 +3,140 @@ package ceph
...
+
+// Return keys of the given set
+func (c ConfigTable) Keys() (keys []string) {
+    for k := range c {
+        keys = append(keys, k)
+    }
+    return keys
+}
+
+func GetConfigTable() ConfigTable {
```
sabaini, line 43: By convention go constructors are named like `NewFoo()`, suggest to follow this here

microceph/ceph/config.go
```diff
@@ -3,15 +3,140 @@ package ceph
...
+	return ConfigTable{
+		"public_network":  {"global", []string{"mon", "osd"}, "", nil},
+		"cluster_network": {"global", []string{"osd"}, "", nil},
+	}
+}
+// Instantiation of config table.
+var configTable = GetConfigTable()
```
sabaini, line 50: If possible I'd suggest trying to avoid global vars and instantiate in the funcs where needed

microceph/ceph/services.go
```diff
@@ -5,12 +5,105 @@ import (
...
+	"github.com/Rican7/retry/backoff"
+	"github.com/Rican7/retry/strategy"
 	"github.com/canonical/microcluster/state"
 
 	"github.com/canonical/microceph/microceph/api/types"
 	"github.com/canonical/microceph/microceph/database"
+	"github.com/tidwall/gjson"
 )
 
+type Set map[string]int
```
sabaini, line 18: Nit: this is often done like the below because we don't actually need an `int` here, we only need to know that a key is present, and `struct{}` is smaller

```
type Set map[string]struct{}
...
var exists = struct{}{}
mySet["foo"] = exists
```

microceph/ceph/services.go
```diff
@@ -5,12 +5,105 @@ import (
...
+	var retval Set
+	output, err := processExec.RunCommand("ceph", "mon", "dump", "-f", "json-pretty")
+	if err != nil {
+		return nil, err
+	}
+
+	// Get a list of mons.
+	mons := gjson.Get(output, "mons.#.name")
+	for _, key := range mons.Array() {
+		retval[key.String()] = 1
```
sabaini, line 86: This looks problematic, as retval at this point is `nil`

microceph/ceph/services.go
```diff
@@ -5,12 +5,105 @@ import (
...
+	}, strategy.Delay(5), strategy.Limit(10), strategy.Backoff(backoff.Linear(5)))
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func getMons() (Set, error) {
+	var retval Set
```
sabaini, line 77: Suggest to create a Set here, you're only declaring it. retval is now nil and you can't assign

You'll need to `retval := make(map[string]int)` or equiv.

microceph/ceph/services.go
```diff
@@ -5,12 +5,105 @@ import (
...
+	mons := gjson.Get(output, "mons.#.name")
+	for _, key := range mons.Array() {
+		retval[key.String()] = 1
+	}
+
+	return retval, nil
+}
+
+func getUpOsds() (Set, error) {
+	var retval Set
```
sabaini, line 93: See above under getMons() for a similar issue with a nil map

microceph/ceph/services.go
```diff
@@ -5,12 +5,105 @@ import (
...
+	return flag
+}
+
+// Table to map fetchFunc for workers (daemons) to a service.
+var serviceWorkerTable = map[string](func () (Set, error)) {
+	"osd": getUpOsds,
+	"mon": getMons,
+}
+
+func RestartCephService(service string) error {
```
sabaini, line 41: Feels like this func would deserve some documentation; other funcs below as well

sabaini (changes requested):
Hi, I did a first pass on code reviewing. Nice work! I have left some comments inline.

microceph/client/client.go
UtkarshBhatthere: Have abstracted the restart the daemons to a separate API.

microceph/ceph/config.go
UtkarshBhatthere: Ack, snapReload has been changed to snapRestart with additional input param to do reloads, since just reloading the osd daemons was not updating the conf.

microceph/ceph/config.go
UtkarshBhatthere: Ack

microceph/api/configs.go
```diff
@@ -0,0 +1,106 @@
...
+	"github.com/lxc/lxd/lxd/response"
+
+	"github.com/canonical/microceph/microceph/api/types"
+	"github.com/canonical/microceph/microceph/ceph"
+	"github.com/canonical/microceph/microceph/client"
+	"github.com/canonical/microceph/microceph/common"
+)
+
+// Config table for ConfigExtras
+var configTable = ceph.GetConfigTable()
```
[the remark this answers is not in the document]
UtkarshBhatthere: Yes, I have moved it to each function where it is required.

microceph/api/configs.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	// Configure the key/value
+	err = ceph.SetConfigItem(common.CephState{State: s}, req)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	services := configTable[req.Key].Daemons
+	defer client.SendRestartRequestToClusterMembers(s, r, services)
```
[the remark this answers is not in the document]
UtkarshBhatthere: My bad, these are supposed to be run as go-routines. This is so that the API call does not waits for the restarting of the ceph daemons.

microceph/ceph/services.go
UtkarshBhatthere: Yeah, since this was being executed in the go routine context, it took me some time to figure out what was causing the daemon exit.

UtkarshBhatthere: Thanks for the thorough review :D

microceph/ceph/services.go
UtkarshBhatthere: ack

microceph/ceph/config.go
UtkarshBhatthere: ack

microceph/ceph/config.go
UtkarshBhatthere: Hmm, My inclination towards using "ConfigTable" was that not only is this particular Map the table for finding services to restart, this is also THE table which indexes all the configs we support being configured. It is then used in various ways (the slice of keys is used to filter configs..) In that essence, using ConfigTable (or ConfigMap) tells that this is the sole Map one needs to be concerned about when looking for configs.

UtkarshBhatthere: I agree with type unification, I have convoluted the ConfigExtras type into the Table itself.

UtkarshBhatthere: @sabaini thoughts about the naming dilemma ?

sabaini: Thanks for the clarification -- agree if it's slated to be used more generically the generic `ConfigTable` works better.

Maybe could add your thoughts about this type as a doc comment ;-)

microceph/ceph/config.go
UtkarshBhatthere: the non-json output is just the value of the key that we inquire for, doing json/json-pretty gives a newline character and the value on the next line. I think we can skip the extra steps and just pump whatever value ceph gives us.

.github/workflows/tests.yml
```diff
@@ -126,6 +126,40 @@ jobs:
           s3cmd --host localhost --host-bucket="localhost/%(bucket)" --access_key=fooAccessKey --secret_key=fooSecretKey --no-ssl put -P ~/test.txt s3://testbucket
           curl -s http://localhost/testbucket/test.txt | grep -F hello-radosgw
 
+      - name: Test Cluster Config
+        run: |
+          set -eux
+          cip=$(ip -4 -j route | jq -r '.[] | select(.dst | contains("default")) | .prefsrc' | tr -d '[:space:]')
+          
+          # pre config set timestamp for service age
+          ts=$(sudo systemctl show snap.microceph.osd.service -f json-pretty | grep -m1 ActiveEnterTimestampMonotonic | awk '{n=split($1,A,"="); print A[n]}')
```
sabaini, line 135: The `systemctl show` command doesn't do json. Also you can directly ask for a property. Suggest to use something like:
```
sudo systemctl show --property ActiveEnterTimestampMonotonic snap.microceph.osd.service | cut -d= -f2
```

.github/workflows/tests.yml
```diff
@@ -126,6 +126,40 @@ jobs:
...
+
+          # set config
+          sudo microceph cluster config set cluster_network $cip/8 --wait
+
+          # post config set timestamp for service age
+          ts2=$(sudo systemctl show snap.microceph.osd.service -f json-pretty | grep -m1 ActiveEnterTimestampMonotonic | awk '{n=split($1,A,"="); print A[n]}')
```
sabaini, line 141: Similar to above comment

.github/workflows/tests.yml
```diff
@@ -126,6 +126,40 @@ jobs:
...
+
+          # Check config output
+          output=$(sudo microceph cluster config get cluster_network | grep -cim1 'cluster_network')
+          if [[ $output -ge 1 ]] ; then
+              echo "Config is present"
+          else
+              exit 1 # Config check failed.
```
sabaini, line 148: Think adding some explicit log output why the check failed would be good, maybe:

echo "config check failed: $output"

.github/workflows/tests.yml
```diff
@@ -126,6 +126,40 @@ jobs:
...
+          fi
+
+          # Check service restarted
+          if [ $ts2 -lt $ts ]; then exit 1; fi
```
sabaini, line 152: Suggest explicit log output incl. timestamps for here as well

.github/workflows/tests.yml
```diff
@@ -126,6 +126,40 @@ jobs:
...
+
+          # reset config
+          sudo microceph cluster config reset cluster_network --wait
+
+          # post config reset timestamp for service age
+          ts3=$(sudo systemctl show snap.microceph.osd.service -f json-pretty | grep -m1 ActiveEnterTimestampMonotonic | awk '{n=split($1,A,"="); print A[n]}')
```
sabaini, line 158: Similar to above

.github/workflows/tests.yml
```diff
@@ -126,6 +126,40 @@ jobs:
...
+
+          # Check service restarted
+          if [ $ts3 -lt $ts2 ]; then exit 1; fi
```
sabaini, line 161: Similar to above

microceph/api/configs.go
```diff
@@ -0,0 +1,119 @@
...
+		return response.InternalError(err)
+	}
+
+	// Clean the key/value
+	err = ceph.RemoveConfigItem(req)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	services := configTable[req.Key].Daemons
```
sabaini, line 98: The restart procedure is similar to the one above in cmdConfigsPut
I'd suggest to factor out the restart logic into a separate function and use that from here and cmdConfigsPut

microceph/ceph/config.go
```diff
@@ -3,58 +3,193 @@ package ceph
...
 	"os"
 	"path/filepath"
 	"strings"
 
+	"github.com/canonical/microceph/microceph/api/types"
 	"github.com/canonical/microceph/microceph/common"
 	"github.com/canonical/microceph/microceph/database"
 )
 
+type ConfigTable map[string]struct{
```
sabaini, line 17: Please add some doc on what the `ConfigTable` type is used for

microceph/ceph/config.go
```diff
@@ -3,58 +3,193 @@ package ceph
...
+	
+	return configs, nil
+}
+
+// updates the ceph config file.
 func updateConfig(s common.StateInterface) error {
 	confPath := filepath.Join(os.Getenv("SNAP_DATA"), "conf")
 	runPath := filepath.Join(os.Getenv("SNAP_DATA"), "run")
-
+
```
sabaini, line 153: Nit: please clean up spurious white space changes here and below

microceph/ceph/services.go
```diff
@@ -4,13 +4,128 @@ import (
...
+	}
+
+	// Restart the service.
+	snapRestart(service, false)
+
+	// Check all the daemons available before Restart are up.
+	err = retry.Retry(func(i uint) error {
+		iWorkers, err := serviceWorkerTable[service]()
+		if err != nil {
+			return err
```
sabaini, line 78: Might add a `logger.Debug()` output logging the outcome of the attempt

sabaini (changes requested):
Hey, thanks for the update
I've added some comments/nits inline. Please also see my previous comment abt GetConfigTable()

.github/workflows/tests.yml
UtkarshBhatthere: ack

.github/workflows/tests.yml
UtkarshBhatthere: ack

.github/workflows/tests.yml
UtkarshBhatthere: ack

.github/workflows/tests.yml
UtkarshBhatthere: ack

microceph/ceph/config.go
UtkarshBhatthere: Ack

.github/workflows/tests.yml
UtkarshBhatthere: ack

.github/workflows/tests.yml
UtkarshBhatthere: ack

microceph/ceph/services.go
UtkarshBhatthere: As discussed, a better place for logging the intermittent failures would be inside the individual handles for ceph services.

microceph/ceph/config.go
UtkarshBhatthere: ack

microceph/ceph/services.go
UtkarshBhatthere: Done.

microceph/api/services.go
```diff
@@ -29,6 +31,33 @@ func cmdServicesGet(s *state.State, r *http.Request) response.Response {
...
+
+func cmdRestartServicePost(s *state.State, r *http.Request) response.Response {
+	var services types.Services
+
+	err := json.NewDecoder(r.Body).Decode(&services)
+	if err != nil {
+		logger.Errorf("Failed decoding restart services: %v", err)
+		return response.InternalError(err)
+	}
+
```
sabaini, line 48: I'm thinking we should right at the boundary of our API service to user input handling so we know we get valid service names. Can we e.g. check that each of the services provided by the user are valid keys in our service table? Then we can later simplify our error handling because we know we sanitized user input at the boundary.

microceph/ceph/services.go
```diff
@@ -4,13 +4,132 @@ import (
...
+var serviceWorkerTable = map[string](func () (Set, error)) {
+	"osd": getUpOsds,
+	"mon": getMons,
+}
+
+// Restarts (in order) all Ceph Services provided in the input slice on the host.
+func RestartCephServices(services []string) error {
+	for i := range services {
+		err := RestartCephService(services[i])
+		if err != nil {
```
sabaini, line 47: If we check user input at the API boundary in the POST handler we could simplify the error handling here

sabaini (changes requested):
Heya :-)

Couple of comments/questions inline.

Probably will also need to rebase

microceph/api/configs.go
```diff
@@ -0,0 +1,86 @@
...
+		return response.InternalError(err)
+	}
+
+	// If a valid key string is passed, fetch that key.
+	if len(req.Key) > 0 {
+		configs, err = ceph.GetConfigItem(req)
+	} else {
+		// Fetch all configs.
+		configs, err = ceph.ListConfigs()
+	}
```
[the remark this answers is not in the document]
UtkarshBhatthere: Thank you :D

microceph/api/configs.go
```diff
@@ -0,0 +1,86 @@
...
+		return response.InternalError(err)
+	}
+
+	// Configure the key/value
+	err = ceph.SetConfigItem(req)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	services := configTable[req.Key].Daemons
```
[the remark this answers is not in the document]
UtkarshBhatthere: The output of GetConfigTable is already set at compile time. So it would always be available.

sabaini (approved):

sabaini:
This lgtm, thanks for the changes @UtkarshBhatthere 
Note the rtd linkcheck issue is being adressed https://github.com/sphinx-doc/sphinx/issues/10927

UtkarshBhatthere:
Thanks :D

PR #117, opened 2023-03-31
Added test script for RGW S3 interface.

UtkarshBhatthere:
Added a script for s3 IO testing on Microceph RGW.
Signed-off-by: utkarshbhatthere <utkarsh.bhatt@canonical.com>

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -21,6 +21,11 @@ jobs:
           sudo iptables -F FORWARD  || true
           sudo ip6tables -F FORWARD || true
 
+      - name: set up python ${{ matrix.python-version }}
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+
       - name: Install dependencies
         run: |
           sudo snap refresh
@@ -28,6 +33,13 @@ jobs:
           sudo lxd init --auto
           sudo snap install snapcraft --classic
           snap list
+          # Python
+          sudo python -m pip install --upgrade pip
+          sudo pip install flake8 pep8-naming boto3
+
+      - name: Lint Check
+        run: |
+          flake8 . --count --show-source --statistics
 
       - name: Build snaps
         run: snapcraft
@@ -118,13 +130,8 @@ jobs:
       - name: Exercise RGW
         run: |
           set -eux
-          sudo microceph.radosgw-admin user create --uid=test --display-name=test
-          sudo microceph.radosgw-admin key create --uid=test --key-type=s3 --access-key fooAccessKey --secret-key fooSecretKey
-          sudo apt-get -qq install s3cmd
-          echo hello-radosgw > ~/test.txt
-          s3cmd --host localhost --host-bucket="localhost/%(bucket)" --access_key=fooAccessKey --secret_key=fooSecretKey --no-ssl mb s3://testbucket
-          s3cmd --host localhost --host-bucket="localhost/%(bucket)" --access_key=fooAccessKey --secret_key=fooSecretKey --no-ssl put -P ~/test.txt s3://testbucket
-          curl -s http://localhost/testbucket/test.txt | grep -F hello-radosgw
+          sudo microceph.radosgw-admin user create --uid=test --display-name=test > keys.json
+          sudo python3 scripts/appS3.py http://localhost:80 keys.json --obj-num 2
 
       - name: Upload artifacts
         if: always()
--- a/scripts/appS3.py
+++ b/scripts/appS3.py
@@ -1,0 +1,91 @@
+import string
+import random
+import boto3
+import json
+import argparse
+
+
+def app_handle(args):
+    keys_path = args.keys
+    endpoint = args.endpoint
+
+    # Fetch Auth
+    with open(keys_path, 'r') as keys_file:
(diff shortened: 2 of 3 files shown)


scripts/appS3.py
```diff
@@ -0,0 +1,103 @@
...
+        type=bool,
+        default=False,
+        help="Setting this to true would not cleanup the pushed objects.",
+    )
+    argparse.set_defaults(func=app_handle)
+
+    # Parse the args.
+    args = argparse.parse_args()
+
+    if not args.keys or not args.endpoint:
```
lmlg, line 99: You can mark both `keys` and `endpoint` with `required=True` in the `add_argument` call.

scripts/appS3.py
```diff
@@ -0,0 +1,103 @@
...
+    argparse.add_argument(
+        "--obj-num",
+        type=int,
+        default=1,
+        help="Number of objects to upload to S3.",
+    )
+    argparse.add_argument(
+        "--no-delete",
+        type=bool,
+        default=False,
```
lmlg, line 91: For boolean arguments, I think it's best to set `action="store_true"` so that both the default is set and also so that the user doesn't need to pass a value.

scripts/appS3.py
```diff
@@ -0,0 +1,103 @@
+import os
+import string
+import time
+import random
+import boto3
+import json
+import argparse
+
+# seed random
+random.seed(time.time())
```
lmlg, line 10: This should be moved to the `if __name__ == "__main__"` block below.

scripts/appS3.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def app_handle(args):
+    keys_path = args.keys
+    endpoint = args.endpoint
+    if not os.path.isfile(keys_path):
```
lmlg, line 16: This is unneeded, as the `open` call will fail below if the file isn't present.

lmlg (changes requested):
Overall looks good, just a few comments.

scripts/appS3.py
UtkarshBhatthere: ack

scripts/appS3.py
UtkarshBhatthere: i think default=False does the same, for the execution used on the github action:
```
ubuntu@microcephtest:~$ python3 script.py http://localhost:80 test.json --obj-num 2
Namespace(endpoint='http://localhost:80', keys='test.json', obj_num=2, no_delete=False, func=<function app_handle at 0x7f1015eefd90>)
```

UtkarshBhatthere: @lmlg thoughts ?

wolsen: @UtkarshBhatthere The default=False is subtly differently. IT makes it a flag on the commandline versus an argument that expects a value if it is specified. In the case where you have wish to pass `--no-delete` you must specify it like `foo.py --no-delete True ...` when in reality all you want is the `--no-delete` as a flag on the commandline.

The 'store_true' option referenced by @lmlg does that.

UtkarshBhatthere: Ack, Thank you for the explanation @wolsen 🙇🏼

lmlg (approved):
LGTM, good stuff :)

.github/workflows/tests.yml
```diff
@@ -21,13 +21,25 @@ jobs:
...
       - name: Install dependencies
         run: |
           sudo snap refresh
           sudo snap set lxd daemon.group=adm
           sudo lxd init --auto
           sudo snap install snapcraft --classic
           snap list
+          # Python
+          sudo python -m pip install --upgrade pip
+          sudo pip install flake8 pep8-naming boto3 argparse
```
sabaini, line 38: The argparse module should be in stdlib since 2.7 / 3.2 -- do we need to support earlier Pythons, or is there any other reasons to install?

scripts/appS3.py
```diff
@@ -0,0 +1,95 @@
...
+
+def rand_str(length: int):
+    return "".join(
+        random.choices(string.ascii_uppercase + string.digits, k=length)
+    )
+
+
+if __name__ == "__main__":
+    # seed random
+    random.seed(time.time())
```
sabaini, line 61: This is probably not necessary. We don't need high level of randomness, and typically the random module would either use an OS-provided randomness source (e.g. /dev/urandom) or use system time automatically

scripts/appS3.py
```diff
@@ -0,0 +1,95 @@
...
+
+    argparse = argparse.ArgumentParser(
+        description="An application which uses S3 for storage",
+        epilog="Ex: python3 app.py --keys keys.txt",
```
sabaini, line 65: Typo in script name, should probably be `python3 appS3.py ...`

sabaini (changes requested):
Hey, good stuff. I've added some minor nits and a Q
cheers!

.github/workflows/tests.yml
UtkarshBhatthere: ack

UtkarshBhatthere: I have removed the explicit install of argparse

UtkarshBhatthere:
Script has been moved to https://github.com/canonical/microceph/pull/127

Issue #112, opened 2023-03-28
OOM will be invoked with the documented steps


nobuto-m:
The installation tutorials uses the default VM size of LXD, which is translated to 1 CPU core and 1GB of memory if I'm not mistaken.
https://github.com/canonical/microceph/blob/38a6bb606617ec00355912af2aeef4d6e6520ffe/docs/tutorial/install.rst#setup-vms

This leads to OOM after some time even with an idle state. When this OOM happens, the system gets into a very tricky state to diagnose especially when microceph is used for other components such as microcloud with LXD and dqlite.

```
Mar 27 21:48:34 microceph-3 kernel: [29726.347384] systemd invoked oom-killer: gfp_mask=0x1100cca(GFP_HIGHUSER_MOVABLE), order=0, oom_score_adj=0
Mar 27 21:48:34 microceph-3 kernel: [29726.347450]  oom_kill_process.cold+0xb/0x10
Mar 27 21:48:34 microceph-3 kernel: [29726.347615] [  pid  ]   uid  tgid total_vm      rss pgtables_bytes swapents oom_score_adj name
Mar 27 21:48:34 microceph-3 kernel: [29726.347697] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=init.scope,mems_allowed=0,global_oom,task_memcg=/system.slice/snap.microceph.mgr.service,task=ceph-mgr,pid=2968,uid=0
Mar 27 21:48:34 microceph-3 kernel: [29726.347724] Out of memory: Killed process 2968 (ceph-mgr) total-vm:1002128kB, anon-rss:250092kB, file-rss:0kB, shmem-rss:0kB, UID:0 pgtables:752kB oom_score_adj:0
Mar 27 21:48:34 microceph-3 kernel: [29728.401021] oom_reaper: reaped process 2968 (ceph-mgr), now anon-rss:0kB, file-rss:0kB, shmem-rss:0kB
Mar 27 21:48:35 microceph-3 kernel: [29729.682899] oom_reaper: reaped process 4956 (snapd), now anon-rss:0kB, file-rss:0kB, shmem-rss:0kB
```

```
Control Group                                                       Tasks   %CPU   Memory  Input/s Output/s
/                                                                     367    4.9   899.0M       0B   113.2K
system.slice                                                          221    6.9   657.6M       0B    18.0K
system.slice/snap.microceph.mgr.service                                70    0.3   274.1M       0B     1.3K
[...]

PR #118, opened 2023-03-31
Doc fix: use larger instance types in tutorial

sabaini:
Fixes issue #112

### DIFF
--- a/docs/tutorial/install.rst
+++ b/docs/tutorial/install.rst
@@ -7,9 +7,10 @@ Setup VMs
 
 .. code-block:: shell
 
-    lxc launch images:ubuntu/22.04/cloud microceph-1 --vm
-    lxc launch images:ubuntu/22.04/cloud microceph-2 --vm
-    lxc launch images:ubuntu/22.04/cloud microceph-3 --vm
+    lxc launch images:ubuntu/22.04/cloud microceph-1 --vm -c limits.cpu=2 -c limits.memory=4GiB
+    lxc launch images:ubuntu/22.04/cloud microceph-2 --vm -c limits.cpu=2 -c limits.memory=4GiB
+    lxc launch images:ubuntu/22.04/cloud microceph-3 --vm -c limits.cpu=2 -c limits.memory=4GiB
+
 
 Add storage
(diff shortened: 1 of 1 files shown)


docs/tutorial/install.rst
```diff
@@ -7,9 +7,9 @@ Setup VMs
 
 .. code-block:: shell
 
-    lxc launch images:ubuntu/22.04/cloud microceph-1 --vm
-    lxc launch images:ubuntu/22.04/cloud microceph-2 --vm
-    lxc launch images:ubuntu/22.04/cloud microceph-3 --vm
+    lxc launch images:ubuntu/22.04/cloud microceph-1 --vm --type m1.large
+    lxc launch images:ubuntu/22.04/cloud microceph-2 --vm --type m1.large
+    lxc launch images:ubuntu/22.04/cloud microceph-3 --vm --type m1.large
```
nobuto-m, line 12: If I understand it correctly, m1.large comes with 2 vCPU cores and 7.5GB memory. Which means that this tutorial requires 6 vCPU and 22.5GB memory in total.

From a Ceph point of view, this requirement is absolutely correct since `osd_memory_target` for example is set to 4GB (per OSD) by default. At the same time, if somebody would like to test **micro**ceph quickly, then most laptops in the market gets out of the picture.

From an user story view, it would be nice if there is an opt-in feature to fit into 16GB of memory by having a lower memory per VM like 4GB and setting osd_memory_target as 2GB or something by microceph.

stgraber: My usual setup is `-c limits.cpu=2 -c limits.memory=4GiB` which works just fine for development and demos.

While Ceph can indeed use up to 4GiB per OSD unless configured otherwise, it won't actually do that instantly, it's just a maximum that it's allowed to reach and in my experience pretty rarely reaches even in production environment.

So I suspect that using `-c limits.cpu=2 -c limits.memory=4GiB` is probably a better alternative here as that will still fit on systems that have 16GiB of RAM. Worth noting that similar to Ceph, LXD won't reserve the memory, so in practice you'll be running well below that potential 12GiB.

PR #123, opened 2023-04-16
Replace symlink with bind-mount for default ceph directories

### LINKED ISSUE #122: Bootstrap fails on a host with ceph-common installed
Due to using the symlink layout in the snap for /etc/ceph, /var/log/ceph, etc, bootstrapping fails after installing the ceph-common package because these locations aren't empty on the host system.

```
root@mceph:~# microceph cluster bootstrap
cannot update snap namespace: cannot create symlink in "/etc/ceph": existing file in the way
snap-update-ns failed with code 1
```

ChrisMacNaughton:
We cannot assume that these directories will be empty which causes the snap layotu handling to fail. Changing to bind works around this issue.

Closes #122

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -19,13 +19,13 @@ layout:
   /usr/lib/$CRAFT_ARCH_TRIPLET/rados-classes:
     symlink: $SNAP/lib/$CRAFT_ARCH_TRIPLET/rados-classes
   /etc/ceph:
-    symlink: $SNAP_DATA/conf
+    bind: $SNAP_DATA/conf
   /usr/share/ceph:
-    symlink: $SNAP/share/ceph
+    bind: $SNAP/share/ceph
   /var/lib/ceph:
-    symlink: $SNAP_COMMON/data
+    bind: $SNAP_COMMON/data
   /var/log/ceph:
-    symlink: $SNAP_COMMON/logs
+    bind: $SNAP_COMMON/logs
 
 apps:
   # Service


ChrisMacNaughton:
I've just manually tested this on a node with ceph-common installed and confirmed the presence of content in /etc/ceph (specifically, /etc/ceph/rbdmap) and confirmed that `microceph bootstrap` works correctly

sabaini:
This makes a lot of sense to me in terms of robustness. 

Two questions -- I wonder what the impact on startup time is? 
Also I think it might be good to have some feedback and/or documentation that we're shadowing their local ceph install?

ChrisMacNaughton:
So, that really crazy thing is that we're not actually consuming things from the hosts's filesystem!!

```
root@mceph:~/microceph# snap run --shell microceph
root@mceph:/root/microceph# ls /etc/ceph
ceph.client.admin.keyring  ceph.conf  ceph.keyring
root@mceph:/root/microceph#
exit
root@mceph:~/microceph# ls /etc/ceph
rbdmap
```

Regarding timing, in a microceph with this change:

```
# time microceph status
MicroCeph deployment summary:
- mceph (10.90.140.1)
  Services: mds, mgr, mon
  Disks: 0

real    0m0.187s
user    0m0.028s
sys     0m0.102s
```

Without (currently on edge):

```
# time microceph status
MicroCeph deployment summary:
- mceph (10.90.140.1)
  Services: mds, mgr, mon
  Disks: 0

real    0m0.139s
user    0m0.041s
sys     0m0.071s
```

sabaini:
Ack, really think we want bind-mount -- merging

PR #127, opened 2023-05-09
Added S3 user management

UtkarshBhatthere:


UtkarshBhatthere:
The PR lacks test, will remove from drafts once I have added them.

sabaini (changes requested):
Heya,

this looks very useful.

Some comments after a first quick look:

User interface: we talked about this a bit but jftr. -- think the `microceph s3 [create|delete|...] ...` might be a bit confusing as it's not clear that this is about creating or deleting users. Maybe could do another subcommand e.g. `microceph s3 user [create|delete|...] ...` or rename the subcommand to something like `microceph s3-user [create|delete|...] ...`

REST API: it's recommended to give every entity in a REST api their own URI, and have the REST verbs (GET/PUT/DELETE/...) act on those entities

For this API the entities would be users. It's a convention (though not required by REST) to have directories map to collections of entities.

A mapping to URIs could therefore look something like this, with an example user fooUser

```
GET /s3/users # list user collection
POST /s3/users # create a new user in collection with data from the req. body

GET /s3/users/fooUser # get details for fooUser (no data required)
PUT /s3/users/fooUser # update fooUser with data from req. body
DELETE /s3/users/fooUser # delete fooUser (no data required)
```

Note there's no POST /s3/users/fooUser as POST is for creating new entities.  

Thanks!

UtkarshBhatthere:
@sabaini I have modified the APIs to be using ""services/rgw/user"".

microceph/api/s3.go
```diff
@@ -0,0 +1,79 @@
...
+	"github.com/canonical/microcluster/state"
+)
+
+// /1.0/resources endpoint.
+var s3Cmd = rest.Endpoint{
+	Path:   "services/rgw/user",
+	Get:    rest.EndpointAction{Handler: cmdS3Get, ProxyTarget: true},
+	Put:    rest.EndpointAction{Handler: cmdS3Put, ProxyTarget: true},
+	Delete: rest.EndpointAction{Handler: cmdS3Delete, ProxyTarget: true},
+}
```
sabaini, line 20: For optimal RESTy-ness I'd like to suggest an API structure like this:

GET /services/rgw/user  # returns a list of users
GET /services/rgw/user/{uid} # returns info about a specific user

POST /services/rgw/user # create a new user 

PUT /services/rgw/user/{name} # update existing user

DELETE /services/rgw/user/{name} # delete user

Something like this can be used to specify a path with a {uid} var:

```
var s3Cmd = rest.Endpoint{
	Path:   "services/rgw/user/{name}",
	Get:    rest.EndpointAction{Handler: cmdS3GetSingle, ProxyTarget: true},
        Put: 
...

func cmdS3GetSingle(s *state.State, r *http.Request) response.Response {
	var err error
	var name string
	name, err = url.PathUnescape(mux.Vars(r)["name"])
	if err != nil {
		return response.BadRequest(err)
	}
	// do something with user name
	user = ceph.GetS3User(name)
...
```

microceph/client/s3.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/canonical/microceph/microceph/api/types"
+	"github.com/canonical/microcluster/client"
+)
+
+func GetS3User(ctx context.Context, c *client.Client, user *types.S3User) (string, error) {
+	queryCtx, cancel := context.WithTimeout(ctx, time.Second*60)
+	defer cancel()
+
+	ret := ""
+	err := c.Query(queryCtx, "GET", api.NewURL().Path("services", "rgw", "user"), user, &ret)
```
sabaini, line 19: With the API structure change I've suggested above the URL would be built with something like
```
err = c.Query(queryCtx, "GET", api.NewURL().Path("services", "rgw", "user", strconv.FormatInt(user.name, 10)), nil, nil)
```

sabaini (changes requested):
Heya Utkarsh, thanks for the update, added a comment about REST API structure

UtkarshBhatthere:
Thanks for the detailed review @sabaini 🙏🏼 . I will implement your RESTy suggestions in and make the PR ready.

docs/how-to/s3-user.rst
```diff
@@ -0,0 +1,106 @@
+Manage S3 users on MicroCeph
+=============================
+
+MicroCeph provides an easy to use interface for creating, viewing and deleting s3 users for interfacing with the RGW endpoint.
```
sabaini, line 4: Suggest: capitalize S3 here and below (proper noun)

docs/how-to/s3-user.rst
```diff
@@ -0,0 +1,106 @@
...
+This enables smooth and easy access to Object Storage.
```
sabaini, line 5: Suggest: s/Object Storage/object storage/

docs/how-to/s3-user.rst
```diff
@@ -0,0 +1,106 @@
...
+    $ sudo microceph s3-user list
+    +---+-------------+
+    | # |    NAME     |
+    +---+-------------+
+    | 1 | newTestUser |
+    +---+-------------+
+    | 2 | testUser    |
+    +---+-------------+
+
+3. Get details of a an s3 user (optionally use --json flag to get complete details):
```
sabaini, line 80: Typo, s/a an/an/

docs/how-to/s3-user.rst
```diff
@@ -0,0 +1,106 @@
...
+     - Fetch key information of the provided s3 (radosgw) user
+   * - list
+     - List all s3 (radosgw) users
+.. note:: Users can additionally provide --json flag to create and get commands to dump a much detailed 
+
+1. Create an S3 user (optionally provide --access-key --secret and --json)
+
+  .. code-block:: shell
+
+    $ sudo microceph s3-user create newTestUser --access-key=ThisIsAccessKey --secret=ThisIsSecret --json
```
sabaini, line 27: Think the cmdline here doesn't match up with the implementation?

microceph/api/client_s3.go
```diff
@@ -0,0 +1,79 @@
...
+	"net/http"
+
+	"github.com/canonical/lxd/lxd/response"
+	"github.com/canonical/microceph/microceph/api/types"
+	"github.com/canonical/microceph/microceph/ceph"
+	"github.com/canonical/microcluster/rest"
+	"github.com/canonical/microcluster/state"
+)
+
+// /1.0/resources endpoint.
```
sabaini, line 14: Copy/paste error :-)

microceph/api/client_s3.go
```diff
@@ -0,0 +1,79 @@
...
+	} else {
+		listOutput, err := ceph.ListS3Users()
+		if err != nil {
+			return response.SmartError(err)
+		}
+		return response.SyncResponse(true, listOutput)
+	}
+}
+
+func cmdClientS3Put(s *state.State, r *http.Request) response.Response {
```
sabaini, line 47: For maximum RESTfulness we should be using a POST to the `/client/s3` url to create a user

microceph/api/client_s3.go
```diff
@@ -0,0 +1,79 @@
...
+
+// /1.0/resources endpoint.
+var clientS3Cmd = rest.Endpoint{
+	Path:   "client/s3",
+	Get:    rest.EndpointAction{Handler: cmdClientS3Get, ProxyTarget: true},
+	Put:    rest.EndpointAction{Handler: cmdClientS3Put, ProxyTarget: true},
+	Delete: rest.EndpointAction{Handler: cmdClientS3Delete, ProxyTarget: true},
+}
+
+func cmdClientS3Get(s *state.State, r *http.Request) response.Response {
```
sabaini, line 22: RESTfulness: 
- retrieving a list of users should be a GET to `/client/s3`
- retrieving a single user should be a GET to `/client/s3/{userid}`

See the /disks/{osdid} handler in microceph/api/disks.go for an example on how to implement a handler for a path variable i.e. {userid}

microceph/api/client_s3.go
```diff
@@ -0,0 +1,79 @@
...
+
+	output, err := ceph.CreateS3User(req)
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	return response.SyncResponse(true, output)
+}
+
+func cmdClientS3Delete(s *state.State, r *http.Request) response.Response {
```
sabaini, line 64: RESTfulness: this should be using a DELETE against `/client/s3/{userid}` (also cf. above for getting users)

microceph/cmd/microceph/s3-user.go
```diff
@@ -0,0 +1 @@
+package main
```
sabaini, line 1: I guess this is not actually used?

sabaini (changes requested):
Hey @UtkarshBhatthere did a quick once over, added some suggestions mostly around API RESTfulness

PR #128, opened 2023-05-11
Add checks for dmcrypt intf connection

UtkarshBhatthere:
```
ubuntu@bastion:~$ sudo microceph disk add  /dev/disk/by-id/virtio-d15384b1-7159-4a91-a --wipe --encrypt
Error: Failed adding new disk: Encryption unsupported on this machine: Missing dm_crypt module: stat /sys/module/dm_crypt: no such file or directory
Use "sudo snap connect microceph:dm-crypt" to enable disk encryption.
```

### DIFF
--- a/microceph/ceph/osd.go
+++ b/microceph/ceph/osd.go
@@ -193,6 +193,12 @@ func checkEncryptSupport() error {
 		return fmt.Errorf("Missing /dev/mapper/control: %w", err)
 	}
 
+	// Check if the dm-crypt interface is not connected.
+	if !isIntfConnected("dm-crypt") {
+		helper := fmt.Sprint("Use \"sudo snap connect microceph:dm-crypt\" to enable encryption.")
+		return fmt.Errorf("dm-crypt interface connection missing: \n%s", helper)
+	}
+
 	// Check if we have the dm_crypt module
 	inf, err := os.Stat("/sys/module/dm_crypt")
 	if err != nil || inf == nil || !inf.IsDir() {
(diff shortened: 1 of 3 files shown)


microceph/ceph/osd.go
```diff
@@ -196,7 +196,8 @@ func checkEncryptSupport() error {
 	// Check if we have the dm_crypt module
 	inf, err := os.Stat("/sys/module/dm_crypt")
 	if err != nil || inf == nil || !inf.IsDir() {
-		return fmt.Errorf("Missing dm_crypt module: %w", err)
+		helper := fmt.Sprint("Use \"sudo snap connect microceph:dm-crypt\" to enable disk encryption.")
+		return fmt.Errorf("Missing dm_crypt module: %w\n%s", err, helper)
```
sabaini, line 200: That's not quite right -- if the dm-crypt snap connection is missing it will surface as an EPERM using the cryptsetup command, even though the actual dm_crypt kernel module will be present.

You will need to check for the missing dm-crypt snap connection separately; suggest to try probing with `snapctl is-connected dm-crypt` for this; also cf. https://forum.snapcraft.io/t/add-dm-crypt-interface-to-microceph/34995/4?u=petersabaini

UtkarshBhatthere:
Hey @sabaini, I see that CI tests use --encrypt and works with this check enabled. Is there any other testing you'd suggest for this to be done ?

sabaini:
Thanks, changes lgtm

> Hey @sabaini, I see that CI tests use --encrypt and works with this check enabled. Is there any other testing you'd suggest for this to be done ?

Some testing (either unit or functest) would be good. I think it should be possible to do functesting by 

- not immediately connecting dm-crypt
- attempt to add the disk and check for expected fault in stderr
- only then connect dm-crypt and add the disk

sabaini:
lgtm, cheers

PR #129, opened 2023-05-11
Added Documentation for Cluster config management

UtkarshBhatthere:


docs/how-to/cluster-cfg.rst
```diff
@@ -0,0 +1,60 @@
+Configure cluster_network/ public_network
+=========================================
+
+
+Overview
+--------
+Network configuration is critical for building a high performance Ceph Storage Cluster. Since, Ceph Clients make requests directly to Ceph OSD Daemons. Ceph OSD Daemons perform data replication on behalf of Ceph Clients, which means replication and other factors impose additional loads on Ceph Storage Cluster networks. Refer to `Ceph Network Config <https://docs.ceph.com/en/latest/rados/configuration/network-config-ref/>`_.
```
sabaini, line 7: This may need some clarification -- I'd suggest something like the below

----

Network configuration is critical for building a secure and well performing Ceph cluster. 

For resilience and performance reasons, dual NIC systems typically should implement active/active bonding or adopt a layer 3 multipath strategy to ensure maximum throughput and availability, and utilize a single network.

However keep in mind that clients are communicating directly with Ceph OSD daemons -- Ceph does not perform request routing. To enhance security and stability it can therefore be advantageous to split public and cluster network traffic so that client traffic flows on a public net while cluster traffic such as replication and backfilling utilizes a separate net. This helps to prevent malicious or malfunctioning clients from disrupting cluster backend operations.

Refer to `Ceph Network Config <https://docs.ceph.com/en/latest/rados/configuration/network-config-ref/>`_ for more detail.

docs/how-to/cluster-cfg.rst
```diff
@@ -0,0 +1,60 @@
...
+
+Implementation
+--------------
+Microceph cluster config subcommands rely on ‘ceph config’ as the single source of truth for config values and use the same for getting/setting the configs. After updating (setting/resetting) a config value the same request is forwarded to other hosts on the microceph cluster for restarting particular daemons on the host so that updated configs can take effect.
```
sabaini, line 11: For a How-To I'd suggest leaving out details of the implementation. Maybe only mention that affected services will be restarted -- this could be moved to the end of the `Usage` section though.

docs/how-to/cluster-cfg.rst
```diff
@@ -0,0 +1,60 @@
...
+
+Usage
```
sabaini, line 13: I wonder if this section as written is not more of an `Explanation` than a `How-To` in the Diataxis sense

A howto should be practical and for work -- i.e. describe steps to accomplish a practical work task. While the section describes in general how to configure MicroCeph

Think it would probably be good to have both a how-to to describe splitting up public and cluster net while also having a more generic explanation on configuring MicroCeph

docs/how-to/cluster-cfg.rst
```diff
@@ -0,0 +1,60 @@
...
+       $ sudo microceph cluster config list
+       +---+-----------------+---------------+
+       | # |       KEY       |     VALUE     |
+       +---+-----------------+---------------+
+       | 0 | cluster_network | 10.5.2.165/16 |
+       +---+-----------------+---------------+
+       | 1 | public_network  | 10.5.2.165/16 |
+       +---+-----------------+---------------+
+
+4. Resetting a config key can performed using the 'reset' command:
```
sabaini, line 49: Minor suggestion: suggest to add a one-sentence explanation on what resetting means i.e. setting to default values

sabaini (changes requested):
Hey, some comments inline -- thanks!

docs/how-to/cluster-cfg.rst
sabaini: Replying to myself here -- this might be better suited for `Explanations` as well actually

docs/how-to/cluster-cfg.rst
```diff
@@ -0,0 +1,60 @@
+Configure cluster_network/ public_network
```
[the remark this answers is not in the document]
UtkarshBhatthere: Well it was originally the whole sentence, but I felt the 'in Microceph' bit was redundant.... will revert to the whole sentence.

docs/how-to/cluster-cfg.rst
UtkarshBhatthere: yeah, exactly my thought. I am moving these bits to a separate doc in explanations.

docs/how-to/cluster-cfg.rst
UtkarshBhatthere: +1 :D

docs/how-to/cluster-cfg.rst
```diff
@@ -0,0 +1,50 @@
+Configure cluster_network/ public_network in Microceph
+======================================================
+
+The microceph cluster configuration CLI supports, setting, getting, resetting and listing supported config keys (currently 'cluster_network' and 'public_network'). It is designed to be consistent and intuitive.
```
sabaini, line 4: Typo: one comma too many
"CLI supports getting, setting, ..."

docs/how-to/cluster-cfg.rst
```diff
@@ -0,0 +1,50 @@
...
+
```
sabaini, line 5: Suggest adding a small overview on the concrete task, something like

> To configure the cluster network, the `cluster_network` config key needs to be set to the desired CIDR. Similarly, for configuring the public network, set the `public_network` config key to a CIDR

sabaini (changes requested):
Nice, cheers! Typo and a suggestion inline

docs/how-to/cluster-cfg.rst
UtkarshBhatthere: ack

docs/how-to/cluster-cfg.rst
UtkarshBhatthere: Placing it as plain text felt a bit out of place, so I added a table with key name a description expressing the exact same thing.

sabaini:
lgtm! :ship:

PR #131, opened 2023-05-17
Doc update: list FDE prereq

sabaini:


docs/explanation/fde-osd.rst
```diff
@@ -26,6 +26,16 @@ Limitations
 **Warning:** It is important to note that MicroCeph FDE *only* encompasses OSDs. Other data, such as state information for monitors, logs, configuration etc., will *not* be encrypted by this mechanism.
 
 
+Prerequisites
```
UtkarshBhatthere, line 29: Minor Nit: I think the pre-requisites section should be moved up maybe just after the overview.

UtkarshBhatthere (changes requested):
Just a minor nit with placement of the section. Unrelated to this particular change.... I feel we need to do some diataxis segregation for FDE doc but tbh I have no idea how that would look like.

docs/explanation/fde-osd.rst
sabaini: Ack, reasonable

sabaini:
Not wrong wrt to the Diataxis split. I'll add that to the todo list

UtkarshBhatthere:
Ship it! 🛳️

PR #135, opened 2023-05-23
Toward end-to-end testing in lxd

nihr43:
I saw the recording of Chris MacNaughton presenting at Ceph Days a few days ago and threw this together.
This is intended to be analogous to [microk8s/tests](https://github.com/canonical/microk8s/tree/master/tests).
I see there are some basic tests under .github/workflows; perhaps this approach will add some value - im thinking upgrade paths, cluster recovery, various distro smoketests, etc.
So far this only knows how to bootstrap and join 'n' nodes as a cluster.  I wanted to see if this seems useful to you all and get some thoughts before putting more effort into it and coming up with some actual test cases.  Some obvious todos are of course adding osds and adding the ability to build and inject the snap locally rather than downloading.
Thanks.

update; local `snapcraft` artifacts can be installed in lieu of 'channel'.


sabaini:
Hey @nihr43,

many thanks for the PR! I wanted to run cluster tests for some time as well, and this would be a nice basis.

One thing I worry about with running clustered tests for now is resource requirements, as default GH action runners are a bit memory constrained here. Their default runner comes with 7GB of RAM; we typically recommend 4GB per node for stable services (OSDs, MGRs, MON, MDS, RGW, ... these all can get mem hungry), so we fall short there.

However! I hope we'll be able to utilize beefier runner machines in the near future though, so we might be able to run clustered tests soon. So this would be good to have then!


Some remarks after a quick read:

- Code organization-wise I'd suggest to move the modules and script to a package, something like `tests/deploytool` or similar
- Please run code through `black` (with default settings)
- Please organize commits around frontend/backend/CI changes
- As for the CI change, we'd need to exercise the deploy script by replacing part of the .github workflow steps for testing

nihr43:
Hi @sabaini ;

I've reorganized things so it can either be imported or run directly.
ci-wise; yeah, i'd be surprised if github hosted runners can run kvm vms in the first place.  I wonder what the microk8s and lxd teams do; i should look.
'Please organize commits around...'  - do you mean commit frequency / style?  I may need more guidance here.

Moving on, how do you envision test cases being organized?  Something along the lines of `--runtests`?  or seperate module/scripts?  As a first pass I was thinking i should refactor some of the instance handling into a Node class, and then create a basic add node -> remove node -> check ceph status test.

sabaini:
Hey @nihr43,

apologies for not getting back sooner -- I thought I had replied but now see I apparently didn't. Sorry about that!

Wrt. commit organization: please squash commits such that there is one commit for (as applicable):

- API changes 
- Documentation (doc: Update XYZ for files in doc/)
- MicroCeph CLI (microceph/client)
- MicroCeph daemon (microcephd)
- Tests (tests/ folder)
- CI changes (.github)

Regarding test runs: when you say `--runtests` do you mean a single test runner that takes the tests to be run as a commandline arg?

Refactoring: yes, a Node class would make sense to me as well, possibly a separate OSD class as well.

Thanks!

UtkarshBhatthere:
@nihr43 Interesting work, Can you confirm if the PR uses LXD Virtual machines or Containers for the test runs ? Have you explored the Containers approach ?

nihr43:
@UtkarshBhatthere - they're VMs because [volumes of content type `block`](https://github.com/canonical/microceph/pull/135/files#diff-abc50b17140878ff421ded24d61a0a4d49609ad777348bad31a11e06e202904fR50) ["can only be attached to virtual machines"](https://github.com/canonical/lxd/blob/main/doc/explanation/storage.md#content-types).

Perhaps it could be done in containers by feeding microceph plain zeroed-out files as block devices; i havent tried that yet.  That may even make a few things a bit simpler now that i think about it..

UtkarshBhatthere:
@nihr43  Yes, At this point we can't add block volumes to lxd containers and can't run lxd virtual machines on the standard github runners. My plan was to use lxd containers since microceph does not use lvm for creating OSDs (lvm is problematic in lxd containers) we can practically use loop devices to back the OSD. (example in the single node test runner). If you could incorporate this change into your PR (maintaining the vm support for future port to self-hosted github runners which do support lxd vms) this would become a helpful contribution to the MicroCeph project.

tomponline:
> 

Yes you should be able to pass a loop block dev into the container as a `disk` see 
https://documentation.ubuntu.com/lxd/en/latest/reference/devices_disk/

UtkarshBhatthere:
With loop file based OSD support and existing cluster tests using lxd containers already present, i am marking this PR closed. Feel free to open it in future with changes or new suggestions.

PR #138, opened 2023-05-26
Added Readme Badges

UtkarshBhatthere:


microceph/database/config.mapper.go
```diff
@@ -110,7 +110,7 @@ func GetConfigItems(ctx context.Context, tx *sql.Tx, filters ...ConfigItemFilter
 	var err error
 
 	// Result slice.
-	objects := make([]ConfigItem, 0)
+	var objects []ConfigItem
```
sabaini, line 113: The *.mapper.go files are generated by lxd-generate, we should avoid editing them if at all possible. I would suggest to report this issue upstream with the lxd project

microceph/database/disk.mapper.go
```diff
@@ -125,7 +125,7 @@ func GetDisks(ctx context.Context, tx *sql.Tx, filters ...DiskFilter) ([]Disk, e
 	var err error
 
 	// Result slice.
-	objects := make([]Disk, 0)
+	var objects []Disk
```
sabaini, line 128: Same as above -- generated code

microceph/database/service.mapper.go
```diff
@@ -133,7 +133,7 @@ func GetServices(ctx context.Context, tx *sql.Tx, filters ...ServiceFilter) ([]S
 	var err error
 
 	// Result slice.
-	objects := make([]Service, 0)
+	var objects []Service
```
sabaini, line 136: Same as above -- generated code

sabaini (changes requested):
Hi @UtkarshBhatthere the badges lgtm but if possible we should avoid updating the generated mapper.go files

microceph/database/config.mapper.go
UtkarshBhatthere: Hmm, that's fair... reversing the changes.

sabaini:
lgtm!

PR #142, opened 2023-06-15
Add contrib guidelines

sabaini:


lmlg (commented):
Looks great overall. The only thing I'd maybe add is a short paragraph about how to report Issues? As in, gather all relevant information, remember to specify version, upload logs if possible, etc. But it's a rather minor thing.

sabaini:
Thanks @lmlg for the suggestion.
With Github issues repo owners can add templates for bug reporting, this probably would be preferable to guide users through creating bug reports. If you're ok with this I'll add a sep. card to create such a template

lmlg:
> Thanks @lmlg for the suggestion. With Github issues repo owners can add templates for bug reporting, this probably would be preferable to guide users through creating bug reports. If you're ok with this I'll add a sep. card to create such a template

absolutely, that seems like the right approach.

lmlg (approved):

sabaini:
Added CEPH-428

PR #143, opened 2023-06-21
Fix config unit testing

sabaini:


microceph/ceph/config_test.go
```diff
@@ -29,9 +29,9 @@ func addConfigSetExpectations(r *mocks.Runner, key string, value string) {
 	}...).Return(value, nil).Once()
 }
 
-func addConfigOpExpectations(r *mocks.Runner, op string, key string, value string) {
+func addConfigOpExpectations(r *mocks.Runner, op, who, key, value string) {
```
lmlg, line 32: any particular reason why the 'op' and 'who' parameters are untyped?

sabaini: @lmlg this is just shorthand because the line was getting long :-) 
in Go you can contract arg types; `func f(a int, b int)` is the same as `func f(a, b int)`

tomponline: FWIW in LXD we tend to avoid that sort of type compaction as we feel it can make it harder to read/review code.

PR #150, opened 2023-06-29
Fix: remove automatic crush rule on upgrading

sabaini:
Remove the osd rule to ensure new pools get the upgraded crush rule


lmlg (approved):
LGTM

Issue #152, opened 2023-07-04
setHostFailureDomain unknown crush rule 'microceph_auto_osd'


sabaini:
User report:

> Hey I'm getting this error very intermittently: Error: Peer "micro03" failed to join the cluster: Failed to update cluster status of services: Failed to join "MicroCeph" cluster: Failed adding new disk: Failed to set host failure domain: Failed to run: ceph osd crush rule dump microceph_auto_osd: exit status 2 (Error ENOENT: unknown crush rule 'microceph_auto_osd') 

When adding several disks we possibly could run a race between setHostFailureDomain and removing the auto crush rule in updateFailureDomain, need to add a safeguard

sabaini:
The patch above should hopefully make the situation better

PR #153, opened 2023-07-04
Fix: make setting host failure domain more resilient

sabaini:
Reduce chance of racing


sabaini:
Should help with issue #152

lmlg (approved):
LGTM

Issue #49, opened 2022-10-20
Incorrect SQL on tables `disks` and `services`


masnax:
These tables should have an `ON DELETE CASCADE` so that we can remove cluster members.

sabaini:
FTR. the symptom is

```
sudo microceph cluster remove -f vm-2
Error: Delete "internal_cluster_members": FOREIGN KEY constraint failed
```

PR #155, opened 2023-07-06
Fix: remove cluster members SQL

sabaini:
Add on delete cascade so foreign key integrity is preserved

Fixes: #49


UtkarshBhatthere (approved):
LGTM

PR #158, opened 2023-07-11
chore: update microcluster dependency

gboutry:
A fix was released in canonical/microcluster@f82e569 solving a deadlock issue that caused timeouts when accessing microcluster apis.


sabaini:
lgtm, thanks!

PR #161, opened 2023-07-13
Service Placement Doc update

UtkarshBhatthere:


docs/how-to/service-placement.rst
```diff
@@ -0,0 +1,71 @@
+Start Ceph services on any node of the MicroCeph cluster
+========================================================
+
+MicroCeph allows users to start supported ceph service on any target host in the MicroCeph cluster. This can be achieved by providing the remote hostname as the --target parameter to the enable command (as described below)
```
sabaini, line 4: I wonder if we should lead in with a sentence about motivation and context, something like:

> MicroCeph will automatically start Mon, Mgr, and Mds services on the first three deployed nodes in the cluster. Users who wish to enable additional instances of these services for scalability or resilience can do so by utilizing the `enable` command and specifying a node via the `--target` parameter.

docs/how-to/service-placement.rst
```diff
@@ -0,0 +1,71 @@
...
+    $ sudo microceph status
+    MicroCeph deployment summary:
+    - node1 (10.111.153.78)
+      Services: mds, mgr, mon, rgw, osd
+      Disks: 3
+    - workbook (192.168.29.152)
+      Services: mds, mgr, mon
+      Disks: 0
+
+2. Enable Mon Service:
```
sabaini, line 54: Nit: suggest to keep service lowercase like the RadosGW entry above (here and below)

sabaini (changes requested):
Two small suggestions, thanks!

docs/how-to/service-placement.rst
UtkarshBhatthere: ackl

docs/how-to/service-placement.rst
```diff
@@ -0,0 +1,71 @@
...
+   :header-rows: 1
+
+   * - Name
+     - Description
+   * - mon
+     - Ceph Monitor Service
+   * - mgr
+     - Ceph Manager Service
+   * - mds
+     - Ceph Monitor Service
```
lmlg, line 17: Should be `metadata daemon service`.

lmlg (approved):
Just  a small nitpick, looks good otherwise :)

docs/how-to/service-placement.rst
UtkarshBhatthere: resolved

PR #165, opened 2023-07-28
Add libtcmalloc to the snap

sabaini:


UtkarshBhatthere:
Thanks for the fix @sabaini! Cheers

PR #167, opened 2023-07-31
Avoid multiple h1 level headings

nobuto-m:
If the following was on purpose, please disregard this PR and close it.

The generated table of contents was as follows.

Tutorial
- Install Microceph using LXD
- Initial Setup
- Add OSDs
- Update management
- Disable automatic updates
- Updating microceph

3 items from the `update.rst` were shown up as the top level topics.

### DIFF
--- a/docs/tutorial/update.rst
+++ b/docs/tutorial/update.rst
@@ -1,5 +1,5 @@
 Update management
-~~~~~~~~~~~~~~~~~
+=================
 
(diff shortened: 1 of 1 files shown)


UtkarshBhatthere:
The updated (and correct) H1 list is as follows:

Tutorial

Install Microceph using LXD
Initial Setup
Add OSDs
Update management

UtkarshBhatthere:
Thanks @nobuto-m

PR #171, opened 2023-08-02
Review service placement howto

pmatulis:


docs/how-to/service-placement.rst
```diff
@@ -1,71 +1,94 @@
...
-   * - mgr
-     - Ceph Manager Service
-   * - mds
-     - Ceph Metadata Daemon Service
-   * - rgw
-     - Ceph Radosgw Service
+* MON (`Monitor service`_)
+* MDS (`Metadata service`_)
+* MGR (`Manager service`_)
+* RGW (`RADOS Gateway service`_)
```
sabaini, line 14: This sounds like MicroCeph will enable 3x RGW services by default, but that is not the case. MON, MDS and MGRs will be started on the first 3 nodes, but RGW always needs to be enabled by the operator

docs/how-to/service-placement.rst
```diff
@@ -1,71 +1,94 @@
...
+   sudo microceph enable rgw --target node1 --port 8080
+
+Finally, view cluster status again and verify expected changes:
+
+.. code-block:: none
+
+   sudo microceph status
+
+   MicroCeph deployment summary:
+   - node1 (10.111.153.78)
```
sabaini, line 82: It seems strange to me that first `microceph status` reports a node `node1-2c3eb41e-14e8-465d-9877-df36f5d80922` above, but here it's just `node1`?

sabaini (changes requested):
Heya, thanks for this -- two comments inline

docs/how-to/service-placement.rst
pmatulis: You're right. Will adjust.

docs/how-to/service-placement.rst
pmatulis: I was using the existing command output. Maybe we keep it for now and when I get around to actual tool testing I can fill this in with realistic output.

pmatulis:
I made a Git error. I will redo this PR.

PR #172, opened 2023-08-02
Multinode testing

sabaini:
Add a GH action runner job to do multi node tests by setting up 4 privileged lxc containers on the runner with loopback devices.

Also refactor tests to utilize a helper script.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,20 +12,16 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
 
       - name: Clear FORWARD firewall rules
-        run: |
-          # Docker can inject rules causing firewall conflicts
-          sudo iptables -P FORWARD ACCEPT  || true
-          sudo ip6tables -P FORWARD ACCEPT || true
-          sudo iptables -F FORWARD  || true
-          sudo ip6tables -F FORWARD || true
+        run: tests/scripts/actionutils.sh cleaript
 
       - name: Install dependencies
         run: |
-          sudo snap refresh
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -199,3 +152,62 @@ jobs:
 
         # Check service restarted
         if [ $ts3 -lt $ts2 ]; then echo "config check failed: TS2: $ts2 TS3: $ts3"; exit 1; fi
+
+  multi-node-tests:
```
UtkarshBhatthere, line 156: The changes look good to me. Thanks for taking this opportunity to refactor the stray bash code from the workflows into a helper script. 
I see that we setup the MicroCeph cluster directly as multi-node however we may also use this to test the automatic crush rule scale up scenario. Tests that verify MicroCeph's ability to fluidly scale up/down would be great as we move more towards cluster level "smartness" in MicroCeph.

sabaini: The crush rule scale up is tested in two places.

- In the single node test above the presence of the auto osd crush rule is tested [here](https://github.com/sabaini/microceph/blob/8911df392c256a3e4862b8fb26c1cbcc6d35d160/.github/workflows/tests.yml#L67)
- In the multi node test the presence of a new host crush rule is tested, also checking the usage in the .mgr pool [here](https://github.com/sabaini/microceph/blob/8911df392c256a3e4862b8fb26c1cbcc6d35d160/.github/workflows/tests.yml#L204)

Do you think we should test more / other things? I'm happy to provide more testing but would prefer to put those in a second PR to keep this PR focused on introducing multi node testing

UtkarshBhatthere: Discussed to add as a separate PR.

PR #176, opened 2023-08-04
Add a test for auto failure domains

sabaini:


lmlg:
Overall LGTM. I'm not a fan of sleep-based synchronization, but it's just a nitpick.

sabaini:
> Overall LGTM. I'm not a fan of sleep-based synchronization, but it's just a nitpick.

Fair. I've replaced some of the longer sleeps with polling, PTAL?

lmlg (approved):

PR #178, opened 2023-08-07
Doc: use ceph alias

sabaini:
MicroCeph has been granted aliases for ceph commands, update the docs to reflect this.


lmlg (approved):
LGTM

PR #181, opened 2023-08-10
Allow access to $HOME for the ceph command

sabaini:
For reading certificates, dumping crushmaps and similar operations

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -97,6 +97,7 @@ apps:
     command: commands/ceph
     plugs:
       - network
+      - home
   "radosgw-admin":
     command: commands/radosgw-admin
     plugs:


UtkarshBhatthere:
Thanks for the PR! Is it absolutely necessary to have $HOME access... Or a good to have for future implementations ?

sabaini:
For various Ceph tools it's required they be able to access the local filesystem, e.g. to configure the dashboard it's required to have it read a cert. Most users would be surprised the tools can't access their $HOME, so following the principle of least surprise we should allow it in my opinion. 

There are workarounds which we could document, i.e. move files to places the snap can access by default, but they're not very user friendly, and in keeping with MicroCephs mission of being low barrier to entry I'd think it better to keep it simple

UtkarshBhatthere:
Sounds and Looks good to me!

Issue #28, opened 2022-09-28
Implement `microceph disk remove`


stgraber:
`microceph disk remove` should be offered as a way to remove an OSD from Ceph and wipe the drive and its configuration.

UtkarshBhatthere:
https://github.com/canonical/microceph/pull/186

PR #186, opened 2023-08-18
OSD removal

sabaini:
Implement `microceph disk remove` command. Update auto failure domain behaviour to also allow downgrading.

Some updates for flaky single node CI tests.


microceph/cmd/microceph/disk_remove.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	cli, err := m.LocalClient()
+	if err != nil {
+		return err
+	}
+
+	osd, err := strconv.ParseInt(args[0], 10, 64)
+	if err != nil {
+		return fmt.Errorf("error converting string to int64: %v", err)
```
wolsen, line 55: I believe this error could be a bit more descriptive as I expect it would get raised to the user. In this case, the OSD Id is invalid and a friendlier error message could be returned. The %v will give srconf.ParseInt: parsing "<foo>": invalid syntax.

microceph/database/disk_extras.go
```diff
@@ -84,3 +123,112 @@ func (m MemberCounterImpl) Count(s *state.State) (int, error) {
...
+`)
+
+var osdPath = cluster.RegisterStmt(`
+SELECT disks.path
+FROM disks
+WHERE disks.OSD = ?
+`)
+
+// HaveOSD returns either false or true depending on whether the given OSD is present in the cluster
+func (o OSDQueryImpl) HaveOSD(s *state.State, osd int64) bool {
```
wolsen, line 150: Should this function signature also return an error?

microceph/ceph/crush.go
```diff
@@ -101,3 +102,46 @@ func setPoolCrushRule(pool string, rule string) error {
...
+	return strings.TrimSpace(configs[0].Value), nil
+}
+
+// ensureCrushRules set up the crush rules for the automatic failure domain handling.
+func ensureCrushRules() error {
+	// Add a microceph default rule with failure domain OSD if it does not exist.
+	if !haveCrushRule("microceph_auto_osd") {
+		err := addCrushRule("microceph_auto_osd", "osd")
+		if err != nil {
+			return fmt.Errorf("Failed to add microceph default rule: %w", err)
```
wolsen, line 136: While this is a refactor, moving the method - I do think its worth s/rule/crush rule/ for avoidance of doubt. Also in the error message below.

microceph/cmd/microceph/disk_remove.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil {
+		return err
+	}
+
+	cli, err := m.LocalClient()
+	if err != nil {
+		return err
+	}
+
+	osd, err := strconv.ParseInt(args[0], 10, 64)
```
wolsen, line 53: If a user is used to Ceph, they may be familiar with using the osd.$id reference rather than just the $id. Should both options be allowed? It's even referred to in the output below as osd.$id.

microceph/ceph/osd.go
```diff
@@ -438,28 +442,353 @@ func AddOSD(s *state.State, path string, wipe bool, encrypt bool) error {
...
+// sanityCheck checks if input is valid
+func sanityCheck(s common.StateInterface, osd int64) error {
+	// check osd is positive
+	if osd < 0 {
+		return fmt.Errorf("OSD must be a positive integer")
+	}
+
+	// check if the OSD exists
+	if !database.OSDQuery.HaveOSD(s.ClusterState(), osd) {
+		return fmt.Errorf("OSD %d not found", osd)
```
wolsen, line 473: We should probably consistently refer to OSDs throughout. In some cases its OSD $id, in other cases its osd.$id. I'd favor the osd.$id as its likely more familiar to users of Ceph and lines up Ceph docs and such.

microceph/ceph/osd.go
```diff
@@ -438,28 +442,353 @@ func AddOSD(s *state.State, path string, wipe bool, encrypt bool) error {
...
+		logger.Infof("Purge failed %v, retrying in %v", err, backoff)
+		backoff = time.Duration(math.Pow(2, float64(i))) * time.Millisecond * 100
+		time.Sleep(backoff)
+	}
+
+	if err != nil {
+		logger.Errorf("Failed to purge osd.%d: %v", osd, err)
+		return fmt.Errorf("Failed to purge osd.%d: %w", osd, err)
+	}
+	logger.Infof("OSD %v purged", osd)
```
wolsen, line 581: Consistency for the osd.$id information.

microceph/ceph/osd.go
```diff
@@ -438,28 +442,353 @@ func AddOSD(s *state.State, path string, wipe bool, encrypt bool) error {
...
-		return nil
-	})
+func checkMinOSDs(s common.StateInterface) error {
+	// check if we have at least 3 OSDs post-removal
+	disks, err := database.OSDQuery.List(s.ClusterState())
+	if err != nil {
+		return err
+	}
+	if len(disks) <= 3 {
+		return fmt.Errorf("Cannot remove osd we need at least 3 OSDs, have %d", len(disks))
```
wolsen, line 706: Suggestion to add in the $id part in the error message

microceph/ceph/osd.go
```diff
@@ -438,28 +442,353 @@ func AddOSD(s *state.State, path string, wipe bool, encrypt bool) error {
...
+	_, err := processExec.RunCommand("ceph", "osd", "out", fmt.Sprintf("osd.%d", osd))
+	if err != nil {
+		logger.Errorf("Failed to take OSD out: %v", err)
+		return fmt.Errorf("Failed to take OSD out: %w", err)
+	}
+	_, err = processExec.RunCommand("ceph", "osd", "down", fmt.Sprintf("osd.%d", osd))
 	if err != nil {
-		return nil, err
+		logger.Errorf("Failed to take OSD down: %v", err)
+		return fmt.Errorf("Failed to take OSD down: %w", err)
```
wolsen, line 720: suggestion: s/OSD/osd.$id/

microceph/ceph/osd.go
```diff
@@ -438,28 +442,353 @@ func AddOSD(s *state.State, path string, wipe bool, encrypt bool) error {
...
+		return fmt.Errorf("Cannot remove osd we need at least 3 OSDs, have %d", len(disks))
+	}
+	return nil
+}
+
+func outDownOSD(osd int64) error {
+	_, err := processExec.RunCommand("ceph", "osd", "out", fmt.Sprintf("osd.%d", osd))
+	if err != nil {
+		logger.Errorf("Failed to take OSD out: %v", err)
+		return fmt.Errorf("Failed to take OSD out: %w", err)
```
wolsen, line 715: suggestion: s/OSD/osd.$id/

microceph/ceph/osd.go
```diff
@@ -438,28 +442,353 @@ func AddOSD(s *state.State, path string, wipe bool, encrypt bool) error {
...
+	// Remove osd config
+	err = removeOSDConfig(osd)
+	if err != nil {
+		return err
+	}
+	// Remove db entry
+	err = database.OSDQuery.Delete(s.ClusterState(), osd)
+	if err != nil {
+		logger.Errorf("Failed to remove OSD from database: %v", err)
+		return fmt.Errorf("Failed to remove OSD from database: %w", err)
```
wolsen, line 694: s/OSD/osd.$id/

microceph/api/disks.go
```diff
@@ -44,3 +56,45 @@ func cmdDisksPost(s *state.State, r *http.Request) response.Response {
...
+
+	var req types.DisksDelete
+	osdid, err := strconv.ParseInt(osd, 10, 64)
+
+	if err != nil {
+		return response.BadRequest(err)
+	}
+	err = json.NewDecoder(r.Body).Decode(&req)
+	if err != nil {
+		return response.InternalError(err)
```
wolsen, line 76: Same as above - I think this is a client error and not an internal server side error when the body cannot be JSON decoded.

microceph/api/disks.go
```diff
@@ -44,3 +56,45 @@ func cmdDisksPost(s *state.State, r *http.Request) response.Response {
 
 	return response.EmptySyncResponse
 }
+
+// cmdDisksDelete is the handler for DELETE /1.0/disks/{osdid}.
+func cmdDisksDelete(s *state.State, r *http.Request) response.Response {
+	var osd string
+	osd, err := url.PathUnescape(mux.Vars(r)["osdid"])
+	if err != nil {
+		return response.InternalError(err)
```
wolsen, line 65: Should this be a BadRequest error rather than an InternalError? The consumer of this REST api would know they've provided a malformed request of some sort. It may end up in an InternalError on the client side, etc. But I think the correct response is a BadRequest here.

microceph/ceph/osd.go
```diff
@@ -438,28 +442,353 @@ func AddOSD(s *state.State, path string, wipe bool, encrypt bool) error {
...
+func reweightOSD(ctx context.Context, osd int64, weight float64) {
+	logger.Debugf("Reweighting osd.%d to %f", osd, weight)
+	_, err := processExec.RunCommand(
+		"ceph", "osd", "crush", "reweight",
+		fmt.Sprintf("osd.%d", osd),
+		fmt.Sprintf("%f", weight),
+	)
+	if err != nil {
+		// only log a warn, don't treat fail to reweight as a fatal error
+		logger.Warnf("Failed to reweight OSD %d: %v", osd, err)
```
wolsen, line 534: s/OSD %d/osd.$d/

wolsen (changes requested):
Some inline comments throughout. One thing that remains unclear to me is if you have a partial failure of removing a disk, how does the software behave? Will running the command again pick up where it left off and proceed?

I believe there's a scenario with the current code in which the OSD can be removed from the Ceph cluster, but fail to get removed from the database. I don't think that the code allows for the OSD to still be removed from the database in this scenario as commands like reweightOSD, outDownOSD, etc will return an ENOENT when running the ceph commands.

lmlg:
> Some inline comments throughout. One thing that remains unclear to me is if you have a partial failure of removing a disk, how does the software behave? Will running the command again pick up where it left off and proceed?
> 
> I believe there's a scenario with the current code in which the OSD can be removed from the Ceph cluster, but fail to get removed from the database. I don't think that the code allows for the OSD to still be removed from the database in this scenario as commands like reweightOSD, outDownOSD, etc will return an ENOENT when running the ceph commands.

Personally, I think there should be a "point of no return", upon which there may be errors, but from where the code cannot pick up. In the case of the DB failing to remove the entry, that should only be possible if something is _really_ wrong (like filesystem corruption on the DB files). In case of such an error, I think it would be crucial to let the user know that the OSD is out of the cluster, but that manual intervention may be needed.

wolsen:
> > Some inline comments throughout. One thing that remains unclear to me is if you have a partial failure of removing a disk, how does the software behave? Will running the command again pick up where it left off and proceed?
> > I believe there's a scenario with the current code in which the OSD can be removed from the Ceph cluster, but fail to get removed from the database. I don't think that the code allows for the OSD to still be removed from the database in this scenario as commands like reweightOSD, outDownOSD, etc will return an ENOENT when running the ceph commands.
> 
> Personally, I think there should be a "point of no return", upon which there may be errors, but from where the code cannot pick up. In the case of the DB failing to remove the entry, that should only be possible if something is _really_ wrong (like filesystem corruption on the DB files). In case of such an error, I think it would be crucial to let the user know that the OSD is out of the cluster, but that manual intervention may be needed.

What do you propose as the point of no return? There's several steps involved, and I think most all of them are reasonable to have a check when performed to see if they need to be performed. They are:

scaleDownFailureDomain
reweightOSD
(safetyCheckStop)
downOSD
stopOSD
(safetyCheckDestroy)
purgeOSD
wipeDevice
removeOSD
deleteFromDatabase

Once you start the purge of the OSD is where you will have challenges continuing forth, though the necessary information should still be in the dqlite database. But if purge, wipe or remove fail (which is more likely during the event of a disk failure) then you won't be able to delete the OSD from the database without running database commands directly.

lmlg:
> > > Some inline comments throughout. One thing that remains unclear to me is if you have a partial failure of removing a disk, how does the software behave? Will running the command again pick up where it left off and proceed?
> > > I believe there's a scenario with the current code in which the OSD can be removed from the Ceph cluster, but fail to get removed from the database. I don't think that the code allows for the OSD to still be removed from the database in this scenario as commands like reweightOSD, outDownOSD, etc will return an ENOENT when running the ceph commands.
> > 
> > 
> > Personally, I think there should be a "point of no return", upon which there may be errors, but from where the code cannot pick up. In the case of the DB failing to remove the entry, that should only be possible if something is _really_ wrong (like filesystem corruption on the DB files). In case of such an error, I think it would be crucial to let the user know that the OSD is out of the cluster, but that manual intervention may be needed.
> 
> What do you propose as the point of no return? There's several steps involved, and I think most all of them are reasonable to have a check when performed to see if they need to be performed. They are:
> 
> scaleDownFailureDomain reweightOSD (safetyCheckStop) downOSD stopOSD (safetyCheckDestroy) purgeOSD wipeDevice removeOSD deleteFromDatabase
> 
> Once you start the purge of the OSD is where you will have challenges continuing forth, though the necessary information should still be in the dqlite database. But if purge, wipe or remove fail (which is more likely during the event of a disk failure) then you won't be able to delete the OSD from the database without running database commands directly.

I'd argue the last step - the DB removal - should be the point of no return, because any other failure can be worked around via command flags (i.e: the 'force' flag). If they can't be worked around, I'd argue it's a design defect. Operating on the DB is the place where I would draw the line because it requires not just operator-level knowledge about Ceph, but about Microceph's internals.

wolsen:
> > Once you start the purge of the OSD is where you will have challenges continuing forth, though the necessary information should still be in the dqlite database. But if purge, wipe or remove fail (which is more likely during the event of a disk failure) then you won't be able to delete the OSD from the database without running database commands directly.
> 
> I'd argue the last step - the DB removal - should be the point of no return, because any other failure can be worked around via command flags (i.e: the 'force' flag). If they can't be worked around, I'd argue it's a design defect. Operating on the DB is the place where I would draw the line because it requires not just operator-level knowledge about Ceph, but about Microceph's internals.

I agree with this particular point. I think that the database removal is a fatal error and requires fixing the database.

However, my primary point is that if the removal of the device fails due to say I/O error or timeout at the wipe phase. At this point, the remove-disk command cannot be run again because the actions like reweight of the OSD will return an ENOENT error, which will cause the command to fail and the user is then unable to proceed further, much less remove the device from the database.

UtkarshBhatthere:
https://github.com/canonical/microceph/issues/28

sabaini:
Thanks for all the valuable input. I've pushed an update, hopefully adressing comments.

Wrt. to failure modes and robustness, I've updated semantics in the following way:

- add a guard against operating on OSDs that are not actually present in the cluster
-  change wiping so that a device fault is not fatal anymore

Let me know what you think

microceph/ceph/osd.go
```diff
@@ -438,28 +439,413 @@ func AddOSD(s *state.State, path string, wipe bool, encrypt bool) error {
...
+}
+
+// timeoutWipe wipes the given device with a timeout, in order not to hang on broken disks
+func timeoutWipe(path string) error {
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	_, err := processExec.RunCommandContext(
+		ctx,
+		"dd", "if=/dev/zero",
```
lmlg, line 620: I know this is unrelated to this change, but I'm not really a fan of this way of wiping a disk. I much prefer `wipefs -a -f ${path}` since it can catch some occasions where the disk may appear 'broken', but that's just a personal preference.

microceph/ceph/osd.go
```diff
@@ -438,28 +439,413 @@ func AddOSD(s *state.State, path string, wipe bool, encrypt bool) error {
...
+		}
+	}
+
+	err = scaleDownFailureDomain(s, osd)
+	if err != nil {
+		return err
+	}
+
+	// check if the osd is still in the cluster -- if we're being re-run, it might not be
+	isPresent, err := haveOSDInCeph(osd)
```
lmlg, line 650: This should probably be moved a bit more up; there's no reason to scale down _before_ checking that the OSD exists in the DB, IMO.

microceph/ceph/osd.go
```diff
@@ -438,28 +439,413 @@ func AddOSD(s *state.State, path string, wipe bool, encrypt bool) error {
...
+	// purge the OSD
+	if isPresent {
+		err = purgeOSD(osd)
+		if err != nil {
+			return err
 		}
+	}
+	// Wipe the underlying blocking device
+	err = wipeDevice(s, osd)
+	if err != nil {
```
lmlg, line 695: Is this an error, though? If the disk is broken, it doesn't seem like it matters that much at this point. I'd simply log it so that the operator is aware (which they probably are, if we're being fair).

microceph/ceph/osd.go
```diff
@@ -438,28 +439,413 @@ func AddOSD(s *state.State, path string, wipe bool, encrypt bool) error {
...
+		return err
+	}
+	if len(disks) <= 3 {
+		return fmt.Errorf("Cannot remove osd.%d we need at least 3 OSDs, have %d", osd, len(disks))
+	}
+	return nil
+}
+
+func outDownOSD(osd int64) error {
```
lmlg, line 724: Is this function necessary? I think if you reweight the OSD to 0 and `ok-to-stop` and `safe-to-destroy` return true, then the OSD can be safely removed; it doesn't need to be marked as out, as far as I recall. At least, that's how we do it in the OSD charm.

lmlg (changes requested):

microceph/ceph/osd.go
sabaini: I've thought about going with wipefs but the downside is you need to pull in a new package into the snap, whereas `dd` is already on board, that's why I opted to keep it

microceph/ceph/osd.go
sabaini: The scale down is designed to be independent of any individual osd, it will determine the need to downgrade solely based on cluster state, so it should not matter if it comes before or after the presence check. 
I have opted to run the check for downscaling before the presence check for robustness, if for some reason (e.g. we're being re-run and there was some fault previously) the OSD already was taken out, we get another chance to downscale

microceph/ceph/osd.go
sabaini: Ack, pushed an update to remove this

microceph/ceph/osd.go
sabaini: In my testing the OSD removal was much more robust if you took out the OSD explicitly (we might want to consider doing it for the charm as well)

sabaini:
Hola @lmlg, I've pushed an update an answered some comments, please take a look :-)

microceph/ceph/osd.go
```diff
@@ -438,28 +439,407 @@ func AddOSD(s *state.State, path string, wipe bool, encrypt bool) error {
...
+			break
+		}
+		if syscall.Errno(exitError.ExitCode()) != syscall.EBUSY {
+			// not a busy error, abort and bubble up the error
+			logger.Warnf("Purge failed with unexpected exit error: %v", exitError)
+			break
+		}
+		// purge failed with EBUSY - retry after a delay, and make delay exponential
+		logger.Infof("Purge failed %v, retrying in %v", err, backoff)
+		backoff = time.Duration(math.Pow(2, float64(i))) * time.Millisecond * 100
```
lmlg, line 573: It's a nitpick, but I frankly don't see an exponential backoff as something useful here. This whole function is called under a timeout, so the backoff can actually lead to some retries not being attempted. I would use a fixed-time sleep, but it's just a personal preference.

microceph/ceph/osd.go
```diff
@@ -438,28 +439,407 @@ func AddOSD(s *state.State, path string, wipe bool, encrypt bool) error {
...
+	if isPresent {
+		err = purgeOSD(osd)
+		if err != nil {
+			return err
+		}
+	}
+	// Wipe the underlying blocking device
+	wipeDevice(s, osd)
+	// Remove osd config
+	err = removeOSDConfig(osd)
```
lmlg, line 693: is this going to succeed if the OSD isn't present?

microceph/ceph/osd.go
```diff
@@ -438,28 +439,407 @@ func AddOSD(s *state.State, path string, wipe bool, encrypt bool) error {
...
+	if err != nil {
+		return err
+	}
+	// Remove db entry
+	err = database.OSDQuery.Delete(s.ClusterState(), osd)
```
lmlg, line 698: Ditto for the above question.

lmlg (commented):
LGTM, just a few questions overall.

microceph/ceph/osd.go
sabaini: Not a must-have IMO but I'd lean towards keeping it in as on a busy system it can give the cluster a bit more breathing room and keeps the logs somewhat cleaner (and it doesn't cost much to implement). Wrt to the timeout, this is user-configurable, so if the timeout kicks in we don't control

microceph/ceph/osd.go
sabaini: removeOSDConfig() removes the local disk config. It will succeed even if the OSD has been removed from the cluster previously i.e. if the OSD is gone from the cluster as well as if the config dir is ENOENT

lmlg (approved):
LGTM!

lmlg (approved):

PR #193, opened 2023-08-24
Use appropriate {public,cluster}_network examples

nobuto-m:
{public,cluster}_network values passed to microceph will be written to the Ceph config database as {public,cluster}_network as is. That means the values are used as cluster wide setting instead of per-node config.

According to the upstream configuration reference[1], the previous example doesn't make sense as 10.5.2.165 looks like a real IP address of a node instead of a network address. Let's use a network address and be clearer about the reset step too.

```
$ sudo microceph cluster config set cluster_network 192.168.102.0/24

$ sudo microceph.ceph config dump
WHO     MASK  LEVEL     OPTION           VALUE             RO
global        advanced  cluster_network  192.168.102.0/24  *
```

[1] https://docs.ceph.com/en/latest/rados/configuration/network-config-ref/#confval-cluster_network

### DIFF
--- a/docs/how-to/cluster-cfg.rst
+++ b/docs/how-to/cluster-cfg.rst
@@ -18,44 +18,43 @@ The microceph cluster configuration CLI supports setting, getting, resetting and
 
   .. code-block:: shell
 
-    $ sudo microceph cluster config set cluster_network 10.5.2.165/16
+    $ sudo microceph cluster config set cluster_network 10.5.0.0/16
 
 2. Config value for a particular key could be queried using the 'get' command:
 
   .. code-block:: shell
 
     $ sudo microceph cluster config get cluster_network
-    +---+-----------------+---------------+
-    | # |       KEY       |     VALUE     |
-    +---+-----------------+---------------+
-    | 0 | cluster_network | 10.5.2.165/16 |
-    +---+-----------------+---------------+
+    +---+-----------------+-------------+
+    | # |       KEY       |     VALUE   |
+    +---+-----------------+-------------+
+    | 0 | cluster_network | 10.5.0.0/16 |
+    +---+-----------------+-------------+
 
 3. A list of all the configured keys can be fetched using the 'list' command:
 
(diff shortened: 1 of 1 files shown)


docs/how-to/cluster-cfg.rst
```diff
@@ -18,44 +18,43 @@ The microceph cluster configuration CLI supports setting, getting, resetting and
...
+    | 0 | cluster_network | 10.5.0.0/16 |
+    +---+-----------------+-------------+
+    | 1 | public_network  | 10.5.0.0/16 |
+    +---+-----------------+-------------+
 
 4. Resetting a config key (i.e. setting the key to its default value) can performed using the 'reset' command:
 
   .. code-block:: shell
 
+   $ sudo microceph cluster config reset cluster_network
```
UtkarshBhatthere, line 52: Is there any specific reason for cleaning both cluster/public network config ? The original intention was to suggest that these 2 configs are independent and need not be set/reset together.

UtkarshBhatthere (commented):
The IP range for subnets 10.5.2.165/16 and 10.5.0.0/16 would be same, however, I do agree that 10.5.0.0/16 is a better example for a reader. Thanks for the PR, I have one question as commented.

docs/how-to/cluster-cfg.rst
nobuto-m: When I read the doc in the first run, I interpreted it as setting "10.5.2.165/16" as the custom network then resetting it. I was confused for one minutes finding the default value from the command output as in "setting the key to its default value". Because the only visible value in the output is still "10.5.2.165/16".

I thought it would be a good idea to show the empty table to be clean that the default value means no entry in the Ceph config database.

UtkarshBhatthere: Aaah, that makes sense. Thanks for the explanation.

UtkarshBhatthere:
Thanks for the contribution @nobuto-m. Kampai!

PR #209, opened 2023-09-06
Document OSD removal

sabaini:


pmatulis:
Thanks for this doc contribution. :confetti_ball: 

Let's merge this as-is. I will then do a run-through of the steps and submit a separate review PR in the near future.

Issue #190, opened 2023-08-23
Error: Failed adding new disk: Failed to open: Failed to luksOpen: /dev/disk/by-path/..., exit status 2, Failed to load key in kernel keyring. No key available with this passphrase.


nobuto-m:
When I follow the FDE document:
https://canonical-microceph.readthedocs-hosted.com/en/latest/explanation/fde-osd/

It fails with the following output.

```
ubuntu@mc-1:~$ sudo microceph disk add /dev/vdc --wipe --encrypt
Error: Failed adding new disk: Failed to open: Failed to luksOpen: /dev/disk/by-path/virtio-pci-0000:08:00.0, exit status 2, Failed to load key in kernel keyring.
No key available with this passphrase.


NOTE: OSD Encryption requires a snapd >= 2.59.1
Verify your version of snapd by running "snap version"

ubuntu@mc-1:~$ snap version
snap    2.60.1+git1186.g661258f
snapd   2.60.1+git1186.g661258f
series  16
ubuntu  22.04
kernel  5.15.0-79-generic
```
```
$ snap list microceph
Name       Version        Rev  Tracking     Publisher   Notes
microceph  0+git.c07ce73  585  latest/edge  canonical✓  -
```

sabaini:
Hi @nobuto-m ,

Apologies for the late reply.

I couldn't immediately repro this issue. Would you be able to share some more details:

- Any apparmor or seccomp messages in `dmesg`? It would be ideal to get a `sosreport --all-logs` report
- I presume you have run `sudo snap connect microceph:dm-crypt`, is that right?
- Can you add this disk without encryption, i.e. run `sudo microceph disk add /dev/vdc --wipe`?

Thanks in advance.

nobuto-m:
Sure thing. I have the following list of snaps installed on 3 machines and the microceph is behind the microcloud cluster.

```
$ snap list
Name        Version                  Rev    Tracking       Publisher   Notes
core20      20230622                 1974   latest/stable  canonical✓  base
core22      20230801                 864    latest/stable  canonical✓  base
lxd         5.17-e5ead86             25505  latest/stable  canonical✓  -
microceph   0+git.a77ce2c            627    latest/edge    canonical✓  -
microcloud  0+git.091ee78            540    latest/edge    canonical✓  -
microovn    22.03.2+snap4a6fea39cd   244    22.03/stable   canonical✓  -
snapd       2.60.3+git1281.g75a2072  20184  latest/edge    canonical✓  snapd
[...]

PR #211, opened 2023-09-06
Doc: need to restart microceph.daemon after dm-crypt conn

sabaini:
This is a stop gap measure for #190. There is an issue to track this at https://bugs.launchpad.net/snapd/+bug/2034585 . For a more permanent workaround we could alternatively also restart microceph.daemon in an interface hook after dm-crypt connection, however it'd be preferable to fix this in snapd so as to avoid API interruptions.


lmlg (approved):
LGTM, although I'm not sure whether we should mention this is a "temporary" workaround or not.

nobuto-m:
I think the following part needs the same change.
https://github.com/canonical/microceph/blob/cecb46e377d833538ce76f42e3387f05b2e7200b/microceph/ceph/osd.go#L206

PR #213, opened 2023-09-07
Fix disk race

### LINKED ISSUE #202: Race: adding disks, ceph-osd restart
Adding a disk causes a restart of the snap.microceph.osd service. These are not synchronous so when adding disks in parallel it can happen that a snap.microceph.osd restart is initiated even though a previous one is already running, resulting in an error from snapctl:

```
Error: Failed adding new disk: Failed to run: snapctl restart --reload microceph.osd: exit status 1 (error: error running snapctl: snap "microceph" has "service-control" change in progress)
```

We should look into making those service restarts sync

sabaini:
Also improve logging, and turn on debug logging by default as we do not have a good way of controlling loglevel yet

Closes #202

### DIFF
--- a/microceph/api/disks.go
+++ b/microceph/api/disks.go
@@ -3,11 +3,13 @@ package api
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/canonical/lxd/shared/logger"
 	"github.com/canonical/microceph/microceph/common"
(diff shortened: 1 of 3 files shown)


lmlg (approved):
LGTM

PR #215, opened 2023-09-11
Review remove osd howto 2

pmatulis:
This PR kicked up an awful lot of dust. Going forward the structure presented here will assist the team. Take note of the way in which reference, howto, and conceptual material are working in concert here.

Noteworthy:

* A reference command section was introduced. The `disk remove` contents attempted to stay closely aligned to the CLI help. It would be nice if the latter could use the term "option" over "flag" as this is the most common terminology, at least within the OpenStack org.

* I found the title of **Removing OSDs** to be out of alignment with the command `disk remove` so I switched to **Remove a disk**. I also find that "disk" is easier for the uninitiated.

* The conceptual page **Scaling** still needs a fair amount of work. Only changes were made that were necessary to accommodate the rest of this PR (besides renaming the page).


docs/how-to/remove-disk.rst
```diff
@@ -0,0 +1,131 @@
...
+       osd: 4 osds: 4 up (since 4m), 4 in (since 4m)
+
+     data:
+       pools:   1 pools, 1 pgs
+       objects: 2 objects, 577 KiB
+       usage:   68 MiB used, 991 GiB / 992 GiB avail
+       pgs:     1 active+clean
+
+.. LINKS
+.. _CRUSH map: https://docs.ceph.com/en/latest/rados/operations/crush-map/
```
sabaini, line 131: Is this link still needed anywhere?

sabaini (changes requested):
Hi, thanks for this! 
Lgtm in general -- one minor nit inline

sabaini (approved):

PR #216, opened 2023-09-12
Unpublish current tutorial

pmatulis:
Unpublish the current multi-page tutorial and replace it with the two recently added tutorials (single-node and multi-node).

The old content has been moved - not deleted. It may be useful in the future. Essential commands:

```
git mv tutorial tutorial.bak
git mv tutorial-2 tutorial
```


sabaini:
Hi @pmatulis , 
Is there a specific reason you'd want to keep the previous tutorial around in-tree?
I'd tend to just delete the previous tutorial instead of keeping it in-tree as a backup for cleanliness. We always have the git history to refer back to in case of need

sabaini:
Thanks!

PR #222, opened 2023-09-21
Add the WAL and DB devices for the disk-add command

lmlg:
This PR aims to add the needed support for write-ahead log (WAL) and database devices when adding OSDs to the cluster.

### DIFF
--- a/microceph/api/disks.go
+++ b/microceph/api/disks.go
@@ -49,6 +49,7 @@ func cmdDisksPost(s *state.State, r *http.Request) response.Response {
 	var req types.DisksPost
 	var wal *types.DiskParameter
 	var db *types.DiskParameter
+	var data types.DiskParameter
 
 	logger.Debugf("cmdDisksPost: %v", req)
 	err := json.NewDecoder(r.Body).Decode(&req)
@@ -59,12 +60,12 @@ func cmdDisksPost(s *state.State, r *http.Request) response.Response {
 	mu.Lock()
 	defer mu.Unlock()
 
-	data := types.DiskParameter{req.Path, req.Wipe, req.Encrypt}
+	data = types.DiskParameter{req.Path, req.Encrypt, req.Wipe}
 	if req.WALDev != nil {
-		wal = &types.DiskParameter{*req.WALDev, req.WALWipe, req.WALEncrypt}
+		wal = &types.DiskParameter{*req.WALDev, req.WALEncrypt, req.WALWipe}
 	}
 	if req.DBDev != nil {
-		db = &types.DiskParameter{*req.DBDev, req.DBWipe, req.DBEncrypt}
+		db = &types.DiskParameter{*req.DBDev, req.DBEncrypt, req.DBWipe}
 	}
 
 	err = ceph.AddOSD(s, data, wal, db)
--- a/microceph/api/disks.go
+++ b/microceph/api/disks.go
@@ -56,7 +56,7 @@ func cmdDisksPost(s *state.State, r *http.Request) response.Response {
 
 	mu.Lock()
 	defer mu.Unlock()
-	err = ceph.AddOSD(s, req.Path, req.Wipe, req.Encrypt)
+	err = ceph.AddOSD(s, req.Path, req.Wipe, req.Encrypt, req.WALDev, req.DBDev)
 	if err != nil {
 		return response.SmartError(err)
 	}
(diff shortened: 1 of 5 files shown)


microceph/ceph/osd.go
```diff
@@ -264,25 +289,20 @@ func updateFailureDomain(s *state.State) error {
 	return nil
 }
 
-// AddOSD adds an OSD to the cluster, given a device path and a flag for wiping
-func AddOSD(s *state.State, path string, wipe bool, encrypt bool) error {
-	logger.Debugf("Adding OSD %s", path)
+// AddOSD adds an OSD to the cluster, given the data, WAL and DB devices and their respective
+// flags for wiping and encrypting.
+func AddOSD(s *state.State, data types.DiskParameter, wal *types.DiskParameter, db *types.DiskParameter) error {
```
sabaini, line 294: Hm, is there a reason to pass `data` as by value and `wal` / `db` by ref?

microceph/ceph/osd.go
```diff
@@ -264,25 +289,20 @@ func updateFailureDomain(s *state.State) error {
...
-	if encrypt {
-		if err := checkEncryptSupport(); err != nil {
-			return fmt.Errorf("Encryption unsupported on this machine: %w", err)
-		}
+	if !shared.IsBlockdevPath(data.Path) {
+		return fmt.Errorf("Invalid disk path: %s", data.Path)
 	}
 
-	_, _, major, minor, _, _, err := shared.GetFileStat(path)
+	_, _, major, minor, _, _, err := shared.GetFileStat(data.Path)
```
sabaini, line 305: We probably should look up a stable path for `wal.Path` and `db.Path` as well, like we do for `data.Path` here and further below. Might make sense to factor the lookup algo into a sep. function for this

sabaini (changes requested):
Hi Luciano,
one request wrt looking up stable paths for WAL/DB devices (sorry for not flagging this earlier), otherwise lgtm

microceph/api/disks.go
```diff
@@ -47,6 +47,8 @@ func cmdDisksGet(s *state.State, r *http.Request) response.Response {
 
 func cmdDisksPost(s *state.State, r *http.Request) response.Response {
 	var req types.DisksPost
+	var wal *types.DiskParameter
+	var db *types.DiskParameter
```
UtkarshBhatthere, line 51: For consistency I'd suggest we also declare "data" here.

microceph/api/disks.go
```diff
@@ -56,7 +58,16 @@ func cmdDisksPost(s *state.State, r *http.Request) response.Response {
 
 	mu.Lock()
 	defer mu.Unlock()
-	err = ceph.AddOSD(s, req.Path, req.Wipe, req.Encrypt)
+
+	data := types.DiskParameter{req.Path, req.Encrypt, req.Wipe}
```
UtkarshBhatthere, line 62: NIT: "data" definition should be consistent with wal/db object initialisation.

microceph/ceph/osd.go
```diff
@@ -80,12 +80,37 @@ func nextOSD(s *state.State) (int64, error) {
...
+		if err != nil {
+			return fmt.Errorf("Encryption unsupported on this machine: %w", err)
+		}
+		path, err := setupEncryptedOSD(disk.Path, osdPath, osdID, suffix)
+		if err != nil {
+			return fmt.Errorf("Failed to encrypt device %s: %w", disk.Path, err)
+		}
+		disk.Path = path
+	}
+	if suffix != "" {
```
UtkarshBhatthere, line 101: A comment describing why execution can return without "symlinking" (at L104) for non-null suffixes could be helpful.

UtkarshBhatthere (changes requested):
Few nitpicks around consistency and code-comments.

microceph/ceph/osd.go
lmlg: I didn't do it because Ceph itself will manage everything related to the wal and db paths (i.e: symlinking and all that), but yeah, it can be done.

microceph/api/disks.go
lmlg: Data can never be nil, hence why it's not a pointer.

sabaini (approved):

sabaini (commented):
Hi @lmlg , changes lgtm!
Think the [WIP] tag could be removed now :-)
May I also suggest to squash commits (into one or more logical units)?

lmlg:
> Hi @lmlg , changes lgtm! Think the [WIP] tag could be removed now :-) May I also suggest to squash commits (into one or more logical units)?

yes, of course.

UtkarshBhatthere:
Note: CI tests for this change will be taken up in a separate PR.

PR #224, opened 2023-09-22
Adds unit-test to microceph CI

UtkarshBhatthere:


.github/workflows/tests.yml
```diff
@@ -36,6 +36,29 @@ jobs:
...
+    name: Run Microceph Unit tests
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - name: Clear FORWARD firewall rules
+        run: tests/scripts/actionutils.sh cleaript
```
sabaini, line 49: Minor comment, but these shouldn't impede the unit tests and could be left in

sabaini (changes requested):
Hi Utkarsh, thanks! One minor request inline but otherwise lgtm

.github/workflows/tests.yml
UtkarshBhatthere: makes sense, will change accordingly.

sabaini (approved):

PR #226, opened 2023-09-26
Add upgrade tests

sabaini:
Add a test to run an upgrade from quincy/stable to local build on push

Also add a manual test to run a test from quincy/stable to quincy/candidate on demand

Some test refactoring


lmlg (approved):
LGTM

PR #235, opened 2023-10-09
Fixes for disk remove issues

sabaini:
Fix regexp for stopping OSD services.

Don't treat a failure to stop an OSD service as fatal, might be down already.

Increase default timeout and retries, the default was overly optimistic.

Test cleanup.


lmlg (approved):
LGTM

PR #238, opened 2023-10-13
docs: Review page titles and filenames

pmatulis:
Improve page titles and underlying filenames.

### DIFF
--- a/docs/tutorial/single-node.rst
+++ b/docs/tutorial/single-node.rst
@@ -1,6 +1,6 @@
-===========
-Single-node
-===========
+===================
+Single-node install
+===================
 
 This tutorial will show how to install MicroCeph on a single machine, thereby
 creating a single-node "cluster".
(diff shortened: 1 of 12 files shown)


docs/tutorial/single-node.rst
```diff
@@ -13,7 +13,7 @@ three entire disks are required to be available on the machine.
 
 .. note::
 
-   Development is underway to allow for loopback device support.
+   Upstream Ceph development is underway to allow for loopback device support.
```
sabaini, line 16: The dev tasks referred to here need to be done in MicroCeph, from an upstream Ceph POV there is nothing to be done

sabaini (changes requested):
Hi, one small issue, otherwise lgtm

docs/tutorial/single-node.rst
pmatulis: Okay, I was thrown by "this will filter down to MicroCeph". If the work is being done in MicroCeph, what is meant by "filter down"?

sabaini: Ah, good observation. I believe this was meant to refer to a dependency in snapd (which has by now been resolved though)

sabaini (approved):

PR #241, opened 2023-10-16
Implement node removal

sabaini:


microceph/cmd/microceph/cluster_remove.go
```diff
@@ -36,15 +41,123 @@ func (c *cmdClusterRemove) Run(cmd *cobra.Command, args []string) error {
...
+	err = client.MClient.DeleteClusterMember(cli, node, force)
+	logger.Debugf("DeleteClusterMember %v: %v", node, err)
 	if err != nil {
 		return err
 	}
 
 	return nil
 }
+
+func checkPrerequisites(cli *microCli.Client, name string) (bool, error) {
```
lmlg, line 87: The boolean returned by this function isn't too useful, imo. The returned error already provides the needed information, and checking for both the error _and_ the bool makes things more complex than they should.

lmlg (changes requested):
LGTM, just a small nit.

microceph/cmd/microceph/cluster_remove.go
sabaini: Ack

UtkarshBhatthere (approved):
LGTM

PR #246, opened 2023-10-18
Doc: cluster commands

sabaini:


docs/reference/commands/cluster.rst
```diff
@@ -0,0 +1,210 @@
+===========
+``cluster``
+===========
+
+Manage the MicroCeph cluster
```
pmatulis, line 5: Let's use this as a model to be applied everywhere:

**Manages the MicroCeph cluster.**

So there's an 's' at the end of the first word and a '.' at the end of the last word.

docs/reference/commands/cluster.rst
```diff
@@ -0,0 +1,210 @@
...
+
+Usage:
+
+.. code-block:: none
+
+   microceph cluster [options]
```
pmatulis, line 11: We use 'options' here but 'flags' elsewhere. Let's stick with 'options' everywhere.

docs/reference/commands/cluster.rst
```diff
@@ -0,0 +1,210 @@
...
+``add``
+-------
+
+Generates a token for a new server
+
+Usage:
+
+.. code-block:: none
+
+  microceph cluster add <NAME> [flags]
```
pmatulis, line 47: Let's align the first command string directly under 'code'.

So here, 'm' should be directly under 'c'.

Let's do this everywhere.

docs/reference/commands/cluster.rst
```diff
@@ -0,0 +1,210 @@
...
+
+
+``bootstrap``
+-------------
+
+Sets up a new cluster
+
+Usage:
```
pmatulis, line 55: In the existing 'disk' command page I've used "Syntax" instead of "Usage" for the sub-commands and for the main command (here 'cluster') I've used "Usage". We can apply "Usage" everywhere I think.

docs/reference/commands/cluster.rst
```diff
@@ -0,0 +1,210 @@
...
+
+.. code-block:: none
+
+  microceph cluster bootstrap [flags]
```
pmatulis, line 59: We'll need to include what the s/flags/options are. The existing 'remove' sub-command on the 'disk' command page includes:

```
Options:

.. code-block:: none

   --bypass-safety-checks               Bypass safety checks
   ...
   ...
```

docs/reference/commands/cluster.rst
```diff
@@ -0,0 +1,210 @@
...
+Manage Ceph Cluster configs
+
+Usage:
+
+.. code-block:: none
+
+  microceph cluster config [flags]
+  microceph cluster config [command]
+
+Available Commands:
```
pmatulis, line 73: s/Commands/commands

docs/reference/commands/cluster.rst
```diff
@@ -0,0 +1,210 @@
...
+
+Clear specified Ceph Cluster config
+
+Usage:
+
+.. code-block:: none
+
+  microceph cluster config reset <key> [flags]
+
+Flags:
```
pmatulis, line 118: 'Options:' everywhere

docs/reference/commands/cluster.rst
sabaini: Hm, this again is generated from the library. We can change it here bit this would then be inconsistent with the help command output

docs/reference/commands/cluster.rst
sabaini: Same as above, can we standardize on `[flags]`?

sabaini:
Hi @pmatulis I've uploaded some changes, two questions inline

docs/reference/commands/cluster.rst
sabaini: I would like to keep this consistent with the help command output -- and this uses the `[flags]` slug (from the library, hard to change in our code)

Would it be ok unify this to `[flags]`?

docs/reference/commands/cluster.rst
sabaini: In this case this would be global flags. I'm not sure we want to repeat them?

docs/reference/commands/cluster.rst
pmatulis: Alright, let's go with **flags** everywhere, unless there is a subtle difference between a flag and an option that I'm not privy to?

docs/reference/commands/cluster.rst
pmatulis: Yep.

docs/reference/commands/cluster.rst
pmatulis: Oh, if they are available as global flags then no, let's not repeat.

pmatulis (changes requested):
Some code blocks remain unaligned.

sabaini:
@pmatulis hope this works better

docs/reference/commands/cluster.rst
```diff
@@ -0,0 +1,209 @@
...
+Usage:
+
+.. code-block:: none
+
+   microceph cluster bootstrap [flags]
+
+``config``
+----------
+
+Manage Ceph Cluster configs
```
pmatulis, line 64: A previous request was to have the initial word in these sections be like "Manages" as opposed to "Manage". There are still several sections that do not do this.

docs/reference/commands/cluster.rst
```diff
@@ -0,0 +1,209 @@
...
+Usage:
+
+.. code-block:: none
+
+   microceph cluster list [flags]
+
+
+``migrate``
+-----------
+
```
pmatulis, line 170: Please remove extra blank line

pmatulis (changes requested):
Another systemic change is needed.

docs/reference/commands/cluster.rst
sabaini: Ack, hope I've got them

pmatulis (approved):
Thank you. Waiting on CI checks.

PR #248, opened 2023-10-19
Fix: tighten requirements for removing nodes

sabaini:
In testing we are seeing situations where a quorum could not be established. Require a min. of 3 mons in the non-forced case to prevent this.


lmlg (approved):
LGTM.

Issue #197, opened 2023-08-28
`microceph cluster config set public_network` doesn't update Ceph MON endpoints


nobuto-m:
I followed the document in:
https://canonical-microceph.readthedocs-hosted.com/en/latest/how-to/cluster-cfg/

After setting a public_network with `sudo microceph cluster config set public_network`, it updated the Ceph config database. However, it doesn't actually update the network to be used for a real traffic.

My networks:
```
enp1s0 192.168.123.0/24 <- main / bootstrap network
enp8s0 192.168.101.0/24 <- Ceph access
enp9s0 192.168.102.0/24 <- Ceph replication
```

Bootstrapping was done over the main network (192.168.123.0/24).
```
$ sudo microcloud cluster list
+------+----------------------+-------+------------------------------------------------------------------+--------+
| NAME |       ADDRESS        | ROLE  |                           FINGERPRINT                            | STATUS |
+------+----------------------+-------+------------------------------------------------------------------+--------+
| mc-1 | 192.168.123.171:9443 | voter | 3d183a5ae053df310584645d799c5b37161705b17ffe6ee2295382900b39024d | ONLINE |
+------+----------------------+-------+------------------------------------------------------------------+--------+
| mc-2 | 192.168.123.229:9443 | voter | 75d3fc81fd7c180c42a51024b82fec8ab13eccaf4cf0e4f358520f8140565fb4 | ONLINE |
+------+----------------------+-------+------------------------------------------------------------------+--------+
| mc-3 | 192.168.123.180:9443 | voter | d41e64ccbd011ca468118e439dd13544b78c8279459476b0be5874dca51b8b18 | ONLINE |
+------+----------------------+-------+------------------------------------------------------------------+--------+
```

Ceph MON endpoints are on 192.168.123.0/24 initially (as expected).
```
$ sudo microceph.ceph mon dump
epoch 4
fsid df66f326-14ad-440b-a60a-c51091d5f510
last_changed 2023-08-28T04:36:50.510744+0000
created 2023-08-28T04:36:30.416710+0000
min_mon_release 17 (quincy)
election_strategy: 1
0: [v2:192.168.123.171:3300/0,v1:192.168.123.171:6789/0] mon.mc-1
1: [v2:192.168.123.229:3300/0,v1:192.168.123.229:6789/0] mon.mc-2
2: [v2:192.168.123.180:3300/0,v1:192.168.123.180:6789/0] mon.mc-3
dumped monmap epoch 4
```

[...]

PR #252, opened 2023-10-23
Adds --mon-ip parameter to specify IP for monitor bootstrap

UtkarshBhatthere:


microceph/common/network.go
```diff
@@ -0,0 +1,40 @@
+package common
+
+import (
+	"fmt"
+	"net"
+)
+
+// ValidateNetworkAddress validates the provided IP on host, returns the network
+// CIDR containing the provided address; empty string on failures.
+func ValidateNetworkAddress(address string) (string, error) {
```
lmlg, line 10: I feel like the name doesn't match the implementation. To me, `validate` alludes to essentially running some checks only; this function rather returns a network address instead (which doesn't seem to be used anywhere either?)

lmlg (approved):
LGTM

microceph/common/network.go
UtkarshBhatthere: done

nobuto-m:
Are you folks going to implement `--cluster-ip` or something together with `--mon-ip`/`--public-ip`? So that we don't have to go through the post-installation steps like `microceph cluster config set cluster_network 192.168.102.0/24.`

nobuto-m:
Doesn't it make more sense if we say the new option as `--public-network` / `--cluster-network` to accept a subnet instead of a bare IP address to be aligned with the Ceph config options?

ref: https://github.com/canonical/microceph/issues/197

UtkarshBhatthere:
> Doesn't it make more sense if we say the new option as `--public-network` / `--cluster-network` to accept a subnet instead of a bare IP address to be aligned with the Ceph config options?

The `--mon-ip` bootstrap method is consistent with popular tools like `cephadm` and is widely documented in [upstream ceph documentation](https://docs.ceph.com/en/latest/cephadm/install/#bootstrap-a-new-cluster) and similar resources. MicroCeph follows `mon-host` addressing for monitors, and, while I am sure we'll implement `--public-network` and `--cluster-network` parameters for bootstrap that would be done as a separate `feature enhancement` PR.
@nobuto-m

nobuto-m:
> > Doesn't it make more sense if we say the new option as `--public-network` / `--cluster-network` to accept a subnet instead of a bare IP address to be aligned with the Ceph config options?
> 
> The `--mon-ip` bootstrap method is consistent with popular tools like `cephadm` and is widely documented in [upstream ceph documentation](https://docs.ceph.com/en/latest/cephadm/install/#bootstrap-a-new-cluster) and similar resources. MicroCeph follows `mon-host` addressing for monitors, and, while I am sure we'll implement `--public-network` and `--cluster-network` parameters for bootstrap that would be done as a separate `feature enhancement` PR. @nobuto-m

Makes sense. Thanks for the clarification.

microceph/ceph/bootstrap.go
```diff
@@ -30,12 +30,19 @@ func Bootstrap(s common.StateInterface) error {
...
+		return nil
+	}
+
+	err = conf.WriteConfig(
 		map[string]any{
-			"fsid":     fsid,
-			"runDir":   pathConsts.RunPath,
+			"fsid":   fsid,
+			"runDir": pathConsts.RunPath,
+			// First monitor bootstrap IP as passed to microcluster.
```
sabaini, line 42: I wonder: why are we setting up the initial mon address with the default ipaddr, even if the user might have passed in a separate `--mon-ip`? It would be good to have some rationale for this approach here

microceph/ceph/bootstrap.go
```diff
@@ -194,36 +201,48 @@ func initMgr(s common.StateInterface, dataPath string) error {
...
 		}
 
 		_, err = database.CreateConfigItem(ctx, tx, database.ConfigItem{Key: "keyring.client.admin", Value: adminKey})
 		if err != nil {
-			return fmt.Errorf("Failed to record keyring: %w", err)
+			return fmt.Errorf("failed to record keyring: %w", err)
+		}
+
+		key := fmt.Sprintf("mon.host.%s", s.ClusterState().Name())
+		_, err = database.CreateConfigItem(ctx, tx, database.ConfigItem{Key: key, Value: s.ClusterState().Address().Hostname()})
```
sabaini, line 238: Should we check if the user passed in a `--mon-ip` and record this instead of the default microcluster ipaddr?

microceph/common/network.go
```diff
@@ -0,0 +1,73 @@
...
+		return "", err
+	}
+
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		return "", err
+	}
+
+	for _, iface := range ifaces {
+		addrs, _ := iface.Addrs()
```
sabaini, line 22: If there is an issue with the iface we might get back an error here. IMHO it's fine to continue in this case but we might want to log the error

microceph/common/network.go
```diff
@@ -0,0 +1,73 @@
...
+	monIp := net.ParseIP(address)
+	nw := []string{}
+
+	ifaces, err := net.Interfaces()
+	if err != nil {
+		return "", err
+	}
+
+	for _, iface := range ifaces {
+		addrs, _ := iface.Addrs()
```
sabaini, line 54: Similar to above, maybe log the error

microceph/common/network.go
```diff
@@ -0,0 +1,73 @@
...
+			ipNet, ok := addr.(*net.IPNet)
+			if !ok {
+				continue
+			}
+
+			if !ipNet.IP.IsGlobalUnicast() {
+				continue
+			}
+
+			nw = append(nw, addr.String())
```
sabaini, line 65: The `nw` slice seems unused, is that right?

microceph/common/network.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+	}
+	return "", fmt.Errorf("no IP belongs to provided subnet %s", subnet)
+}
+
+// FindNetworkAddress locates the provided IP address on host's network interfaces.
+// It returns the containing subnet address on success or an empty string on failure.
+func FindNetworkAddress(address string) (string, error) {
+	// Parse provided address.
+	monIp := net.ParseIP(address)
```
sabaini, line 45: If `address` is invalid we'd be getting back a nil monIp, suggest to check for that to avoid a nil panic

sabaini (changes requested):
Hey Utkarsh, 

Many thanks for that -- I've added some comments.

Would it be possible to add some testing for the multinode test case as well?

microceph/common/network.go
UtkarshBhatthere: ack

microceph/ceph/join.go
```diff
@@ -103,12 +103,46 @@ func Join(s common.StateInterface) error {
...
+
+				monHost, err := common.FindIpOnSubnet(configItem.Value)
+				if err != nil {
+					return fmt.Errorf("failed to record mon host: %w", err)
+				}
+
+				key := fmt.Sprintf("mon.host.%s", s.ClusterState().Name())
+				_, err = database.CreateConfigItem(ctx, tx, database.ConfigItem{Key: key, Value: monHost})
+				if err != nil {
+					return fmt.Errorf("failed to record mon host: %w", err)
```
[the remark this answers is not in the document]
UtkarshBhatthere: This one is correct 🙂

microceph/cmd/microceph/cluster_bootstrap.go
```diff
@@ -33,18 +37,27 @@ func (c *cmdClusterBootstrap) Run(cmd *cobra.Command, args []string) error {
...
 	}
 
 	// Get system address.
-	address := util.NetworkInterfaceAddress()
-	address = util.CanonicalNetworkAddress(address, 7443)
+	var address string
+	if len(c.flagMonIp) > 0 {
+		_, err = common.FindNetworkAddress(c.flagMonIp)
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
UtkarshBhatthere: ack

microceph/common/network.go
UtkarshBhatthere: The nw list is used to provided details of available networks that could be provided instead (by the user) in case of incorrect parameters being passed to bootstrap.

microceph/common/network.go
UtkarshBhatthere: good catch, thanks 😄

microceph/api/services.go
```diff
@@ -22,6 +22,7 @@ var servicesCmd = rest.Endpoint{
 	Path: "services",
 
 	Get: rest.EndpointAction{Handler: cmdServicesGet, ProxyTarget: true},
+	Put: rest.EndpointAction{Handler: cmdServicesPut, ProxyTarget: false},
```
sabaini, line 25: I'd suggest to use a POST request here, as with PUT there is an expectation that requests are idempotent
Also I'd lean towards introducing a separate endpoint e.g. /services/bootstrap for this to make explicit which operation is intended

microceph/cmd/microceph/cluster_bootstrap.go
```diff
@@ -33,18 +39,56 @@ func (c *cmdClusterBootstrap) Run(cmd *cobra.Command, args []string) error {
...
+		return fmt.Errorf("failed to parse bootstrap data: %w", err)
+	}
+
+	// Bootstrap microceph daemon.
+	err = m.NewCluster(hostname, address, time.Second*30)
+	if err != nil {
+		return err
+	}
+
+	// TODO: Check if polling is required here.
```
sabaini, line 73: Do we or does the client retry do this for us?

sabaini (changes requested):
Hi Utkarsh, many thanks -- think the approach with separating out bootstrapping will serve well for future feature additions

Some comments inline.

Two things will be needed in addition:

- Functional testing
- Updateing the `microceph init` command to work with the new approach

microceph/ceph/bootstrap.go
```diff
@@ -10,32 +10,40 @@ import (
...
 		}
 	}
 
 	// Generate a new FSID.
 	fsid := uuid.NewRandom().String()
 	conf := newCephConfig(pathConsts.ConfPath)
-	err := conf.WriteConfig(
+	pubNet, err := common.Network.FindNetworkAddress(data.MonIp)
+	if err != nil {
+		return nil
```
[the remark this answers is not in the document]
UtkarshBhatthere: that is, definitely, my bad.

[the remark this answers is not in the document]
UtkarshBhatthere: fixed.

microceph/api/services.go
UtkarshBhatthere: well, my only reservation about that would be that it diverges from the `services/{name}` that we follow for the services endpoint.

microceph/cmd/microceph/cluster_bootstrap.go
UtkarshBhatthere: Added retry.

microceph/cmd/microceph/cluster_bootstrap.go
```diff
@@ -33,18 +39,56 @@ func (c *cmdClusterBootstrap) Run(cmd *cobra.Command, args []string) error {
...
+
+	err = client.BootstrapCephCluster(context.Background(), cli, &data)
+	if err != nil {
+		return fmt.Errorf("bootstrap command failed: %w", err)
+	}
+
+	return nil
+}
+
+func getCephBootstrapData(monip string) (types.Bootstrap, error) {
```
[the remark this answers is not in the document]
UtkarshBhatthere: ack

microceph/api/services.go
UtkarshBhatthere: The endpoint for the time being is changed to "ops/bootstrap", subject to change if the discussion with microcluster changes how we pull this up.

PR #253, opened 2023-10-31
WIP: Fix disk add race

sabaini:
Get rid of the racy OSD id calculation, make the database the single source of truth for OSD ids.


microceph/ceph/osd.go
```diff
@@ -31,56 +29,6 @@ import (
 	"github.com/canonical/microceph/microceph/database"
 )
 
-func nextOSD(s *state.State) (int64, error) {
```
UtkarshBhatthere, line 34: Appreciating the fact that this fix reduces a lot of code too :D

microceph/database/diskpath.go
```diff
@@ -1,37 +1,35 @@
...
+//go:generate mapper method -i -d github.com/canonical/microcluster/cluster -e DiskPath Exists
+//go:generate mapper method -i -d github.com/canonical/microcluster/cluster -e DiskPath Create
+//go:generate mapper method -i -d github.com/canonical/microcluster/cluster -e DiskPath DeleteOne-by-Member-and-Path
+//go:generate mapper method -i -d github.com/canonical/microcluster/cluster -e DiskPath DeleteMany-by-Member
+//go:generate mapper method -i -d github.com/canonical/microcluster/cluster -e DiskPath Update
 
-// Disk is used to track the Ceph disks on a particular server.
-type Disk struct {
+// DiskPath is used to track the Ceph disks on a particular server.
+type DiskPath struct {
```
UtkarshBhatthere, line 25: One issue with using a very specific name such as DiskPath for the Disks table struct is that in future if we extend OSD functionality which requires additional columns in the table, the name `DiskPath` might not be relevant for the struct then. Which would mean another sql migration going forward ? Since the db change is basically dropping a (column in table/ field in struct) I suggest we keep the `Disk` name and not change it ?

UtkarshBhatthere (changes requested):
One minor nit with naming. I love the fact that this change reduces the amount of code for spinning up OSDs too! That's awesome :D

UtkarshBhatthere (approved):
LGTM

PR #263, opened 2023-11-24
Doc: loop file backed OSDs

sabaini:


docs/reference/commands/disk.rst
```diff
@@ -33,13 +33,27 @@ Global flags:
...
+The specification for loop files is of the form "loop,<size>,<nr>"
+
+- size is a number with M, G, or T suffixes for megabytes, gigabytes, or terabytes.
+- nr is the number of file-backed loop OSDs to create.
+
+For instance, a spec of `loop,8G,3` will create 3 file-backed loop OSDs of 8GB each.
+
+While loop files provide a convenient way to quickly set up small test and development clusters, for
+production deployments it is **strongly** recommended to use dedicated block devices for safety and
+performance reasons.
```
pmatulis, line 49: I feel that this paragraph should be put into the single-node tutorial.

docs/reference/commands/disk.rst
```diff
@@ -33,13 +33,27 @@ Global flags:
 ``add``
 -------
 
-Adds a disk to the cluster, alongside optional devices for write-ahead logging and database management.
+Adds a new Ceph disk (OSD) to the cluster, alongside optional devices for write-ahead logging and database management.
+
+The command takes a parameter SPEC which is either a path to a block device such as /dev/sdb or a specification for one or more loop files.
```
pmatulis, line 38: Is "loop file" a thing? Not "loopback file" or "loopback device"?

So it's written "a block device" but "one or more loopback devices". Should they not both be either singular or multiple?

docs/reference/commands/disk.rst
```diff
@@ -33,13 +33,27 @@ Global flags:
...
+
+The specification for loop files is of the form "loop,<size>,<nr>"
```
pmatulis, line 40: please monospace `loop,<size>,<nr>`

docs/reference/commands/disk.rst
```diff
@@ -33,13 +33,27 @@ Global flags:
...
+
+- size is a number with M, G, or T suffixes for megabytes, gigabytes, or terabytes.
```
pmatulis, line 42: `M`, `G`, or `T`

pmatulis, line 42: `size` is an integer

docs/reference/commands/disk.rst
```diff
@@ -33,13 +33,27 @@ Global flags:
...
+- nr is the number of file-backed loop OSDs to create.
```
pmatulis, line 43: `nr`

i think we should settle upon one way of expressing these devices. If we are qualifying the OSD then perhaps: "loopback-based OSD" or "file-backed OSD" but "file-backed loop OSD" sounds off to me

docs/reference/commands/disk.rst
pmatulis: what is the safety concern with loopback devices? it's not clear to me, but if there is a security concern then we might want to spell that out

docs/reference/commands/disk.rst
```diff
@@ -48,7 +62,7 @@ Flags:
    --db-device string    The device used for the DB
    --db-encrypt          Encrypt the DB device prior to use
    --db-wipe             Wipe the DB device prior to use
-   --encrypt             Encrypt the disk prior to use
+   --encrypt             Encrypt the disk prior to use (only block devices)
```
pmatulis, line 65: i would s/only block devices/loopback devices not supported/

a loopback device is technically a block device

docs/reference/commands/disk.rst
```diff
@@ -64,6 +78,10 @@ Flags:
    as long as they are faster than the data device. WAL should take priority
    over DB if there isn't enough storage for both.
 
+   WAL and DB devices can only be used with data devices that reside on a
+   block device, not with loop files. Loop files do not support encryption.
```
pmatulis, line 82: on a physical block device?

still not sure about "loop files". again i think we should settle on describing the device and not a type of file

docs/reference/commands/disk.rst
```diff
@@ -33,13 +33,27 @@ Global flags:
...
+production deployments it is **strongly** recommended to use dedicated block devices for safety and
+performance reasons.
+
 
 Usage:
 
 .. code-block:: none
 
-   microceph disk add <PATH> [flags]
+   microceph disk add <SPEC> [flags]
```
pmatulis, line 56: we started capitalising arguments but i think we should be consistent by staying with lowercase

docs/reference/commands/disk.rst
pmatulis: i see you have also added this to the single-node file. same concerns but i would simply remove this content from this current **reference** file

docs/tutorial/single-node.rst
```diff
@@ -89,20 +59,27 @@ yet any storage available.
 Add storage
 -----------
 
-.. warning::
-
-   This step will remove the data found on the target storage disks. Make sure
-   you don't lose data unintentionally.
+Three OSDs will be required to form a minimal Ceph cluster. In a
+production system, typically we would assign one physical block device
+for an OSD. However for this tutorial, we will make use of file backed
```
pmatulis, line 64: assign **a** ... **to** an OSD

docs/tutorial/single-node.rst
```diff
@@ -89,20 +59,27 @@ yet any storage available.
...
+OSDs for simplicity. In this configuration, we won't be using real
+block devices for storage but instead simulate using regular files.
```
pmatulis, line 66: this second sentence is redundant. either reword into a single sentence or just remove the second sentence

docs/tutorial/single-node.rst
```diff
@@ -89,20 +59,27 @@ yet any storage available.
...
 
-Add the three disks to the cluster by using the :command:`disk add` command:
+Add the three file-backed OSDs to the cluster by using the
+:command:`disk add` command. In the example, three 4Gb files are being
```
pmatulis, line 69: s/4Gb/4GB

OpenStack docs would use `4GiB`

docs/tutorial/single-node.rst
```diff
@@ -89,20 +59,27 @@ yet any storage available.
...
+created -- adjust the size to your needs, with a recommended minimum
+of 2G per OSD:
```
pmatulis, line 71: s/2G/2GB

pmatulis (changes requested):

docs/reference/commands/disk.rst
sabaini: The safety concern is about OSD failure domains. If you use several file-backed OSDs they will necessarily live on a single physical disk. So even though you have e.g. 3xOSD your data will be gone if that single disk fails

sabaini: I'll remove this para here

docs/reference/commands/disk.rst
sabaini: We're not actually using a loop device (i.e. the things `losetup` configures) since Ceph has the ability to directly writing to a backing file. "Loop file" is our shorthand for that (but happy to take suggestions on wording that better). A "loopback device" refers to the `lo` device in networking, so that doesn't really fit here. 

In  the current implementation we indeed have the ability to either add a single (real) block device or several OSDs backed by files

pmatulis: Okay,let's go with this.

pmatulis (approved):

PR #267, opened 2023-11-27
Refactor wording around loop file overhead

pmatulis:
Move some words around and streamline. Make users more aware about loop file usage and their overhead.

### DIFF
--- a/docs/.custom_wordlist.txt
+++ b/docs/.custom_wordlist.txt
@@ -58,4 +58,4 @@ conf
 qemu
 writethrough
 writeback
-
+IOPS
--- a/docs/tutorial/single-node.rst
+++ b/docs/tutorial/single-node.rst
@@ -5,6 +5,16 @@ Single-node install
 This tutorial will show how to install MicroCeph on a single machine, thereby
 creating a single-node "cluster".
 
+The above will be achieved through the use of loop files placed on the root
+disk, which is a convenient way for setting up small test and development
+clusters.
+
+.. warning::
+
+   Due to the extra overhead involved in loop file usage, for production
+   environments, it is **strongly** recommended to use dedicated block devices.
+   Basing a Ceph cluster on a single disk also necessarily creates a common
+   failure domain for all OSDs.
 
 Install the software
 --------------------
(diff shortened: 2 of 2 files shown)


sabaini (changes requested):
I'm a bit wary of the wording here, it makes it sound like OSDs on loop files need more resources than OSDs on phys. devices. There's really two things at play here:

1) loop file access exacts some latency penalty which will result in less IOPS, but that is independent of actual number of loop files

2) OSDs are relatively heavyweight processes, this will scale linearly with the number of OSDs, irrespective of the backing -- it's the same for loop files and phys. devices. It's just easier to bring up large number of loop files than to install large numbers of disks :-)

The other thing that comes to mind here is that there's really no good reason to bring up more than 3 loop file backed OSDs per node. Failure domain will be the single disk the loop files are located on anyways. This maybe could be noted here as well

pmatulis:
Well the original states:

> it is **strongly** recommended to utilise dedicated block devices in production environments for enhanced safety and performance. Utilising loop files involves extra
overhead compared to physical block devices, leading to a performance penalty.

Are you saying that the above is also misworded or is my change over emphasising something?

sabaini:
> Well the original states:
> 
> > it is **strongly** recommended to utilise dedicated block devices in production environments for enhanced safety and performance. Utilising loop files involves extra
> > overhead compared to physical block devices, leading to a performance penalty.
> 
> Are you saying that the above is also misworded or is my change over emphasising something?

My wording might have been unclear. The two effects at play here have different impacts

1) additional latency for loop file backed OSDs: this will lower IOPS for clients but not impact the Ceph cluster nodes itself

2) the resource needs for OSDs (whether for loop files or phys. block devices) require resources from Ceph cluster nodes, so you need to be careful not to overload them with too many OSDs

My original recommendation against using loop files in prod is because of 1) and the safety implications of running OSDs off a single disk. 

Item 2) impacts any Ceph installation, regardless of type of backing medium

I think we might want to add wording to the effect that for normal use cases, running more than 3 OSDs off of loop files is not advised. It doesn't buy users anything really, and only might be interesting if you actually want to test out running many OSDs on a node

pmatulis:
Okay let me simplify this then.

sabaini (approved):
Sweet, thank you!

PR #272, opened 2023-12-05
Fix emoji heading in README again

pmatulis:


README.md
```diff
@@ -15,7 +15,7 @@
 
 * [💡 Philosophy](#-philosophy)
 * [🎯 Features](#-features)
-* [⚡️Quickstart](#-quickstart)
+* [⚡️Quickstart](#%EF%B8%8Fquickstart)
```
UtkarshBhatthere, line 18: This change confuses me a bit.

UtkarshBhatthere (changes requested):
I think a small description on why these characters are needed maybe helpful ?

pmatulis:
This is the link that shows up when the Markdown is rendered. It was the only way that I was able to get this particular emoji to show up properly. I agree that it looks odd. Feel free to suggest a different solution.

PR #278, opened 2023-12-13
Docs: howto on upgrading to Reef

sabaini:


lmlg (commented):
overall lgtm. I was wondering if we should warn users about filestore being deprecated in Reef?

docs/how-to/reef-upgrade.rst
```diff
@@ -0,0 +1,72 @@
...
+Procedure
+---------
+
+
+Optional but Recommended: Preparation Steps
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Carry out these precautionary steps before initiating the upgrade:
+
+1. **Back up your data**: to avoid any potential data loss, it is recommended to back up all your important data before starting the upgrade.
```
UtkarshBhatthere, line 21: I feel we should clarify a bit here on what we are asking them to backup. Backup Cephx keys ? or the actual data stored in the ceph cluster (makes me wonder how).... something like that.

UtkarshBhatthere (changes requested):
one small clarification suggestion, I also agree with lmlg's opinion on file-store support.

docs/how-to/reef-upgrade.rst
sabaini: This is specifically about the users' data -- this will be application dependent on what to backup. I don't think it's necessary to backup credentials because the snap machinery takes snapshots of those so they should be safe to recover

sabaini:
I didn't mention FileStore support because in MicroCeph we never provided any support for creating FileStore OSDs -- we only ever created BlueStore OSDs. With some heavy contortions users might be able to create FS OSDs by hand but I do wonder if this really a usecase?

lmlg:
> I didn't mention FileStore support because in MicroCeph we never provided any support for creating FileStore OSDs -- we only ever created BlueStore OSDs. With some heavy contortions users might be able to create FS OSDs by hand but I do wonder if this really a usecase?

Yes, that's a good point. With that out, I think this is good to merge :)

docs/how-to/reef-upgrade.rst
sabaini: @UtkarshBhatthere if you have ideas on clarifying wording happy to include those :-)

UtkarshBhatthere: I have nothing specific, just reading the doc made me question 'what should I backup ?'. A user may wonder what is at risk of loss on upgrades.

sabaini: I've hopefully clarified a bit

PR #283, opened 2024-01-10
Cleanup duplicate upgrade workflow

sabaini:
candidate-upgrade.yaml is a duplicate of q2r-candidate-upgrade.yaml


UtkarshBhatthere:
LGTM, on a separate note, the Disk Removal test is very flaky and needs to be fixed.

PR #308, opened 2024-02-20
Fix flaky removal test

lmlg:
This PR aims to fix the removal test to take into consideration that a node that is being removed may be temporarily reported as 'out of quorum' instead of completely out.


sabaini (approved):
Sweet, thanks!

PR #309, opened 2024-02-21
Fix: allow mount observe

sabaini:
Also test listing disks


UtkarshBhatthere (approved):
LGTM, thanks for a quick patch.

PR #311, opened 2024-02-22
Fix: move logrus to direct dep

sabaini:
Make it a direct dependency (otherwise the "go get" during build dirties the worktree)


lmlg (approved):
lgtm

PR #313, opened 2024-02-23
docs: Add a pointer to Ubuntu Pro

pmatulis:
Users should be informed about the possibility of adding Pro to their Ceph nodes.


lmlg (approved):
lgtm, thanks!

Issue #318, opened 2024-02-27
Missing public_network config key after upgrade


sabaini:
# Issue report

## What version of MicroCeph are you using ?

Multiple versions

## What are the steps to reproduce this issue ?

1. Install and bootstrap 
2. Upgrade 2x

## What happens (observed behaviour) ?

After the second upgrade, the MON fails to come up

## What were you expecting to happen ?

All services to be started 

## Relevant logs, error output, etc.

Detailed steps

```
# Install and bootstrap

root@ic-2:~# snap install microceph --channel quincy/stable
microceph (quincy/stable) 0+git.4a608fc from Canonical✓ installed
root@ic-2:~# microceph cluster bootstrap

# Observe: we have fsid and keyring.client.admin as conf keys

root@ic-2:~# sudo microceph cluster sql "select * from config"
+----+----------------------+------------------------------------------+
| id |         key          |                  value                   |
+----+----------------------+------------------------------------------+
| 1  | fsid                 | de5fc10b-f2d2-4653-b69a-228173419b30     |
| 2  | keyring.client.admin | AQAQDN5l6MM9HhAA8+lJViDNRBeIK2vZMdnvJw== |
+----+----------------------+------------------------------------------+

# Upgrade

root@ic-2:~# snap refresh microceph --channel reef/stable
2024-02-27T16:21:58Z INFO Waiting for "snap.microceph.mds.service" to stop.
microceph (reef/stable) 18.2.0+snap3f8909a69a from Canonical✓ refreshed

# Observe: conf keys are unchanged

root@ic-2:~# sudo microceph cluster sql "select * from config"
+----+----------------------+------------------------------------------+
| id |         key          |                  value                   |
+----+----------------------+------------------------------------------+
| 1  | fsid                 | de5fc10b-f2d2-4653-b69a-228173419b30     |
| 2  | keyring.client.admin | AQAQDN5l6MM9HhAA8+lJViDNRBeIK2vZMdnvJw== |
+----+----------------------+------------------------------------------+

# Observe: current versions of the snap on the system

root@ic-2:~# ls -la /var/snap/microceph
total 20
drwxr-xr-x 5 root root 4096 Feb 27 16:22 .
drwxr-xr-x 7 root root 4096 Feb 27 16:21 ..
drwxr-xr-x 4 root root 4096 Feb 27 16:21 793
drwxr-xr-x 4 root root 4096 Feb 27 16:21 862
[...]

PR #322, opened 2024-02-28
Fix: ensure backward compat config

sabaini:
Fixes: #318

Ensure we populate the config key value db with required values if not present. Newer microceph version set these values on bootstrap, older releases might be missing them

Also add some testing for ceph.conf

### DIFF
--- a/microceph/ceph/config.go
+++ b/microceph/ceph/config.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"github.com/canonical/microceph/microceph/interfaces"
+	"net"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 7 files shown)


UtkarshBhatthere (approved):
Thanks for the change Peter! LGTM, Big + on the logging entries as well.

microceph/ceph/config.go
```diff
@@ -162,43 +163,104 @@ func ListConfigs() (types.Configs, error) {
...
+// which will ensure that the public_network is set in the database
+func backwardCompatPubnet(s interfaces.StateInterface) error {
+	config, err := getConfigDb(s)
+	if err != nil {
+		return fmt.Errorf("failed to get config from db: %w", err)
+	}
 
-	// Get the configuration and servers.
+	// do we have a public_network configured?
+	pubNet := config["public_network"]
```
UtkarshBhatthere, line 176: NIT: Should we not check if the key exists ?
```
val, ok := myMap["foo"]
if !ok {
    // handle absence of public_network.
}
```

sabaini: The reasoning is if we don't have a pub net, config["public_network"] will evaluate to the empty string, which will fail the ParseCIDR() coming below

UtkarshBhatthere: ah okay, I was assuming this to be the case. Can we just add a short comment which describes this in place of the `do we have a public_network configured?` comment ? I think that could be helpful.

PR #328, opened 2024-03-13
Fixup build check

sabaini:
Don't check untracked files so as not to fail for parts/ and stage/ dirs


UtkarshBhatthere (approved):
LGTM, ship it :)

PR #330, opened 2024-03-22
Added `--microceph-ip` flag for node join

UtkarshBhatthere:


sabaini (commented):
Thanks for picking this up! LGTM in general -- I wonder if we could have some testing around this?

UtkarshBhatthere:
yeah sure, I will add a testcase similar to public/cluster network verification test.

lmlg (approved):
lgtm

PR #340, opened 2024-04-16
Use microcluster API to remove cluster members

### LINKED ISSUE #255: Expose complete node removal process via the microcluster API
Currently, node removal is coupled with the CLI here with multiple API calls: https://github.com/canonical/microceph/blob/aae9f94dd349132283b08902b23aee27795e60fe/microceph/cmd/microceph/cluster_remove.go#L49

This means anyone using the MicroCeph API (MicroCloud) must re-implement the logic here, and maintain parity going forward, which is not ideal.

MicroCluster has hooks that can be used to automatically execute these necessary bits before/after a node is removed: https://github.com/canonical/microcluster/blob/e3b7b60a4d71577f5d0f5460f71dbb33dfcb5d9c/config/hooks.go#L25-L28

If the logic in `removeNode` is included in the remove hooks, then the only thing the MicroCeph CLI will need to do is call `DeleteClusterMember` which is currently here: https://github.com/canonical/microceph/blob/aae9f94dd349132283b08902b23aee27795e60fe/microceph/cmd/microceph/cluster_remove.go#L75

Then, anyone using the MicroCeph API would also be able to just remove the node and maintain all the necessary preparations for node removal.

masnax:
Closes #255 

As discussed in Riga, this refactors the node removal process to rely on the microcluster API instead of directly implementing removal in the MicroCeph CLI.

The main changes are that the `removeNode` function and all its helpers have been moved to the `ceph` package. The `removeNode` function itself has been wrapped in a `PreRemove` to match the corresponding hook in microcluster. Now in the CLI, we can simply call `DeleteClusterMember` and the `PreRemove` hook will automatically execute on that node as it is passed to microcluster when we call `app.Start`.

This will allow MicroCloud to utilize MicroCeph's API to tear down the cluster in case of an error during initialization. Additionally, it cleans up MicroCeph so there isn't so much internal implementation in the CLI package.

### DIFF
--- a/microceph/ceph/pre_remove_test.go
+++ b/microceph/ceph/pre_remove_test.go
@@ -1,12 +1,13 @@
-package main
+package ceph
 
 import (
+	"testing"
+
 	"github.com/canonical/microceph/microceph/api/types"
 	"github.com/canonical/microceph/microceph/client"
 	"github.com/canonical/microceph/microceph/mocks"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/mock"
-	"testing"
 
 	"github.com/stretchr/testify/suite"
 )
@@ -46,7 +47,6 @@ func (s *clusterRemoveSuite) TestRemoveNode() {
 		nil,
 	)
 	m.On("DeleteService", mock.Anything, "foonode", "mon").Return(nil).Once()
-	m.On("DeleteClusterMember", mock.Anything, "foonode", false).Return(nil).Once()
 
 	err := removeNode(nil, "foonode", false)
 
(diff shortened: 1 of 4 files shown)


masnax:
Setting as draft until I can sort out the last test.

masnax:
@UtkarshBhatthere @sabaini This one is now ready for review, thank you :)

masnax:
This should now be rebased and ready for review, thanks :)

UtkarshBhatthere (approved):
LGTM

masnax:
@sabaini Thanks for merging the other one, I've now rebased this one.

sabaini (approved):
lgtm, thanks!

PR #345, opened 2024-04-29
Build squid Microceph

UtkarshBhatthere:
# Description
Build MicroCeph using Ceph Squid from the Jammy-Caracal pocket

## Type of change

- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How Has This Been Tested?
The new build is expected to successfully comply with all existing CI tests.

## Contributor's Checklist

Please check that you have:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -215,6 +215,11 @@ jobs:
         output=$(sudo microceph log get-level)
         if [[ "$output" != "3" ]] ; then echo "incorrect log level: $output"; exit 1; fi 
 
+    - name: Print logs for failure
+      if: failure()
+      run: |
+        sudo snap logs microceph -n 1000
+
   multi-node-tests:
     name: Multi node testing
     runs-on: ubuntu-22.04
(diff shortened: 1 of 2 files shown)


.github/workflows/tests.yml
```diff
@@ -556,3 +581,8 @@ jobs:
 
     - name: Exercise RGW again
       run: ~/actionutils.sh headexec testrgw
+
+    - name: Print logs for failure
+      if: failure()
+      run: |
+        sudo snap logs microceph -n 1000
```
sabaini, line 588: Hm, I wish we could reduce the duplication of those print log stanzas here

snap/snapcraft.yaml
```diff
@@ -12,7 +12,7 @@ confinement: strict
 
 package-repositories:
   - type: apt
-    cloud: bobcat
+    cloud: caracal
```
sabaini, line 15: Not sure we want to build for Squid by default before having a stable upstream release.
In any case we should cut a reef branch before switching master

sabaini (changes requested):
Great to see the squid build works! -- think we might want to hold off before making S the default though

.github/workflows/tests.yml
UtkarshBhatthere: I am not sure if we can abstract this step out given, The actual step is a one-liner (and rest is step description in each run.).

snap/snapcraft.yaml
UtkarshBhatthere: I think these are independent. As soon as we switch the reef track to a stable branch we can and should also build/publish the squid snap (without a stable risk level maybe).

sabaini:
@UtkarshBhatthere think we did this now elsewhere, guess this PR could be closed?

PR #346, opened 2024-05-09
Add setuptools dependency

wolsen:
Ceph has a dependency on distutils, which was removed from the packaging since it was an undeclared package dependency. This change adds setuptools package to work around this dependency.

Since distutils has been removed from python 12, this also reverts to core22.


UtkarshBhatthere (approved):
Thanks @wolsen

PR #373, opened 2024-06-25
Documentation update for MicroCeph

UtkarshBhatthere:
# Description
Documentation updates for consuming and Backup of RBD and Cephfs.

Fixes #NA

## Type of change

- [ ] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?
Doc only change.

## Contributor's Checklist

Please check that you have:


lmlg (approved):
lgtm, good stuff

docs/tutorial/mount-block-device.rst
```diff
@@ -0,0 +1,172 @@
...
+   machine is the MicroCeph node itself.
+
+MicroCeph Operations:
+---------------------
+
+Check Ceph cluster's status:
+
+.. code-block:: none
+
+    $ sudo ceph -s
```
medvied, line 25: [* defect *] `ceph` command (and the `rbd` command below) no longer works after doing `sudo apt install ceph-common` below because before installing `ceph-common` package `ceph` was referring to the binary from MicroCeph
```
$ which ceph
/snap/bin/ceph
```
and after installing the package it's referring the the binary from the package:
```
$ which ceph
/usr/bin/ceph
```

docs/tutorial/mount-block-device.rst
```diff
@@ -0,0 +1,172 @@
...
+    1 .mgr
+    2 block_pool
+
+    $ rbd pool init block_pool
+
+Create RBD image:
+
+.. code-block:: none
+
+    $ sudo rbd create bd_foo --size 8192 --image-feature layering -p block_pool
```
medvied, line 58: [* style *] Here and elsewhere in this howto: it makes sense to use consistent naming for options and it's generally better to use long options in the documentation. `--size` is used here instead of `-s` which helps to understand what the the option is about without opening the man page, but `-p` is more cryptic, even if it's followed by the pool name. And a long version of the `-p` option exists.

PR #378, opened 2024-07-04
Update microcluster dependency

masnax:
This introduces a bit more robust schema upgrade processing, as well as better error handling during first start.

In particular, during an upgrade, the cluster will report its current status instead of blocking. The table below shows `microceph cluster list` output when 1 node is refreshed and awaiting the upgrade to appear on all nodes before proceeding to apply the update:

```
+--------+----------------+-------+-----------------------------+---------------+
|  NAME  |    ADDRESS     |  ROLE |         FINGERPRINT         |    STATUS     |
+--------+----------------+-------+-----------------------------+---------------+
| micro0 | 127.0.0.1:9000 | voter | c3d485471729f2f04f6401530f7 |   UPGRADING   |
+--------+----------------+-------+-----------------------------+---------------+
| micro1 | 127.0.0.1:9001 | voter | 515c7eace0aba1f7a42761824ab | NEEDS UPGRADE |
+--------+----------------+-------+-----------------------------+---------------+
| micro2 | 127.0.0.1:9002 | voter | 4f461765de6659b0e683a5b11ed | NEEDS UPGRADE |
+--------+----------------+-------+-----------------------------+---------------+
| micro3 | 127.0.0.1:9003 | voter | d94955e2c26cb2e987489933bdz | NEEDS UPGRADE |
+--------+----------------+-------+-----------------------------+---------------+
```

PR #387, opened 2024-07-12
Fix run dir issues

### LINKED ISSUE #386: Deadlock in upgrade process
# Issue report

When upgrading a multinode cluster we typically want to iterate over all nodes sequentially so as to keep the cluster available.

With the recent changes in microcluster, the microcephd now waits for all daemons to be on the same version before starting up. 

However, ceph daemons (mon, osd etc.) wait for the microcephd to update ceph.conf after an upgrade (fixing up `run dir` e.g.)

Thus we have a deadlock: upgrade can only continue with the next node after the current node is completely up, but this can't complete until all nodes are on the same version.

We'll need to decouple microcephd from the ceph daemons for the upgrade process, e.g. by having a separate post-upgrade hook that updates the ceph.conf run dir.

latest/edge, rev 1071

sabaini:
# Description

With the from PR #378 microcephd daemons now wait for all nodes to be on the same schema level in a multinode upgrade scenario. However we used to depend on the microcephd to update the `run dir` parameter in ceph.conf; for sequential updates (where we iteratively upgrade nodes while keeping the cluster operational) this creates a circular dependency.

This PR removes the need for updateing run dir during an upgrade.

Also contains func test improvements to verify the sequential upgrade scenario.

Fixes #386 

## Type of change

## How Has This Been Tested?

Functional tests updated

## Contributor's Checklist

Please check that you have:

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -549,7 +549,7 @@ jobs:
       run: ~/actionutils.sh headexec testrgw
 
     - name: Install local build
-      run: ~/actionutils.sh install_multinode
+      run: ~/actionutils.sh upgrade_multinode
 
     - name: Wait until 3 OSDs are up
       run:  ~/actionutils.sh headexec wait_for_osds 3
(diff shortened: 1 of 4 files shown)


UtkarshBhatthere (approved):
LGTM, Thanks @sabaini

PR #395, opened 2024-08-01
Use LTS microcluster

masnax:
Updates MicroCeph to use the stable release of microcluster at `v2.0.0`.

For the most part, the changes to the code in MicroCeph are just package imports and changing the definition of `state.State` to an interface. 

The actual behavioural changes are these:
* `state.Context` is no longer exposed, so `request.Context` from the API handler is passed around explicitly to each helper. This should actually result in more precise context handling for MicroCeph. 
  * One caveat is a few async API operations which spawn a goroutine. For these `context.Background()` is used because the request context may cancel before the goroutine is complete. 

* Adds a schema update to MicroCeph that changes their foreign key references to `core_cluster_members`. This table has been renamed from `internal_cluster_members` to make it clear that accessing it is expected behaviour. I also took the liberty of renaming the previous schema update function to use an ordered naming scheme so that it doesn't become ambiguous where an update fits among the others.

* Join tokens now have expiration dates. By default, MicroCeph will set a 3 hour expiration for join tokens, but this can be configured with the `--timeout` flag added to `cluster add`. 

The project version supplied to microcluster has been set to `UNKOWN` for now because microcluster requires a value here.


masnax:
Test failures appear to be because of the go module transition. Should be fixed when https://github.com/canonical/microcluster/pull/219 is merged.

UtkarshBhatthere:
Thanks for the change @masnax will take a look :D

microceph/database/schema.go
```diff
@@ -39,51 +64,109 @@ CREATE TABLE services (
...
+  `, tableName)
+	_, err = tx.Exec(stmt)
+
+	return err
+}
+
+// schemaUpdate4 updates all tables referencing `internal_cluster_members` to now reference `core_cluster_members` instead.
+func schemaUpdate4(ctx context.Context, tx *sql.Tx) error {
+	stmt := `
+CREATE TABLE disks2 (
```
UtkarshBhatthere, line 129: @sabaini It would be good to have a discussion for table names being introduced here.

microceph/cmd/microcephd/main.go
```diff
@@ -63,33 +62,42 @@ func (c *cmdDaemon) Command() *cobra.Command {
...
 	}
 
 	h.PreRemove = ceph.PreRemove(m)
 
-	m.AddServers(api.Servers)
-	return m.Start(context.Background(), database.SchemaExtensions, nil, h)
+	daemonArgs := microcluster.DaemonArgs{
+		Verbose:          c.global.flagLogVerbose,
+		Debug:            c.global.flagLogDebug,
+		Version:          "UNKNOWN", // FIXME: Add an explicit version to MicroCeph.
```
UtkarshBhatthere, line 93: @masnax can you please add some comments to describe what versioning is this.

microceph/cmd/microceph/status.go
```diff
@@ -27,7 +27,7 @@ func (c *cmdStatus) Command() *cobra.Command {
 }
 
 func (c *cmdStatus) Run(cmd *cobra.Command, args []string) error {
-	m, err := microcluster.App(microcluster.Args{StateDir: c.common.FlagStateDir, Verbose: c.common.FlagLogVerbose, Debug: c.common.FlagLogDebug})
```
UtkarshBhatthere, line 30: Hmmmmm this is a little confusing. Can this change be described a little (dropping the Verbose and Debug flags.)

UtkarshBhatthere (changes requested):
Did a round of reviews, have left questions and comments inline.

microceph/cmd/microceph/status.go
masnax: These args only applied to the daemon as it starts, but since we use the same struct to set up the microcluster client, these fields were populated even though they are never read. 

We have split the daemon and client args now, so these unused fields were removed.

microceph/cmd/microcephd/main.go
masnax: When using the MicroCeph API, it is useful to determine the current version of MicroCeph so that compatibility can be verified. We have an endpoint at `/core/1.0` that provides basic identifying information like the cluster name, address, and project version. 

For example, from MicroCloud, we should be able to tell from the API that MicroCeph is installed from a revision belonging to a supported snap track/risk. Like MicroCloud LTS should verify that MicroCeph is also installed as an LTS version. 

Microcluster validates that a version must be present, but I wasn't able to find any value in the project that matched this criteria so I put `UNKOWN` here. 


I'll update the commit with the above information as well.

microceph/database/schema.go
masnax: FWIW, the tables in MicroCeph have not been renamed, it is only the foreign key reference to the microcluster table that has changed. We ensure that foreign key references are not erased when we changed the table name internally, so no data will be lost. 

This had to be retroactively applied to all schema updates in case a very old version of MicroCeph tries to update to the latest version.

microceph/cmd/microceph/status.go
masnax: The flags are defined here now:
https://github.com/canonical/microceph/blob/fd6a044aa00a5201237c339755f8b4973bfeb4ae/microceph/cmd/microcephd/main.go#L90

This is the same place that they were read from before, but now the args type is different.

microceph/cmd/microcephd/main.go
masnax: I've added a comment in-line with the `Version` field now.

roosterfish (approved):
Thanks @masnax, looks good to me from microcluster perspective.

Aside from those changes I am wondering if we should parse the `--timeout` duration from string instead of requiring specifying it in seconds. 
This way if you intend to use a duration of a few hours you don't have to provide this big int which represents the seconds but instead provide only `6h` or `1d`.
Something like we do in the microcluster [example package](https://github.com/canonical/microcluster/blob/v3/example/cmd/microctl/tokens.go#L53).

masnax:
> Thanks @masnax, looks good to me from microcluster perspective.
> 
> Aside from those changes I am wondering if we should parse the `--timeout` duration from string instead of requiring specifying it in seconds. This way if you intend to use a duration of a few hours you don't have to provide this big int which represents the seconds but instead provide only `6h` or `1d`. Something like we do in the microcluster [example package](https://github.com/canonical/microcluster/blob/v3/example/cmd/microctl/tokens.go#L53).

Good idea, this has been applied now.

microceph/cmd/microcephd/main.go
```diff
@@ -63,33 +62,49 @@ func (c *cmdDaemon) Command() *cobra.Command {
...
-	m.AddServers(api.Servers)
-	return m.Start(context.Background(), database.SchemaExtensions, nil, h)
+	daemonArgs := microcluster.DaemonArgs{
+		// Microcluster requires an explicit version to be supplied to the daemon.
+		// This will be readable alongside other server information over the `GET /core/1.0` endpoint.
+		// MicroCeph should define a project version here that can be used by consumers of the MicroCeph API
+		// to ensure that MicroCeph is installed within a supported series of revisions.
+		// In particular, a semantic version would be very useful here to distinguish between breaking changes
+		// and non-breaking bugfixes to a supported series.
+		Version: "UNKNOWN", // FIXME: Add an explicit version to MicroCeph.
```
sabaini, line 97: Suggest to set this to the full MicroCeph version, and use a build flag as done for the CLI here:
https://github.com/canonical/microceph/blob/36f71d77006388164e312a2a5d5c4ba699a140a2/snap/snapcraft.yaml#L320

sabaini (approved):
This lgtm, thanks @masnax!
One comment inline wrt to versioning

microceph/cmd/microcephd/main.go
masnax: I've added this as the version string now. 

There was an unused `version` package which now stores this value both for the client and the daemon.

microceph/cmd/microceph/status.go
UtkarshBhatthere: Ack thanks for this :D

PR #398, opened 2024-08-08
MicroCeph Remote Replication (2/3): RBD Mirroring

UtkarshBhatthere:
# Description
This PR adds the RBD Mirroring FSM backing the replication API. 
[ ] Add description

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?
[ ] Add functional tests

## Contributor's Checklist

Please check that you have:


microceph/ceph/keyring.go
```diff
@@ -106,3 +106,39 @@ func parseKeyring(path string) (string, error) {
 
 	return cephSecret, nil
 }
+
+// CreateKey creates a client key and returns the said key hash without saving it as a file.
+var CreateKey = func(clientName string, caps ...[]string) (string, error) {
```
sabaini, line 111: See above re: mocking. The `cephRun()` func you're using here to perform system commands can be mocked already, see `var processExec Runner = RunnerImpl{}`

microceph/ceph/services_placement_generic.go
```diff
@@ -63,17 +66,17 @@ func genericHospitalityCheck(service string) error {
 	return nil
 }
 
-func genericServiceInit(s interfaces.StateInterface, name string) error {
+func genericServiceInit(s interfaces.StateInterface, name string, isClientService bool) error {
```
sabaini, line 69: I wonder if instead of adding a flag to the genericServiceInit the rbd-mirror service could subclass GenericServicePlacement, like the RGW service placement does? This would feel cleaner

sabaini (changes requested):
Hi Utkarsh, did a quick once-over review -- please find some comments inline

microceph/api/cluster.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		return response.InternalError(err)
+	}
+
+	// Check that the cluster name is conformant.
+	isOk, err := regexp.MatchString(constants.ClusterNameRegex, req.RemoteName)
+	if err != nil || !isOk {
+		err := fmt.Errorf("cluster names can only have [a-z] or [0-9] characters: %w", err)
+		logger.Error(err.Error())
+		return response.InternalError(err)
```
lmlg, line 38: This should be a `BadRequest`, since `InternalError` doesn't represent a user error, imo.

microceph/api/cluster.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	// fetch the cluster configurations from dqlite
+	configs, err := ceph.GetConfigDb(interfaces.CephState{State: s})
+	if err != nil {
+		err := fmt.Errorf("failed to get config db: %w", err)
+		logger.Error(err.Error())
+		return response.InternalError(err)
```
lmlg, line 46: Same as above.

microceph/api/cluster.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	// generate client keys
+	clientKey, err := ceph.CreateKey(
+		req.RemoteName,
+		[]string{"mon", "allow *"},
+		[]string{"osd", "allow *"},
+		[]string{"mds", "allow *"},
+		[]string{"mgr", "allow *"},
```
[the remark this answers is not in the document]
lmlg: Yes, I believe we should use the bare minimum caps needed. If possible, this should be more fine grained.

microceph/api/remote.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	err = RenderConfAndKeyringFiles(req.Name, req.LocalName, req.Config)
+	if err != nil {
+		return response.InternalError(fmt.Errorf("couldn't render files: %w", err))
+	}
+
+	if !req.RenderOnly {
+		// Asynchronously persist this on db and send request to other cluster members.
+		go func() {
```
lmlg, line 50: I'm not sure if we should return before this call completes. Why not make this a synchronous call?

lmlg (changes requested):

UtkarshBhatthere:
This Pr is being closed and moved to #437

PR #409, opened 2024-08-29
Fetch current OSD pool configuration over the API

masnax:
`PUT /1.0/pools-op` lets you set the replication size for several OSD pools. 

This PR adds the ability to fetch the current sizes (and some other relevant OSD pool information) via `GET /1.0/pools`. The goal of this is to obtain the correct OSD pool names to supply to `PUT /1.0/pools-op`, as well as to determine what the current size of each pool is. 

As discussed on mattermost: https://chat.canonical.com/canonical/pl/g3heo1kmz7r6db39re9qyzfw4w

### DIFF
--- a/microceph/api/pool.go
+++ b/microceph/api/pool.go
@@ -15,9 +15,27 @@ import (
 )
 
 // /1.0/pools-op endpoint.
-var poolsCmd = rest.Endpoint{
+var poolsOpCmd = rest.Endpoint{
 	Path: "pools-op",
 	Put:  rest.EndpointAction{Handler: cmdPoolsPut, ProxyTarget: true},
+}
+
+// /1.0/pools endpoint.
+var poolsCmd = rest.Endpoint{
+	Path: "pools",
+	Get:  rest.EndpointAction{Handler: cmdPoolsGet, ProxyTarget: true},
+}
+
+func cmdPoolsGet(s state.State, r *http.Request) response.Response {
+	logger.Debug("cmdPoolGet")
+	pools, err := ceph.GetOSDPools()
+	if err != nil {
+		return response.SmartError(err)
+	}
+
+	logger.Debug("cmdPoolGet done")
+
+	return response.SyncResponse(true, pools)
 }
 
 func cmdPoolsPut(s state.State, r *http.Request) response.Response {
--- a/microceph/ceph/osd.go
+++ b/microceph/ceph/osd.go
@@ -1096,12 +1096,15 @@ func SetReplicationFactor(pools []string, size int64) error {
 
 	if len(pools) == 1 && pools[0] == "*" {
 		// Apply setting to all existing pools.
-		out, err := processExec.RunCommand("ceph", "osd", "pool", "ls")
+		out, err := processExec.RunCommand("ceph", "osd", "pool", "ls", "--format", "json")
 		if err != nil {
 			return fmt.Errorf("failed to list pools: %w", err)
 		}
 
-		pools = strings.Split(out, "\n")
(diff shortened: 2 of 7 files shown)


microceph/ceph/osd.go
```diff
@@ -1117,3 +1117,35 @@ func SetReplicationFactor(pools []string, size int64) error {
...
 }
+
+// GetOSDPools returns a list of OSD Pools and their configurations.
+func GetOSDPools() ([]types.Pool, error) {
+	out, err := processExec.RunCommand("ceph", "osd", "pool", "ls")
+	if err != nil {
+		return nil, fmt.Errorf("failed to list pools: %w", err)
+	}
+
+	poolNames := strings.Split(out, "\n")
```
UtkarshBhatthere, line 1128: Although I agree that parsing the output is simple, still I would prefer this output be read with `--format json` and be unmarshal'd into a `[]string{}`.

what are your thoughts ?

UtkarshBhatthere (commented):
LGTM, one minor nit.

microceph/ceph/osd.go
masnax: Makes sense, I've updated the call in `SetReplicationFactor` to also use this pattern.

masnax:
@UtkarshBhatthere @sabaini @lmlg Sort of a side note/question here: I've been thinking about the `PUT /1.0/pools-op` implementation a bit and I'm wondering if it's a bit aggressive?

Right now it takes 1 size value, and applies it to all specified OSD pools, but also sets the default OSD pool size for all future pools. This means if you want to resize just 1 pool, you have to call this API twice, because the first call will update the default OSD pool size as well, so you'd have to restore that with another request but with a blank set of OSD pools.


I wonder if we can make the default OSD pool size update based on a query parameter? Like `?set-default=1` so that you're not locked into changing the default pool size. To accommodate the query param in the CLI, we can add a `--set-default` flag too, to set the default OSD pool size.


I noticed the CLI help is also wrong. It says `microceph pool set-rf <SIZE> <POOLS>...` but the only arguments it interprets are `<POOLS>...` with `size` being supplied by `--size`.

microceph/ceph/osd.go
UtkarshBhatthere: great :D

UtkarshBhatthere:
@masnax fair enough, can you please create a GH issue for this behaviour ? We will take it up soon.

sabaini (changes requested):
Thanks @masnax , this lgtm. It would be great to have a test for this though

microceph/api/pool.go
```diff
@@ -18,6 +18,19 @@ import (
 var poolsCmd = rest.Endpoint{
 	Path: "pools-op",
```
UtkarshBhatthere, line 19: So the `pools-op` endpoint was created for special ops implemented for Ceph pools. There is no `/pools/` endpoint today, but for this operation, I feel a GET handler there would be more appropriate as when we do come up with the create/delete implementation it would fit nicely. Reservations against going ahead with `pools-op` because it will be LTS and we won't be able to deprecate it all-through 24.04's life.

UtkarshBhatthere (changes requested):
One change requested

microceph/api/pool.go
masnax: So should we leave `PUT` on `pools-op` or move that over to `/pools` as well?

UtkarshBhatthere:
Fair enough @lmlg, I will share a post in Ceph Announcements room with the newly created ticket and the description of the change.

PR #427, opened 2024-09-24
Move multi-node tutorial to how-to section

skoech:
# Description

The guide to installing MicroCeph on multiple machines is currently presented as a tutorial although it is more appropriate as a how-to guide. This change moves it to the how-to guides section of the documentation.

## Type of change


sabaini (changes requested):
This needs a TOC update as well

skoech:
Hey @UtkarshBhatthere! After testing the multi-node tutorial, I thought it didn't qualify to be a tutorial either because it didn’t feel like a learning experience. Rather, I was able to get something done (deploy a multi-node cluster), which is great, but it didn't teach me any MicroCeph concepts e.g. nodes, clusters, why they are useful or even which basic hardware requirements I needed to perform the deployment. It assumed lots of prior knowledge on my part, something an entry-level tutorial shouldn't do.

We need an end-to-end tutorial suitable for new users. I suggested repurposing the previous single-node tutorial because at the end of it, the user will have achieved a simple yet meaningful task (it's also memorable since it takes only 3 commands to set it up) but in order to make it end-to-end, we need to add a prerequisites section at the beginning and a clean-up section at the end. Also, it would be nice to teach the user ways in which they can interact with their new cluster e.g. check health, enable radosgw, and consume their cluster storage e.g. host a static website.

And then, after giving the user this basic introduction, we can now show them HOW TO do the same thing but with multiple nodes.  What do you think about this approach?

sabaini:
thanks, lgtm!

PR #430, opened 2024-09-27
Adds support for RO cluster configs

### LINKED ISSUE #353: Trying to change network binds after bootstrap does not seem to work
# Issue report
Trying to change network binds after bootstrap does not seem to work

quincy/stable (0+git.4a608fc)

```bash
# ip a 
[..]
2: enp5s0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc mq state UP group default qlen 1000
    link/ether 00:16:3e:30:33:01 brd ff:ff:ff:ff:ff:ff
    inet 192.168.10.99/24 metric 100 brd 192.168.10.255 scope global dynamic enp5s0
       valid_lft 572785sec preferred_lft 572785sec
    inet 10.5.0.3/16 scope global enp5s0
       valid_lft forever preferred_lft forever
```

```bash
# microceph bootstrap

# sudo ceph config generate-minimal-conf
[global]
	fsid = d2b14f41-10a0-44e2-96a8-a1f10171e5b9
	mon_host = [v2:192.168.10.99:3300/0,v1:192.168.10.99:6789/0]
```

Now I followed this to change the network bind to another one
https://canonical-microceph.readthedocs-hosted.com/en/quincy-stable/how-to/configure-network-keys/

```bash
# sudo microceph cluster config set public_network 10.5.0.0/16
# sudo microceph cluster config set cluster_network 10.5.0.0/16
# sudo snap restart microceph
2024-05-31T00:06:40Z INFO Waiting for "snap.microceph.mds.service" to stop.
2024-05-31T00:06:42Z INFO Waiting for "snap.microceph.osd.service" to stop.
2024-05-31T00:06:47Z INFO Waiting for "snap.microceph.rgw.service" to stop.
Restarted.
```

But it still binds to the old network
```bash
# sudo ceph config generate-minimal-conf
[global]
	fsid = d2b14f41-10a0-44e2-96a8-a1f10171e5b9
	mon_host = [v2:192.168.10.99:3300/0,v1:192.168.10.99:6789/0]
```

I was expecting to be able to change the network configuration binding after having bootstrapped a ceph cluster to migrate from one network to another

UtkarshBhatthere:
# Description
Adds support for RO cluster configs and exposes public_network as a RO config.

Fixes #353 

## Type of change

- [X] Bug fix (non-breaking change which fixes an issue)

## How Has This Been Tested?

> **_NOTE:_** All functional changes should accompany corresponding tests (unit tests, functional tests etc).

Please describe the addition/modification of tests done to verify this change. Please also list any relevant details for your test configuration.

## Contributor's Checklist

Please check that you have:

- [X] added code comments, particularly in hard-to-understand areas.
- [X] added tests to verify effectiveness of this change.

### DIFF
--- a/microceph/ceph/config.go
+++ b/microceph/ceph/config.go
@@ -18,12 +18,22 @@ import (
 	"github.com/canonical/microceph/microceph/database"
 )
 
+type ClusterConfigPermission string
+
+const (
+	ClusterConfigRO ClusterConfigPermission = "read_only"
+	ClusterConfigRW ClusterConfigPermission = "read_write"
+)
+
+type ClusterConfigDefinition struct {
+	Who        string                  // Ceph Config internal <who> against each key
+	Permission ClusterConfigPermission // read only or read write
+	Daemons    []string                // List of Daemons that need to be restarted across the cluster for the config change to take effect.
+}
+
 // Config Table is the source of additional information for each supported config key
 // Refer to GetConfigTable()
-type ConfigTable map[string]struct {
-	Who     string   // Ceph Config internal <who> against each key
-	Daemons []string // List of Daemons that need to be restarted across the cluster for the config change to take effect.
-}
+type ConfigTable map[string]ClusterConfigDefinition
 
 // Check if certain key is present in the map.
 func (c ConfigTable) isKeyPresent(key string) bool {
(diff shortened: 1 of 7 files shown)


microceph/ceph/config.go
```diff
@@ -93,20 +104,24 @@ type ConfigDumpItem struct {
 }
 type ConfigDump []ConfigDumpItem
 
-func SetConfigItem(c types.Config) error {
-	configTable := GetConstConfigTable()
+func SetConfigItem(c types.Config, bypassChecks bool) error {
```
sabaini, line 107: From a separation of concerns POV I would suggest to -- instead of adding a flag -- split this function up, something like:

```
func SetConfigItemUnsafe(...) {
 // construct args, run ceph command
}

func SetConfigItem(...) {
  // perform checks
  if !canSet && !bypassChecks { 
...
  }

  SetConfigItemUnsafe(...)
}
```

I feel this clarifies the intent and is less error-prone wrt to passing in the flag.

Call sites of these functions would need to be updated too of course.

microceph/ceph/config.go
```diff
@@ -93,20 +104,24 @@ type ConfigDumpItem struct {
...
+	// safety checks
+	canSet, err := canSetConfig(c.Key)
+	if !canSet && !bypassChecks {
+		return err
```
sabaini, line 111: I think adding some context to the error would be good here, sth like

```
return fmt.Errorf("not allowed to set config item ...")
```

sabaini (changes requested):
Heya @UtkarshBhatthere, lgtm in general -- one change request wrt to func signatures 
cheers!

microceph/ceph/config.go
gboutry: I agree with @sabaini, using flags is prone to error, more so from the caller side.

UtkarshBhatthere: Ack.

microceph/ceph/config.go
UtkarshBhatthere: ack, the internal error is also descriptive so I have stacked them.

Issue #424, opened 2024-09-19
Square brackets around IPv6 addresses in `ceph.conf`


boltmark:
# Issue report

## What version of MicroCeph are you using?

`latest/edge` (19.2.0~git+snapbfd2c916c8)

## What are the steps to reproduce this issue ?

See https://github.com/canonical/microcloud/issues/380 for the steps taken to produce the error (during `microcloud init` questionnaire, set up CephFS and use IPv6).

Can also see a similar failure when setting up LXD with a CephFS storage pool via MicroCeph:
1. Create VM
```
lxc init ubuntu-minimal:24.04 v1 --vm -c limits.cpu=2 -c limits.memory=4GiB
```
2. Add volume to be used as disk
```
lxc storage volume create default disk1 size=5GiB --type=block
lxc config device add v1 disk1 disk pool=default source=disk1
```
3. Start VM and note IPv6 for `enp5s0`
```
lxc start v1
lxc ls v1
```
4. Enter VM and install MicroCeph and LXD
```
lxc exec v1 bash
snap install lxd --channel latest/edge
snap install microceph --channel latest/edge
```
5. Bootstrap MicroCeph with IPv6 address saved from above. Set up a new cluster, and add the disk created above
```
microceph init
```
6. Allow fewer than 3 disks for the sake of the reproducer
```
microceph pool set-rf "" --size=1
```
7. Initialize LXD and attempt to create CephFS storage pool
```
lxd init
lxc storage create lxd_cephfs cephfs cephfs.create_missing=true cephfs.meta_pool=lxd_cephfs_meta cephfs.data_pool=lxd_cephfs_data source=lxd_cephfs
```

## What happens (observed behaviour) ?

As per https://github.com/canonical/microcloud/issues/380, failure to mount error is observed:
`Error: Failed to mount "fd42:4c88:a791:d989:216:3eff:fe97:243c:6789,fd42:4c88:a791:d989:216:3eff:fe6d:c28e:6789:/" on "/tmp/lxd_cephfs_3143952555/mount" using "ceph": invalid argument`

## Additional comments.
[...]

PR #432, opened 2024-10-01
Use square brackets around IPv6 in ceph.conf

boltmark:
Small PR which adds square brackets around mon host addresses in `ceph.conf` if IPv6 is used.

Closes https://github.com/canonical/microceph/issues/424.

### DIFF
--- a/microceph/ceph/config.go
+++ b/microceph/ceph/config.go
@@ -356,6 +356,9 @@ func UpdateConfig(ctx context.Context, s interfaces.StateInterface) error {
 		}
 	}
 
+	// Ensure that IPv6 addresses have square brackets around them (if IPv6 is used).
+	monitorAddresses = formatIPv6(monitorAddresses)
+
 	conf := NewCephConfig(constants.CephConfFileName)
 
 	// Check if host has IP address on the configured public network.
@@ -441,4 +444,19 @@ func getMonitorAddresses(configs map[string]string) []string {
 		}
 	}
 	return monHosts
+}
+
+// formatIPv6 returns a slice in which all IPv6 addresses are formatted with square brackets.
+func formatIPv6(addrs []string) []string {
+	formatted := []string{}
+	for _, addr := range addrs {
+		ip := net.ParseIP(addr)
+		if ip != nil && strings.Contains(addr, ":") {
+			addr = fmt.Sprintf("[%s]", addr)
+		}
+
+		formatted = append(formatted, addr)
+	}
+
+	return formatted
 }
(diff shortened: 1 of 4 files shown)


sabaini (changes requested):
Hey @boltmark , thanks for this PR! One comment inline. 

Also, would you be able to add some tests around this so we can be confident the change addresses the issue?

microceph/ceph/config.go
```diff
@@ -281,6 +281,17 @@ func UpdateConfig(ctx context.Context, s interfaces.StateInterface) error {
...
+	for i, addr := range monitorAddresses {
+		if !strings.Contains(addr, ":") {
+			continue
+		}
+
+		if !strings.HasPrefix(addr, "[") && !strings.HasSuffix(addr, "]") {
+			monitorAddresses[i] = fmt.Sprintf("[%s]", addr)
+		}
+	}
+
```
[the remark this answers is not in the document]
boltmark: Thanks for the comment -- factored this out to a small function.

microceph/ceph/config.go
```diff
@@ -367,3 +370,17 @@ func getMonitorAddresses(configs map[string]string) []string {
 	}
 	return monHosts
 }
+
+// formatIPv6 returns a slice in which all IPv6 addresses are formatted with square brackets.
+func formatIPv6(addrs []string) []string {
+	formatted := []string{}
+	for _, addr := range addrs {
+		if strings.Contains(addr, ":") && !strings.HasPrefix(addr, "[") && !strings.HasSuffix(addr, "]") {
```
UtkarshBhatthere, line 378: something like this would be more conclusive:
```
ip := net.ParseIP(addr)
if ip != nil && strings.Contains(addr, ":") {
    addr = fmt.Sprintf("[%s]", addr)
}
```

microceph/ceph/config.go
```diff
@@ -367,3 +370,18 @@ func getMonitorAddresses(configs map[string]string) []string {
 	}
 	return monHosts
 }
+
+// formatIPv6 returns a slice in which all IPv6 addresses are formatted with square brackets.
+func formatIPv6(addrs []string) []string {
+	formatted := []string{}
+	for _, addr := range addrs {
+		ip := net.ParseIP(addr)
+		if ip != nil && strings.Contains(addr, ":") {
```
lmlg, line 379: should the ip be added to the list if parsing it returned `nil`?

boltmark: Hi @lmlg, `net.ParseIP` will return nil if the IPv6 address is already enclosed in square brackets, so we would fail to add the addresses in this case. If it is not possible for this to happen, let me know and I'll make this change.

lmlg: If we're certain that `ParseIP` can only fail for that reason, then it's fine by me.

PR #440, opened 2024-10-14
Remove tutorial section

skoech:
# Description

Since we no longer have a tutorial, I am removing this section for the time being as we work on a new tutorial.

## Type of change


lmlg (approved):
lgtm, thanks :)

PR #442, opened 2024-10-15
Added QAT configs with experimental flag.

UtkarshBhatthere:
# Description

Added cluster configs for QAT enablement

## Type of change

## How Has This Been Tested?
Manual Testing

## Contributor's Checklist

Please check that you have:


microceph/ceph/config.go
```diff
@@ -222,6 +233,10 @@ func setConfigItem(c types.Config) error {
 		return err
 	}
 
+	if config.Permission == ClusterConfigEXPRW {
+		return fmt.Errorf("WARNING: set operation on experimental config (%s)", c.Key)
```
lmlg, line 237: I would advise against doing this. The pattern `if err != nil` is way too common in Go, and overloading errors for warnings seems like a footgun to me. Instead, I would simply log the warnings here instead of down below.

lmlg (changes requested):

microceph/ceph/config.go
UtkarshBhatthere: The problem with logging is that the end user will not know about it until they check the logs. An acceptable (rather better) alternative is to require `--experimental` flag for such configs.

UtkarshBhatthere:
Blocked as we do not have a the QAT changes in the distribution packages yet.

PR #454, opened 2024-11-07
Pin LXD and use microcluster with dqlite LTS

roosterfish:
### Pin LXD

When defining the microcluster API handlers you have to import from `"github.com/canonical/lxd/lxd/response"` for the handlers return value.
As this dependency can diverge from the same import that microcluster users internally, you might end up with build issues if you don't use the same version as microcluster.

We have already [filed an improvement](https://github.com/canonical/microcluster/issues/282) in microcluster to address this in future releases. 
But for now pinning down the version is a good measure to prevent build errors in the future during maintenance of the LTS.
The same measure is [applied in microcluster](https://github.com/canonical/microcluster/blob/v2/Makefile#L51) to ensure every downstream can stay on track with the used version of LXD.

See [this PR](https://github.com/canonical/microcloud/pull/464) for the same change in MicroCloud and [this PR](https://github.com/canonical/microovn/pull/202) for MicroOVN.

### Dqlite LTS

In addition microcluster is updated to `v2.0.5` which uses dqlite `v2` LTS.


roosterfish:
Should I open this PR against another LTS branch (I can only see `squid/devel`)?

sabaini:
Thank you @roosterfish

PR #463, opened 2024-11-11
Fix: add python3-packaging

sabaini:
Prometheus module needs python3-packaging


lmlg (approved):
lgtm. I'd go ahead and merge since it's pointless to wait for the tests to run with a simple a change as this one.

PR #482, opened 2024-12-19
Introduce maintenance mode

chanchiwai-ray:
# Description

MicroCeph currently supports scaling up and down, but not maintenance. In the actual operation, the operators may also need to perform regular maintenance to the storage nodes, and MicroCeph does not offer a simple solution to this problem. We would like to introduce sub-commands (e.g. microceph cluster maintenance enter and microceph cluster maintenance exit) to MicroCeph to enable the feature.

## Type of change

- [x] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?

> **_NOTE:_** All functional changes should accompany corresponding tests (unit tests, functional tests etc).

Tested in functional tests, and also manual test with 4 microceph nodes. 

## Contributor's Checklist

Please check that you have:

- [x] added code comments, particularly in hard-to-understand areas.
- [ ] updated the user documentation with corresponding changes. (will do in separated PR)
- [x] added tests to verify effectiveness of this change.


microceph/ceph/maintenance.go
```diff
@@ -0,0 +1,297 @@
...
+			fmt.Println(ops.DryRun(m.nodeName))
+		} else {
+			err := ops.Run(m.nodeName)
+			if err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
```
jneo8, line 89: **suggestion:** 

*maintenance* struct seems not necessary here.
I think it can be merge to one simple function *RunOperations*

microceph/ceph/maintenance.go
```diff
@@ -0,0 +1,297 @@
...
+//
+// operations
+//
+
+type operation interface {
+	Run(string) error
+	DryRun(string) string
+}
+
+type checkNodeInClusterOps struct {
```
jneo8, line 100: **suggestion:** 

Suggestion to explain the purpose of operation here.

microceph/ceph/osd.go
```diff
@@ -954,27 +954,55 @@ func outDownOSD(osd int64) error {
 	return nil
 }
 
-func safetyCheckStop(osd int64) error {
+func osdNooutFlag(set bool) error {
```
jneo8, line 957: **nitpick:** 

Change function name to `setOSDNoOut`?
End with *Flag* seems like an object.

microceph/ceph/osd.go
```diff
@@ -954,27 +954,55 @@ func outDownOSD(osd int64) error {
...
+	output, err := processExec.RunCommand("ceph", "osd", "dump")
+	if err != nil {
+		logger.Errorf("Failed to dump osd info: %v", err)
+		return false, fmt.Errorf("Failed to dump osd info: %w", err)
+	}
+	logger.Infof("OSD dump: %s", output)
+	return strings.Contains(output, "noout"), nil
+}
+
+func safetyCheckStop(osds []int64) error {
```
jneo8, line 985: **praise:** 

I think you are doing the right approach.

microceph/ceph/maintenance.go
```diff
@@ -0,0 +1,297 @@
...
+		}
+	}
+	return nil
+}
+
+//
+// operations
+//
+
+type operation interface {
```
jneo8, line 95: **suggestion:** 

I more prefer to put all the operation to another file instead of `maintenance.go`.

microceph/ceph/maintenance.go
```diff
@@ -0,0 +1,297 @@
...
+	ops := []operation{}
+
+	// pre-flight checks
+	if !force {
+		ops = append(ops, []operation{
+			&checkNodeInClusterOps{cephClient, clusterClient},
+			&checkOsdOkToStopOps{cephClient, clusterClient},
+			&checkNonOsdSvcEnoughOps{cephClient, clusterClient, 3, 1, 1},
+		}...)
+	}
```
jneo8, line 24: **question:** 

Should we should still run the pre flight-check but ignore the error?

jneo8 (commented):
Overall no big issue. Only some suggestions.

microceph/ceph/maintenance.go
chanchiwai-ray: good question, how do we want to define "force"? ignore all errors or skip preflight checks?

microceph/api/osds.go
```diff
@@ -0,0 +1,51 @@
+package api
```
jneo8, line 1: **suggestion (non-blocking):** 

It will be nice if you can separate the API part to another PR, which may be more easy for reviewer.

jneo8 (approved):
Thanks, LGTM.

microceph/ceph/maintenance.go
jneo8: Can be resolved.

microceph/ceph/maintenance.go
jneo8: Can be resolved.

microceph/ceph/osd.go
jneo8: Can be resolved.

microceph/ceph/maintenance.go
jneo8: Can be resolved.

microceph/ceph/maintenance.go
jneo8: We still run the pre-flight checks but ignore the error.
The ignore error should be logged.

microceph/ceph/osd.go
jneo8: Can be resolved.

microceph/api/types/osds.go
```diff
@@ -0,0 +1,8 @@
+// Package types provides shared types and structs.
+package types
+
+// OsdPut holds data structure for updating the state of osd service
+type OsdPut struct {
+	State    string `json:"state" yaml:"state"`
+	Location string `json:"location" yaml:"location"`
```
sabaini, line 7: I can see the Location field being updated in the client code, but on server (API) side it's not being used. 
I think we could omit the Location field in this struct and instead rely on the targetting mechanism of microcluster; by convention cli commands that support this have a `--target` argument to address a specific node.

I see below you're already running the c.UseTarget() method, that should take care of the location.

microceph/ceph/operations.go
```diff
@@ -0,0 +1,262 @@
...
+
+	// DryRun returns the string representation of the operation.
+	DryRun(string) string
+}
+
+// CheckNodeInClusterOps is an operation to check if a node is in the microceph cluster.
+type CheckNodeInClusterOps struct {
+	CephClient    client.ClientInterface
+	ClusterClient *microCli.Client
+}
```
sabaini, line 42: Suggestion: introduce a base class `ClusterOps` with the CephClient and ClusterClient fields, then all the other classes / structs could inherit from `ClusterOps`

microceph/ceph/operations.go
```diff
@@ -0,0 +1,262 @@
...
+		return fmt.Errorf("osd.%v cannot be safely stopped", OsdsToCheck)
+	}
+
+	logger.Infof("osd.%v can be safely stopped.", OsdsToCheck)
+	return nil
+}
+
+// DryRun prints out the action plan.
+func (o *CheckOsdOkToStopOps) DryRun(name string) string {
+	return fmt.Sprintf("Check if osds in node '%s' are ok-to-stop.", name)
```
sabaini, line 96: Suggest to list which OSDs exactly are going to be stopped

sabaini (changes requested):
Hey @chanchiwai-ray thanks for this, think this is a sound approach! I've did a first pass on review and left some suggestions / nits.

Also it would be good to have functional testing on this too.

microceph/ceph/maintenance.go
chanchiwai-ray: applied

microceph/ceph/operations.go
```diff
@@ -0,0 +1,284 @@
...
+	ClusterClient *microCli.Client
+}
+
+// CheckNodeInClusterOps is an operation to check if a node is in the microceph cluster.
+type CheckNodeInClusterOps struct {
+	ClusterOps
+}
+
+// Run checks if a node is in the microceph cluster.
+func (o *CheckNodeInClusterOps) Run(name string) error {
```
sabaini, line 54: I wonder if its necessary to go via the API client here. At this point we're in server code and could call the resp. handler directly I think. 
Similarly in the below client calls

microceph/cmd/microceph/cluster_maintenance_exit.go
```diff
@@ -0,0 +1,64 @@
...
+
+type cmdClusterMaintenanceExit struct {
+	common *CmdControl
+
+	flagDryRun bool
+}
+
+func (c *cmdClusterMaintenanceExit) Command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "exit <NAME>",
```
sabaini, line 20: What is NAME here?

tests/scripts/actionutils.sh
```diff
@@ -110,6 +110,16 @@ function enable_rgw() {
     wait_for_rgw 1
 }
 
+function enable_mon() {
+    set -x
+    sudo microceph enable mon
+}
+
+function enable_mon() {
```
sabaini, line 118: Duplicate

tests/scripts/actionutils.sh
```diff
@@ -824,6 +851,198 @@ function bombard_rgw_configs() {
...
+    set -xe
+
+    # Debug
+    nodeexec $node "microceph status"
+    nodeexec $node "microceph.ceph -s"
+
+    # Count expected steps when --set-noout=false --stop-osds=false
+    lines=$(nodeexec $node "microceph cluster maintenance enter $node --set-noout=false --stop-osds=false --dry-run | tee output.txt | wc -l")
+    nodeexec $node "cat output.txt"
+    if [ $lines != "3" ]; then exit 1; fi
```
sabaini, line 888: It would be great if the test func would log some additional details on the failure to make the issue more obvious / easier to debug. Could e.g. log the contents of the output.txt

Similarly in the tests below.

sabaini (changes requested):
Hey @jneo8 thanks for these!
Some suggestions inline.

I'd also like to suggest running an API test against the newly created /osds endpoint. See the hurl() function and tests/hurl dir. I think a simple dry-run test would be sufficient.

.github/workflows/tests.yml
```diff
@@ -434,6 +434,90 @@ jobs:
...
+    - name: Add and wait for OSDs
+      run: |
+        set -uex
+        for i in 0 1 2 3 ; do
+          ~/actionutils.sh add_osd_to_node "node-wrk$i"
+        done
+        ~/actionutils.sh headexec wait_for_osds 4
+
+    - name: Wait for ceph cluster to settle
+      run: sleep 30
```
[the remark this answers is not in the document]
chanchiwai-ray: Thanks, I didn't notice the `verify_health` function

tests/scripts/actionutils.sh
chanchiwai-ray: Yeah, thanks for spotting it

microceph/ceph/operations.go
```diff
@@ -0,0 +1,284 @@
...
+type CheckNonOsdSvcEnoughOps struct {
+	ClusterOps
+
+	MinMon int
+	MinMds int
+	MinMgr int
+}
+
+// Run checks if non-osds service in a node are enough.
+func (o *CheckNonOsdSvcEnoughOps) Run(name string) error {
```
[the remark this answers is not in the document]
chanchiwai-ray: Yes! My thought is to remove the duplication when we implement additional checks to remove node logic (as we discussed privately)

microceph/ceph/operations.go
chanchiwai-ray: Sorry, I am not sure what do you mean here, can you elaborate more?

> At this point we're in server code

The operations are not run in server side (i.e. `microcephd`)

tests/scripts/actionutils.sh
chanchiwai-ray: Added

microceph/cmd/microceph/cluster_maintenance_exit.go
chanchiwai-ray: it the name of the node. I made it more explicit now

microceph/ceph/operations.go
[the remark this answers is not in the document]
sabaini: Ack.

microceph/cmd/microceph/cluster_maintenance_enter.go
```diff
@@ -0,0 +1,79 @@
...
+			&ceph.AssertNooutFlagSetOps{ClusterOps: clusterOps},
+		}...)
+	}
+
+	// optionally stop osd service
+	if c.flagStopOsds {
+		operations = append(operations, []ceph.Operation{
+			&ceph.StopOsdOps{ClusterOps: clusterOps},
+		}...)
+	}
```
samuelallan72, line 71: I'm curious, if these are optional, what actually happens if the user runs `microceph maintenance enter --set-noout=false --stop-osds=false`? What does maintenance mode mean without those?

microceph/ceph/osd.go
```diff
@@ -954,27 +954,55 @@ func outDownOSD(osd int64) error {
 	return nil
 }
 
-func safetyCheckStop(osd int64) error {
+func setOsdNooutFlag(set bool) error {
```
samuelallan72, line 957: This might be clearer as two functions:

```
func setOsdNooutFlag() error
func UnsetOsdNooutFlag() error
```

Then we don't need to deal with a boolean argument.

microceph/ceph/operations.go
```diff
@@ -0,0 +1,284 @@
...
+		return err
+	}
+	logger.Info("unset osd noout.")
+	return nil
+}
+
+// DryRun prints out the action plan.
+func (o *UnsetNooutOps) DryRun(name string) string {
+	return "Run `ceph osd unset noout`."
+}
```
samuelallan72, line 284: There's a little bit of abstraction leaking between Run and DryRun here: Run uses the setOsdNooutFlag() function, but DryRun prints the lower level ceph command used. Just an observation. Not sure how nitpicky this is, or the best way to go - perhaps inlining the setOsdNooutFlag function to remove the extra layer of indirection?

microceph/cmd/microceph/cluster_maintenance_enter.go
chanchiwai-ray: it will then be only checks

microceph/ceph/operations.go
chanchiwai-ray: The `DryRun` function should give you the actual action performed, not the string representation of the function

microceph/ceph/operations.go
chanchiwai-ray: I believe this is addressed

microceph/ceph/operations.go
chanchiwai-ray: Should be addressed in the recent commits

microceph/ceph/osd.go
chanchiwai-ray: That's my initial implementation, but it turns out it's just simple to add a boolean argument to remove the all-most-identical code. Moreover, I think the function name is a bit clear now: "set the osd noout flag" true means set the flag, false means don't set the flag.

microceph/api/ops_maintenance.go
```diff
@@ -0,0 +1,65 @@
...
+	case "non-maintenance":
+		results = maintenance.Exit(maintenancePut.DryRun)
+	default:
+		err = fmt.Errorf("unknown status encounter: '%s', can only be 'maintenance' or 'non-maintenance'", status)
+		return response.BadRequest(err)
+	}
+
+	for _, result := range results {
+		if result.Error != "" && !maintenancePut.Force {
+			return response.SyncResponse(false, results)
```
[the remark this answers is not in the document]
chanchiwai-ray: updated to customized response which will failed with HTTP 400 (not just in the metadata) and return the results in metadata

chanchiwai-ray:
Not sure why DCO is failing

microceph/api/ops_maintenance.go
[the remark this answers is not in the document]
chanchiwai-ray: Example 

```
> PUT /1.0/ops/maintenance/microceph-2 HTTP/1.1                                                                                                                                                                                                                                           > Host: localhost                                                     
> User-Agent: curl/7.81.0                                             
> Accept: */*                                                                                                                                                                                                                                                                             
> Content-Type: application/json                                                                                                             
> Content-Length: 78                                                                                                                         
>                                                                     
* Mark bundle as not supporting multiuse                                                                                                     
< HTTP/1.1 400 Bad Request                                                                                                                   
< Content-Type: application/json                                      
< Date: Wed, 22 Jan 2025 11:42:48 GMT                                                                                                        
< Content-Length: 498                                                                                                                        
<                                                                                                                                            
{"type":"sync","status":"Failure","status_code":400,"operation":"","error_code":0,"error":"","metadata":[{"name":"check-osd-ok-to-stop-ops","error":"","action":"Check if osds.[5 21 23] in node 'microceph-2' are ok-to-stop."},{"name":"check-non-osd-svc-enough-ops","error":"need at l
east 3 mon, 1 mds, and 1 mgr services in the cluster besides those in node 'microceph-2'","action":"Check if there are at least 3 mon, 1 mds, and 1 mgr services in the cluster besides those in node 'microceph-2'"}]}
```

microceph/ceph/maintenance.go
```diff
@@ -0,0 +1,104 @@
...
+import (
+	"fmt"
+)
+
+type Maintenance struct {
+	Node       string
+	ClusterOps ClusterOps
+}
+
+func (m *Maintenance) Exit(dryRun, checkOnly, ignoreCheck bool) ([]Result, error) {
```
sabaini, line 12: It would be good to have some func docs here

microceph/ceph/maintenance.go
```diff
@@ -0,0 +1,104 @@
...
+			}
+		}
+		// Otherwise, continue with the main operations
+		results = append(results, RunOperations(m.Node, operations, dryRun, false)...)
+	}
+
+	return results, nil
+}
+
+func (m *Maintenance) Enter(force, dryRun, setNoout, stopOsds, checkOnly, ignoreCheck bool) ([]Result, error) {
```
sabaini, line 53: It would be good to have some func docs here

microceph/ceph/maintenance.go
```diff
@@ -0,0 +1,104 @@
...
+	// General sanity checks
+	if checkOnly && ignoreCheck {
+		err := fmt.Errorf("'check_only' and 'ignore_check' are mutually exclusive.")
```
sabaini, line 56: Spelling nit: should be `check-only` and `ignore-check`

sabaini (changes requested):
Thanks @chanchiwai-ray , this is looking pretty good. I only have added some doc requests and spel nitpickery

One thing we will need is user facing docs for this functionality but this could be a separate PR

microceph/ceph/maintenance.go
chanchiwai-ray: Fixed

microceph/ceph/maintenance.go
chanchiwai-ray: added

sabaini (approved):
Thanks @chanchiwai-ray , lgtm

sabaini:
> Not sure why DCO is failing

The error message says "Argument list too long" and from the log it looks like it passes commit metadata as args. As there's quite a few commits contained in the PR I have a feeling this is what is getting overrun. You might need to squash the changes to make that test green.

chanchiwai-ray:
Squashed all commits into one

microceph/api/types/maintenance.go
```diff
@@ -0,0 +1,31 @@
...
+
+// Options for bringing a node into maintenance
+type MaintenanceEnterFlags struct {
+	Force    bool `json:"force"`
+	SetNoout bool `json:"set_noout"`
+	StopOsds bool `json:"stop_osds"`
+}
+
+// MaintenancePut holds data structure for bringing a node into or out of maintenance
+type MaintenancePut struct {
```
UtkarshBhatthere, line 27: Naming nit: Suggestion to rename the structs as follows:
```
MaintenancePut -> MaintenanceRequest
MaintenanceFlags -> CommonMaintenanceFlags
MaintenanceEnterFlags -> EnterMaintenanceFlags
```

microceph/api/ops_maintenance.go
```diff
@@ -0,0 +1,119 @@
...
+		ClusterOps: ceph.ClusterOps{
+			State:   s,
+			Context: r.Context(),
+		},
+	}
+
+	status := maintenancePut.Status
+	switch status {
+	case "maintenance":
+		results, err = maintenance.Enter(maintenancePut.Force, maintenancePut.DryRun, maintenancePut.SetNoout, maintenancePut.StopOsds, maintenancePut.CheckOnly, maintenancePut.IgnoreCheck)
```
UtkarshBhatthere, line 101: I see that this method is only being used here, Why not just pass the whole structure ?

UtkarshBhatthere (commented):
some comments, still reviewing.

tests/hurl/osds-put-failed.hurl
```diff
@@ -0,0 +1,8 @@
+PUT http://localhost/1.0/osds
```
UtkarshBhatthere, line 1: +1 for using HURL api tests.

tests/scripts/actionutils.sh
```diff
@@ -445,7 +450,7 @@ function upgrade_multinode() {
         done
         res=$( ( lxc exec $container -- sh -c "microceph.ceph osd status" | fgrep -c "exists,up" ) )
         if [[ $res -ne $expect ]] ; then
-            echo "Expected $expect OSD up, got $res"
```
UtkarshBhatthere, line 448: please refrain from modifying unrelated lines specially in larger PRs.

tests/scripts/actionutils.sh
```diff
@@ -639,7 +644,7 @@ function wait_for_osds() {
     if [[ $res -lt $expect ]] ; then
         echo "Never reached ${expect} OSDs"
         return -1
-    fi
```
UtkarshBhatthere, line 642: I appreciate such fixes but ideally they should be carried out in separate PRs.

microceph/client/wrap.go
```diff
@@ -55,5 +56,10 @@ func (c ClientImpl) DeleteClusterMember(cli *microCli.Client, name string, force
 	return cli.DeleteClusterMember(context.Background(), name, force)
 }
 
+// PutOsds wraps the PutOsds function
+func (c ClientImpl) PutOsds(cli *microCli.Client, up bool, target string) error {
```
UtkarshBhatthere, line 60: I have some doubts,
1. Where is this method being used ?
2. Shouldn't the client functions for maintenance API be also a part of this interface ?

UtkarshBhatthere (changes requested):
need some clarifications.

microceph/client/wrap.go
chanchiwai-ray: oh, I guess you are right! These (api + client wrapper + mock interface for osds) WERE used before, and now they are not used. I will remove them

tests/scripts/actionutils.sh
chanchiwai-ray: Thanks, that's me being lazy. Will definitely do next time

tests/scripts/actionutils.sh
chanchiwai-ray: These were really unexpected change, probably I did a '%s/Expected/Expect/g`

microceph/api/types/maintenance.go
chanchiwai-ray: Updated

microceph/api/ops_maintenance.go
chanchiwai-ray: Updated

tests/scripts/actionutils.sh
chanchiwai-ray: Reverted this part

microceph/client/wrap.go
chanchiwai-ray: Removed

tests/scripts/actionutils.sh
UtkarshBhatthere: good editor 😃

tests/scripts/actionutils.sh
UtkarshBhatthere: Thanks

microceph/api/types/maintenance.go
UtkarshBhatthere: Ack, thanks

microceph/api/ops_maintenance.go
UtkarshBhatthere: Thanks

microceph/client/wrap.go
UtkarshBhatthere: Thanks

PR #492, opened 2025-01-28
Use the less common TCP port 18000

skoech:
# Description

When trying to run my local build of the docs, I run into this:

`ERROR:    [Errno 98] error while attempting to bind on address ('127.0.0.1', 8000): address already in use
make: *** [Makefile:62: run] Error 1`

Port 8000 seems to be a poor choice to use as it is very common, so I have modified the Makefile to use a different, less common port.

## Type of change

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -80,7 +80,7 @@ epub: install
 
 
 serve: html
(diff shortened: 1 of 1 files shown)


sabaini (approved):
lgtm!

PR #493, opened 2025-01-28
docs: refer to command microceph.ceph instead of ceph

xypron:
There is no /snap/bin/ceph. But /snap/bin/microceph.ceph exists. Reflect this in the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ your root drive:
 
 You're done! Check Ceph status:
 
-    sudo ceph status
+    sudo microceph.ceph status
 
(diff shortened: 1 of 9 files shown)


UtkarshBhatthere:
hello @xypron we do not intend to use the `ceph` client as `microceph.ceph`. If you install microceph from snapstore (`sudo snap install microceph`) it should always come with the aliases pre-created. Otherwise you can also do `sudo snap alias microceph.ceph ceph`) if you installed a locally built version on microceph.

PR #535, opened 2025-04-02
Fix: Wait for monitor responsiveness during bootstrap

### LINKED ISSUE #473: Request timeout during cluster bootstrap
As part of the MicroCloud test suite we are deploying single node MicroCeph clusters in a lot of situations to simulate already existing MicroCeph clusters on different nodes.
We do this by running `microceph cluster bootstrap`.

On possibly slow systems we recently saw that firing requests against the `/1.0/configs` API right after `microceph cluster bootstrap` returns might not always succeed and run into a timeout as the context deadline of the respective MicroCeph [GetConfig](https://github.com/canonical/microceph/blob/main/microceph/client/configs.go#L37) client function expires. 
MicroCloud is using this endpoint to retrieve information about MicroCeph required for bootstrapping.

We found that in the pipeline blocking until `microceph status` reports `Services: mds, mgr, mon` will fix it but it looks `microceph cluster bootstrap` could wait a bit longer and return only if the API can respond.

I made the following test and hit the `/1.0/configs` endpoint with requests whilst running `microceph cluster bootstrap` on another terminal:

```bash
# Start this before running the bootstrap
while true; do curl --unix-socket /var/snap/microceph/common/state/control.socket http:/ceph/1.0/configs -X GET -d '{}' -s | awk '{print strftime("%r") " " $1}' | tee -a log; sleep .1; done

# In another terminal
microceph cluster bootstrap
```

See the log here [microceph_bootstrap_monitor.txt](https://github.com/user-attachments/files/17817353/microceph_bootstrap_monitor.txt)
Something interesting (maybe the actual issue) is around 03:08:26 PM. There you see a window of around 2 seconds in which the request didn't receive a response.
The `microceph cluster bootstrap` command was started around this time.
It could be that on slow systems this extends the 5s timeout which is set in the `GetConfig` MicroCeph client func.

sabaini:
# Description

Wait for MON to become active before returning after bootstrapping

Fixes #473 

## Type of change

## How Has This Been Tested?

Updated unit tests

## Contributor's Checklist

Please check that you have:

- [x] added code comments, particularly in hard-to-understand areas.

### DIFF
--- a/microceph/ceph/bootstrap.go
+++ b/microceph/ceph/bootstrap.go
@@ -9,7 +9,9 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
+	"github.com/canonical/lxd/shared/logger"
 	"github.com/canonical/microceph/microceph/constants"
 	"github.com/canonical/microceph/microceph/interfaces"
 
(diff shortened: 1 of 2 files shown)


lmlg (approved):
lgtm

PR #545, opened 2025-04-17
Microcloud integration annotations

minaelee:
# Description

The MicroCeph docs set is pulled in as part of the [MicroCloud integrated docs set](https://canonical-microcloud.readthedocs-hosted.com/en/latest/microovn/), which allows the user to switch between the MicroCloud, LXD, MicroCeph, and MicroOVN products from a header menu.

For MicroCloud users, some pages of the other docs sets are not relevant (such as installation docs, which MicroCloud handles). This PR adds a `note`-type annotation to such pages advising users of this, which uses the `only` Sphinx directive to _only_ show those notes when the page is viewed as part of the MicroCloud integrated docs set. It has no effect on the standalone MicroCeph docs set.

## Type of change

- [X] Documentation update (Doc only change)

## How Has This Been Tested?

I tested manually in local build, both as a standalone docs set and how it's viewed when pulled into a MicroCloud integrated docs set. It behaves as expected: invisible in the standalone, visible in the integrated.

### DIFF
--- a/docs/how-to/multi-node.rst
+++ b/docs/how-to/multi-node.rst
@@ -2,6 +2,15 @@
 Multi-node install
 ==================
 
+.. The note below is only visible when viewed through MicroCloud's docs site:
+   https://documentation.ubuntu.com/microcloud/latest/microceph/
+.. only:: integrated
+
+   .. admonition:: For MicroCloud users
+      :class: note
+      
+      The MicroCloud setup process handles MicroCeph installation. Thus, you do not need to follow the steps on this page.
+
 This tutorial will show how to install MicroCeph on three machines,
 thereby creating a multi-node cluster. For this tutorial, we will
 utilise physical block devices for storage.
--- a/docs/how-to/single-node.rst
+++ b/docs/how-to/single-node.rst
@@ -1,6 +1,15 @@
 How to install MicroCeph on a single node
 =========================================
 
+.. The note below is only visible when viewed through MicroCloud's docs site:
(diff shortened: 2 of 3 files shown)


skoech (approved):
LGTM, thanks for doing this, @minaelee.

Just one concern (unrelated to the annotations): on the MicroCeph landing page, there's a misplaced line before the four top statements.

minaelee:
> LGTM, thanks for doing this, @minaelee.
> 
> Just one concern (unrelated to the annotations): on the MicroCeph landing page, there's a misplaced line before the four top statements.

Thanks @skoech! This is ready for re-review; I've updated the annotations to use a custom title ("For MicroCloud users" instead of "Note"), slightly reworded the content, and added inline comments to explain it for other contributors viewing the source. 

Do you still see a misplaced line on the landing page? What does the line look like?

skoech:
Looks really good, @minaelee! Good idea to specify that the annotation is "For MicroCloud users" to draw their attention, and emphasise that it's not just a general note. I also like that you added an informative note in the landing page regarding installation instructions and instructions for removing cluster members. I'm approving.

skoech (approved):
Looks really good, @minaelee! Good idea to specify that the annotation is "For MicroCloud users" to draw their attention, and emphasise that it's not just a general note. I also like that you added an informative note in the landing page regarding installation instructions and instructions for removing cluster members.

Also, I no longer see the misplaced line, thanks. :)

PR #546, opened 2025-04-23
Added description for monitor address value

UtkarshBhatthere:
# Description

Added a decription for where the IP address value should be taken 

## Type of change

- [X] Documentation update (Doc only change)

## How Has This Been Tested?

Doc only change

## Contributor's Checklist

Please check that you have:

### DIFF
--- a/docs/how-to/mount-block-device.rst
+++ b/docs/how-to/mount-block-device.rst
@@ -98,12 +98,15 @@ Moving forward, we will assume that these files are located at mentioned path.
 
 Map the RBD image on client:
 
+> Note
+> The monitor address is required to be provided for the rbd map command. Replace `192.168.X.Y` with relevant value from your cluster.
+
 .. code-block:: none
 
     $ sudo rbd map \
         --image bd_foo \
         --name client.admin \
-        -m 192.168.29.152 \
+        -m 192.168.X.Y \
         -k /var/snap/microceph/current/conf/ceph.keyring \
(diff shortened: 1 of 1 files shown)


docs/how-to/mount-block-device.rst
```diff
@@ -98,12 +98,15 @@ Moving forward, we will assume that these files are located at mentioned path.
 
 Map the RBD image on client:
 
+> Note
+> The monitor address is required to be provided for the rbd map command. Replace `192.168.X.Y` with relevant value from your cluster.
```
sabaini, line 102: To be extra explicit suggest to spell it out e.g.

> The monitor address is required to be provided for the rbd map command. Replace `192.168.X.Y` with the `mon host` value from your clusters `ceph.conf` file.

sabaini (changes requested):
Thanks, small suggestion inline

docs/how-to/mount-block-device.rst
medvied: The command also works for me (microceph (squid/stable) 19.2.0+snap3b53da1c21, ceph-common 19.2.0-0ubuntu0.24.04.2) without providing a monitor address:
```
$ sudo rbd map --image test_image_100G --name client.admin -k /var/snap/microceph/current/conf/ceph.keyring -c /var/snap/microceph/current/conf/ceph.conf -p block_pool /dev/rbd0
/dev/rbd0
```

Issue #534, opened 2025-04-01
[Adjustment] maintenance pre-flight check adjustment on CheckNonOsdSvcEnoughOps


jneo8:
The current behavior of CheckNonOsdSvcEnoughOps is to verify whether at least three Ceph MON services remain active—excluding the target node—after maintenance mode is enabled.

However, this approach has two key issues:

- It doesn't support three-node setups.

- It doesn't strictly follow quorum best practices.

I’d like to propose a refinement to this logic:

Instead of checking for a hard coded number of active MON services, we should ensure that at least half of the MON services—excluding the target node—remain available after enabling maintenance mode.

Examples:

- For a 3-node cluster with a MON on each node, excluding the target node, at least 2 MONs should remain active.

- For a 5-node cluster with a MON on each node, excluding the target node, at least 3 MONs should remain active.

samuelallan72:
> we should ensure that at least half of the MON services—excluding the target node—remain available after enabling maintenance mode.

I think we need to ensure that *more than* half of the mon services remain available. It is possible (although not recommended) for an even number of nodes - in this case I believe half is not enough. Eg. for a 4 node cluster, 3 nodes need to remain available.

PR #557, opened 2025-05-14
Ensure a majority of mon services remain available before entering maintenance mode

samuelallan72:
# Description

Previously, the number of mon services was hardcoded to 3 - this means this preflight check only correctly supported clusters of 4-5 mon services. See #534 for more information.

This change dynamically calculates the required number of remaining ceph-mon services to ensure a majority will be available after entering maintenance mode.

Fixes: #534

## Type of change

## How has this been tested?

Locally, on a 3 node microceph deployment on LXD VMs.

Before:

```
root@microceph-0:~# microceph status
MicroCeph deployment summary:
- microceph-0 (10.180.54.6)
  Services: mds, mgr, mon, osd
  Disks: 3
- microceph-1 (10.180.54.82)
  Services: mds, mgr, mon, osd
  Disks: 3
- microceph-2 (10.180.54.165)
  Services: mds, mgr, mon, osd
  Disks: 3

root@microceph-0:~# microceph cluster maintenance enter microceph-0
Error: failed to enter maintenance mode: error bringing node 'microceph-0' into maintenance: maintenance operations failed: [(need at least 3 mon, 1 mds, and 1 mgr services in the cluster besides those in node 'microceph-0')]
```

After:

```
root@microceph-0:~# microceph cluster maintenance enter microceph-0
Check if osds.[1 2 3] in node 'microceph-0' are ok-to-stop. (succeeded)
Check if there are at least a majority of mon services, 1 mds service, and 1 mgr service in the cluster besides those in node 'microceph-0' (succeeded)
Run `ceph osd set noout`. (succeeded)
Assert osd has 'noout' flag set. (succeeded)
```

## Contributor checklist

Please check that you have:

- [ ] added tests to verify effectiveness of this change [I'm not very familiar with this codebase, and I couldn't find existing tests for maintenance mode]

### DIFF
--- a/microceph/ceph/operations.go
+++ b/microceph/ceph/operations.go
@@ -110,10 +110,6 @@ func (o *CheckOsdOkToStopOps) GetName() string {
 // CheckNonOsdSvcEnoughOps is an operation to check if non-osd service in a node are enough.
 type CheckNonOsdSvcEnoughOps struct {
 	ClusterOps
-
-	MinMon int
-	MinMds int
-	MinMgr int
 }
 
 // Run checks if non-osds service in a node are enough.
@@ -128,17 +124,29 @@ func (o *CheckNonOsdSvcEnoughOps) Run(name string) error {
 		"mgr": 0,
 		"mds": 0,
 	}
+	totals := map[string]int{
+		"mon": 0,
+		"mgr": 0,
+		"mds": 0,
+	}
 	for _, service := range services {
 		// do not count the service on this node
 		if service.Location != name {
 			remains[service.Service]++
 		}
+		totals[service.Service]++
 	}
+
+	// a majority of ceph-mon services must remain active to retain quorum
+	minMon := totals["mon"] / 2 + 1
+	// only need one ceph-mds and one ceph-mgr: they operate as one active, the rest in standby
+	minMds := 1
+	minMgr := 1
 
 	// the remaining services must be sufficient to make the cluster healthy after the node enters
 	// maintanence mode.
-	if remains["mon"] < o.MinMon || remains["mds"] < o.MinMds || remains["mgr"] < o.MinMgr {
-		return fmt.Errorf("need at least %d mon, %d mds, and %d mgr services in the cluster besides those in node '%s'", o.MinMon, o.MinMds, o.MinMgr, name)
+	if remains["mon"] < minMon || remains["mds"] < minMds || remains["mgr"] < minMgr {
+		return fmt.Errorf("need at least %d mon, %d mds, and %d mgr services in the cluster besides those in node '%s'", minMon, minMds, minMgr, name)
 	}
 	logger.Infof("remaining mon (%d), mds (%d), and mgr (%d) services in the cluster are enough after '%s' enters maintenance mode", remains["mon"], remains["mds"], remains["mgr"], name)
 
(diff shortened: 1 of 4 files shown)


samuelallan72:
Warning: currently, getting the list of services does not take into account whether those services are already down or in maintenance mode. I don't know if this is expected due to the feature being in progress, or if we need to add more logic to ListServices or CheckNonOsdSvcEnoughOps.Run to specifically check for active services.

microceph/ceph/operations.go
```diff
@@ -128,17 +124,29 @@ func (o *CheckNonOsdSvcEnoughOps) Run(name string) error {
...
+		"mon": 0,
+		"mgr": 0,
+		"mds": 0,
+	}
 	for _, service := range services {
 		// do not count the service on this node
 		if service.Location != name {
 			remains[service.Service]++
 		}
+		totals[service.Service]++
```
UtkarshBhatthere, line 137: ok, this line is a little confusing to me. On L134-L136 we conditionally increment the service count, and on L137 we do it unconditionally ? Can you please a code comment which explains this behavior ?

samuelallan72: Sure, I added comments and refactored the code too for clarity. :)

samuelallan72: (force pushed to add the sign off trailer to the new commits)

UtkarshBhatthere: much better, thanks :D

microceph/ceph/operations.go
```diff
@@ -128,17 +124,29 @@ func (o *CheckNonOsdSvcEnoughOps) Run(name string) error {
...
+	minMon := total_mon_services / 2 + 1
+	// only need one ceph-mds and one ceph-mgr: they operate as one active, the rest in standby
+	minMds := 1
+	minMgr := 1
+
 	// the remaining services must be sufficient to make the cluster healthy after the node enters
 	// maintanence mode.
-	if remains["mon"] < o.MinMon || remains["mds"] < o.MinMds || remains["mgr"] < o.MinMgr {
-		return fmt.Errorf("need at least %d mon, %d mds, and %d mgr services in the cluster besides those in node '%s'", o.MinMon, o.MinMds, o.MinMgr, name)
+	if remains["mon"] < minMon || remains["mds"] < minMds || remains["mgr"] < minMgr {
```
chanchiwai-ray, line 148: This change can be tested in `operations_test.go` (I did not know why I did not add unit test for this function..., may be I over looked)

microceph/ceph/operations.go
```diff
@@ -128,17 +124,29 @@ func (o *CheckNonOsdSvcEnoughOps) Run(name string) error {
...
 		}
+
+		// count the total number of mon services, so we can calculate the minimum number of mon services required for quorum
+		if service.Service == "mon" {
+			total_mon_services += 1
+		}
 	}
 
+	// a majority of ceph-mon services must remain active to retain quorum
+	minMon := total_mon_services / 2 + 1
```
chanchiwai-ray, line 141: May be this calculation should be lower bounded to 3? 

As recommended by official [docs](https://docs.ceph.com/en/mimic/rados/operations/add-or-rm-mons/#adding-monitors):

> We recommend at least 3 monitors for a production cluster

microceph/ceph/operations.go
chanchiwai-ray, line 148: (non-blocking, for microceph team) this function is mostly a subset of this function here https://github.com/canonical/microceph/blob/main/microceph/ceph/remove.go#L84-L106, do you think we need to refactor this later in another PR?

microceph/ceph/operations.go
samuelallan72: This is about the minimum required to retain quorum, so the minimum theoretically is 1 (for a single node deployment). For a 2 or 3 node cluster, the minimum for quorum is 2. :)

chanchiwai-ray: I should read the comment carefully 👀

microceph/ceph/operations.go
UtkarshBhatthere: Ideally, the refactor should be a part of this PR as otherwise we would be creating technical debt (and possibly untested code paths) for the project.

samuelallan72: Can we refactor in a separate PR? Refactoring really should be separate from this, otherwise the scope of the PR will keep growing.

Issue #558, opened 2025-05-14
Maintenance mode enter pre-flight check does not account for other nodes in maintenance mode


samuelallan72:
The checks for `microceph cluster maintenance enter` do not check if other nodes are already in maintenance mode. This means you can currently successfully run:

```
microceph cluster maintenance enter node-1
microceph cluster maintenance enter node-2
... all the nodes
```

There are checks already to try to ensure enough nodes remain active to retain quorum, but they don't account for other nodes being in maintenance mode.

This can be misleading, as in practice, it's risky to have a majority of nodes in maintenance mode.

UtkarshBhatthere:
Good catch @samuelallan72 for a future traveller tools like `ceph report` could be used to get a unified view of active ceph services. @samuelallan72 wdyt ?

samuelallan72:
@UtkarshBhatthere I could be wrong, but doesn't the new maintenance mode by default simply mark it as in maintenance, but not stop any services? So `ceph report` would still show the nodes as active. Perhaps we need something specific to microceph keep the maintenance mode in/out state?

chanchiwai-ray:
The above case is impossible for microceph currently see [test case](https://github.com/canonical/microceph/blob/main/.github/workflows/tests.yml#L512) **currently** since the number of mon correlates to the number of node (up to maximum of 3 mons). Validate the NonOsdSvcs is the same as validating the number of node.

In additional, 

> Maintenance mode enter pre-flight check does not account for other nodes in maintenance mode

is a design decision of being idempotent (but we can improve that).
 
What we could do or thought of doing is adding GET method to `/ops/maintenance/{node}` API endpoint to retrieve (or re-validate) the maintenance status of a particular node.

PR #565, opened 2025-06-04
Ensure a majority of mon services remain available before entering maintenance mode

chanchiwai-ray:
# Description

Supersede #557 (the author of that PR is on a long break) 

Fixes: https://github.com/canonical/microceph/issues/534

## Type of change

- Clean code (code refactor, test updates)
- Documentation update (change to documentation

## How has this been tested?

Locally, on a 3 node microceph deployment on LXD VMs.

Before
```
root@microceph-0:~# microceph status
MicroCeph deployment summary:
- microceph-0 (10.180.54.6)
  Services: mds, mgr, mon, osd
  Disks: 3
- microceph-1 (10.180.54.82)
  Services: mds, mgr, mon, osd
  Disks: 3
- microceph-2 (10.180.54.165)
  Services: mds, mgr, mon, osd
  Disks: 3

root@microceph-0:~# microceph cluster maintenance enter microceph-0
Error: failed to enter maintenance mode: error bringing node 'microceph-0' into maintenance: maintenance operations failed: [(need at least 3 mon, 1 mds, and 1 mgr services in the cluster besides those in node 'microceph-0')]
```

After
```
root@microceph-0:~# microceph cluster maintenance enter microceph-0
Check if osds.[1 2 3] in node 'microceph-0' are ok-to-stop. (succeeded)
Check if there are at least a majority of mon services, 1 mds service, and 1 mgr service in the cluster besides those in node 'microceph-0' (succeeded)
Run `ceph osd set noout`. (succeeded)
Assert osd has 'noout' flag set. (succeeded)
```
## Contributor checklist

Please check that you have:

### DIFF
--- a/docs/explanation/cluster-maintenance.rst
+++ b/docs/explanation/cluster-maintenance.rst
@@ -28,7 +28,7 @@ Enabling maintenance mode
 
 - Check if OSDs on the node are ``ok-to-stop`` to ensure sufficient redundancy to tolerate the loss
   of OSDs on the node.
-- Check if the number of running services is greater than the minimum (3 MON, 1 MDS, 1 MGR)
+- Check if the number of running services is greater than the minimum (majority of MON, 1 MDS, 1 MGR)
   required for quorum.
 - *(Optional)* Apply noout flag to prevent data migration from triggering during the planned
   maintenance slot. (default=True)
--- a/microceph/ceph/monitor.go
+++ b/microceph/ceph/monitor.go
@@ -6,6 +6,8 @@ import (
 	"github.com/canonical/microceph/microceph/common"
 	"os"
 	"path/filepath"
+
+	"github.com/tidwall/gjson"
 )
 
 func genMonmap(path string, fsid string) error {
@@ -86,4 +88,23 @@ func removeMon(hostname string) error {
 		return fmt.Errorf("failed to remove monitor %q: %w", hostname, err)
 	}
 	return nil
+}
+
+func getActiveMons() ([]string, error) {
+	output, err := common.ProcessExec.RunCommand("ceph", "-s", "-f", "json")
+	if err != nil {
+		logger.Errorf("Failed fetching ceph status: %v", err)
+		return nil, err
+	}
+
+	logger.Debugf("Ceph Status:\n%s", output)
+
+	// Get the active mons services.
+	activeMons := []string{}
+	result := gjson.Get(output, "quorum_names")
+	for _, name := range result.Array() {
+		activeMons = append(activeMons, name.String())
+	}
+
+	return activeMons, nil
 }
--- a/microceph/ceph/operations.go
+++ b/microceph/ceph/operations.go
@@ -7,6 +7,8 @@ import (
 	"github.com/canonical/lxd/shared/logger"
 
 	"github.com/canonical/microcluster/v2/state"
+
+	"github.com/canonical/microceph/microceph/api/types"
 )
 
 // RunOperations runs the provided operations or return the action plan.
@@ -111,10 +113,6 @@ func (o *CheckOsdOkToStopOps) GetName() string {
 // CheckNonOsdSvcEnoughOps is an operation to check if non-osd service in a node are enough.
 type CheckNonOsdSvcEnoughOps struct {
 	ClusterOps
-
-	MinMon int
-	MinMds int
-	MinMgr int
 }
 
 // Run checks if non-osds service in a node are enough.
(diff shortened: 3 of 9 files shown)


microceph/ceph/remove.go
```diff
@@ -86,23 +111,11 @@ func checkPrerequisites(cli *microCli.Client, name string) error {
...
-			continue
-		}
-		foundMap[service.Service]++
-	}
-	logger.Debugf("Services: %v, foundMap: %v", services, foundMap)
-	if foundMap["mon"] < 3 || foundMap["mgr"] < 1 || foundMap["mds"] < 1 {
-		return fmt.Errorf("Need at least 3 mon, 1 mds, and 1 mgr besides %v", name)
+
+	// check if the remaining non OSD services is enough to maintain a healthy cluster
+	err = EnsureNonOsdSvcEnough(services, name, 3, 1, 1)
```
chanchiwai-ray, line 116: To avoid introducing breaking change, I keep the same number as before.

microceph/ceph/operations.go
```diff
@@ -123,31 +119,28 @@ func (o *CheckNonOsdSvcEnoughOps) Run(name string) error {
...
 
-	// the remaining services must be sufficient to make the cluster healthy after the node enters
-	// maintanence mode.
-	if remains["mon"] < o.MinMon || remains["mds"] < o.MinMds || remains["mgr"] < o.MinMgr {
-		return fmt.Errorf("need at least %d mon, %d mds, and %d mgr services in the cluster besides those in node '%s'", o.MinMon, o.MinMds, o.MinMgr, name)
+	minMds := 1               // only need at least one: they operate as one active, the rest in standby
+	minMgr := 1               // only need at least one: they operate as one active, the rest in standby
+	minMon := totalMons/2 + 1 // a majority of ceph mon services must remain active to retain quorum
+
+	err = EnsureNonOsdSvcEnough(services, name, minMon, minMgr, minMds)
```
chanchiwai-ray: This is not the perfect way (or even wrong) to calculate quorum, but we can't easily do that here without some changes in API (`/services` endpoint). Basically, we need to know the total number of **active** mon services, not the number of mon services recorded in the database (as it can be in inactive state via `snap stop microceph.mon`, or for some reason the snap service dies). Excluding these two reasons, the calculation should still be fine because if there a record in the database, the mon service should be active and enabled by default.

This issue was mentioned in https://github.com/canonical/microceph/issues/558,  I am not sure how big the impact is if we want to get the status of the snap services for each node (via API). Maybe @UtkarshBhatthere and @sabaini can provide some insight on this?

chanchiwai-ray: I used `microceph.ceph` to get the info, not sure if that's a good practice in microceph

UtkarshBhatthere: using microceph.ceph is good, there are some limitations (cannot access files located at arbitrary paths) as it is under snap confinement.

microceph/ceph/services.go
```diff
@@ -138,6 +138,64 @@ func getMons() (common.Set, error) {
 	return retval, nil
 }
 
+func getActiveMons() ([]string, error) {
```
UtkarshBhatthere: The file services.go was originally written to host methods for all services however, we added files for individual services later (monitor.go for mon, metadata.go for mds, and manager.go for mgr). It would be nicer to have these methods in the respective files.

microceph/ceph/services.go
```diff
@@ -138,6 +138,64 @@ func getMons() (common.Set, error) {
...
+	for _, name := range result.Array() {
+		activeMgrs = append(activeMgrs, name.String())
+	}
+	activeMgrs = append(activeMgrs, gjson.Get(output, "active_name").String())
+
+	return activeMgrs, nil
+}
+
+func getActiveMdss() ([]string, error) {
+	output, err := processExec.RunCommand("ceph", "fs", "status", "-f", "json")
```
UtkarshBhatthere: Similar to the `ceph mon dump` command there are daemon dumps available for mds and mgr as well. Just wondering if they were not sufficient ?

chanchiwai-ray: The  `ceph * dump` command(s) does not always seem to provide the information we need: we need the name of the node + their corresponding status (active or inactive). The `ceph mgr dump` provides it, while this one does not

microceph/ceph/services.go
chanchiwai-ray: oops, I missed this comment. Updated now

chanchiwai-ray:
@UtkarshBhatthere any chance you could take a look again, and maybe re-triggering timeout tests? Thanks a lot!

docs/explanation/cluster-maintenance.rst
```diff
@@ -28,7 +28,7 @@ Enabling maintenance mode
 
 - Check if OSDs on the node are ``ok-to-stop`` to ensure sufficient redundancy to tolerate the loss
   of OSDs on the node.
-- Check if the number of running services is greater than the minimum (3 MON, 1 MDS, 1 MGR)
+- Check if the number of running services is greater than the minimum (majority of MON, 1 MDS, 1 MGR)
```
UtkarshBhatthere, line 31: Is this criteria modification done so that a minimal functional ceph cluster is available when a particular node is going into maintenance ?

UtkarshBhatthere: Asking because 1MDS daemon may not be able  to sustain FS IO requests if more than one FS Volume are being consumed (one mds per fs volume)

microceph/ceph/monitor.go
```diff
@@ -86,3 +88,22 @@ func removeMon(hostname string) error {
 	}
 	return nil
 }
+
+func getActiveMons() ([]string, error) {
+	output, err := processExec.RunCommand("ceph", "-s", "-f", "json")
```
UtkarshBhatthere: Have you experimented with `ceph mon dump` ? 
Not sure if it works for the maintenance use-case but there is an existing method in `services.go` called `getMons` which returns a set of  mons.

microceph/ceph/operations.go
```diff
@@ -123,31 +121,53 @@ func (o *CheckNonOsdSvcEnoughOps) Run(name string) error {
...
 		}
 	}
 
-	// the remaining services must be sufficient to make the cluster healthy after the node enters
-	// maintanence mode.
-	if remains["mon"] < o.MinMon || remains["mds"] < o.MinMds || remains["mgr"] < o.MinMgr {
-		return fmt.Errorf("need at least %d mon, %d mds, and %d mgr services in the cluster besides those in node '%s'", o.MinMon, o.MinMds, o.MinMgr, name)
+	minMds := 1               // only need at least one: they operate as one active, the rest in standby
+	minMgr := 1               // only need at least one: they operate as one active, the rest in standby
+	minMon := totalMons/2 + 1 // a majority of ceph mon services must remain active to retain quorum
```
UtkarshBhatthere, line 134: For a two node cluster (active mons == 2) minMon comes up to be two. similarly for single node clusters active mons == minMons == 1.  Does it make sense to add a minimum threshold like `min mons should atleast be 2 for going into maintenance mode` ?

docs/explanation/cluster-maintenance.rst
chanchiwai-ray: No, this modification only focuses on the mon services. But I agree that a proper check can be implemented for MDS / MGR too, would bug reports sufficient for now?

microceph/ceph/monitor.go
chanchiwai-ray: I remember the `ceph mon dump` gives you the mons (active or inactive), it does not tell you if it's part of quorum or not. `ceph -s` -> `quorum_names` gives you only the mons that are in quorum, which is what we need. 

I will double check and come back to you

microceph/ceph/operations.go
chanchiwai-ray: For a two node cluster: total active mons == 2, minMon == 2, active mons after maintenance mode will be 2-1=1. So maintenance mode will fail because active mons after maintenance mode (1) < minMon (2).

For a single node cluster: total active mons == 1, minMon == 1, active mons after maintenance mode will be 1-1=0.  So maintenance mode will fail because active mons after maintenance mode (0) < minMon (1).

The `EnsureNonOsdSvcEnough` determines if the services are enough after the `node` is in maintenance mode. So we don't need minimum threshold

microceph/ceph/monitor.go
chanchiwai-ray: Yes, indeed, if I `snap stop microceph.mon`, that mon still shows up in `ceph mon dump`, but it will not show up in `ceph -s`

microceph/ceph/services.go
UtkarshBhatthere: ack

docs/explanation/cluster-maintenance.rst
UtkarshBhatthere: yeah, a bug will help keep track of status. Something like `ensure FS (mds) availability during maintenance`.

microceph/ceph/monitor.go
UtkarshBhatthere: ack, thanks for the clarification.

microceph/ceph/operations.go
UtkarshBhatthere: ack, nice.

PR #567, opened 2025-06-09
Enforce v2 for monitors

lmlg:
In order to avoid ambiguities, this PR strictly enforces version 2 for all monitors in the Ceph cluster.


microceph/ceph/bootstrap.go
```diff
@@ -49,6 +49,9 @@ func Bootstrap(ctx context.Context, s interfaces.StateInterface, data common.Boo
 		monIp = fmt.Sprintf("[%s]", monIp)
 	}
 
+	// Support both v1 and v2 for mons.
+	monIp = "any:" + monIp
```
sabaini: Hm, must admit I'm a bit confused here -- do we not want to enforce v2 only, while `any` would allow both v1 and v2?

sabaini (changes requested):
Hey @lmlg could you maybe clarify the config change here. Aiui we do want to only support v2 here; however in the code we allow `any` ie. both v1 and v2?
https://docs.ceph.com/en/latest/rados/configuration/msgr2/#address-formats

.github/workflows/tests.yml
```diff
@@ -895,3 +895,49 @@ jobs:
 
     - name: Verify Remote removal
       run: ~/actionutils.sh remote_remove_and_verify
+
+  messenger-v2-tests:
+    name: Messenger v2 tests
+    runs-on: ubuntu-22.04
```
sabaini, line 1043: Coud we use 24.04?

.github/workflows/tests.yml
```diff
@@ -895,3 +895,49 @@ jobs:
...
+    - name: Bootstrap
+      run: ~/actionutils.sh bootstrap_head public --v2-only
+
+    - name: Verify messenger v2 on head node
+      run: ~/actionutils.sh test_v2_single_node
+
+    - name: Setup cluster
+      run: ~/actionutils.sh cluster_nodes public
+
+    - name: Veriy messenger v2 on all nodes
```
sabaini: typo: `Verify`

tests/scripts/actionutils.sh
```diff
@@ -1117,6 +1118,24 @@ function test_maintenance_enter_set_noout_stop_osds_and_exit_force() {
...
+
+function test_v2_all_nodes() {
+    for i in 0 1 2 3 ; do
+        lxc exec -- sh -c "cat /var/snap/microceph/current/conf/ceph.conf | grep -q v1:"
+        if [ "$?" -eq 0 ]; then
+            echo "messenger V1 is active on node $i"
+            exit 1
+        fi
+    done
+}
```
sabaini, line 1242: These functests check the conf file -- for added safety could this check be made for the actual ports the ceph-mon listens on?

sabaini (changes requested):
Heya @lmlg left some comments inline.
One thing I'm wondering -- the config gets rewritten on regular intervals in [UpdateConfig](https://github.com/canonical/microceph/blob/4910cfbd937860690fba764bd9c85ace5707f969/microceph/ceph/config.go#L331), I think we might need to cater for the rewrite that's happening there too?

.github/workflows/tests.yml
lmlg: I don't think of any reason not to. I only used 22.04 because every other test was using it.

tests/scripts/actionutils.sh
lmlg: This test checks the conf file and the mon dump as well, which will show if there's a V1 monitor. I can check for the port there as well.

.github/workflows/tests.yml
sabaini: Ack, suggest to roll to 24.04 over time

sabaini:
> > Heya @lmlg left some comments inline. One thing I'm wondering -- the config gets rewritten on regular intervals in [UpdateConfig](https://github.com/canonical/microceph/blob/4910cfbd937860690fba764bd9c85ace5707f969/microceph/ceph/config.go#L331), I think we might need to cater for the rewrite that's happening there too?
> 
> It's my understanding that the config rewrite via `UpdateConfig` does it via the addresses saved in the DB. For that case, the DB entries will already have the V2 prefix. I'll leave a cluster up and running for a few hours to see how it goes.

Ah, you're right! -- I missed the fact you're saving the prefix too.

This does open an issue with ipv6 support, as this is checking for a colon in the address, cf. [here](https://github.com/canonical/microceph/blob/4910cfbd937860690fba764bd9c85ace5707f969/microceph/ceph/config.go#L471), guess that would need to be adapted

sabaini (changes requested):
Thanks @lmlg , the ipv6 stuff lgtm :+1:  -- only things left would be to have test runner on ubuntu-24.04 and also check that the ceph-mon daemon listens on 3300

E.g. `sudo ss -Htnpl 'sport :3300' | grep -c ceph-mon` or similar

sabaini (approved):
Excellent, thanks @lmlg

PR #572, opened 2025-06-18
Migrate docs to the extension-based starter pack

skoech:
# Description

We need to migrate our MicroCeph docs from the [legacy Sphinx Documentation Starter Pack ](https://github.com/canonical/sphinx-docs-starter-pack/tree/pre-extension)(pre-extension version) to the latest version that adopts the `canonical-sphinx` Sphinx extension to ease documentation maintenance.

The [extension-based documentation starter pack](https://github.com/canonical/canonical-sphinx) provides a set of features and configurations that are common across Canonical documentation projects. Key components, such as configuration and styling, are loaded as an add-on to your documentation project.

The new [documentation starter pack](https://canonical-starter-pack.readthedocs-hosted.com/latest/) includes the `canonical-sphinx` extension which is a group of Sphinx extensions recommended by Canonical that:

- provides a design override of the Furo theme for the creation of Canonical branded documentation
- provides a minimal installation by default, and also offers a `[full]` optional installation which provides significant additional functionality for Sphinx based documentation
- is installed and added to extensions within the configuration file of a Sphinx deployment

## Type of change

## How has this been tested?
I've tested it locally by running these commands:
`make run`
`make spelling`
`make linkcheck`

## Contributor checklist


.github/workflows/markdown-style-checks.yml
```diff
@@ -0,0 +1,24 @@
...
+    - 'docs/**'   # Only run on changes to the docs directory
+
+jobs:
+  markdown-lint:
+    runs-on: ubuntu-22.04
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+    - uses: DavidAnson/markdownlint-cli2-action@v16
```
sabaini: Please pin external dependencies to a specific git commit (instead of a tag) to avoid supply-chain attacks

docs/Makefile
```diff
@@ -1,130 +1,196 @@
 # Minimal makefile for Sphinx documentation
 #
+# Add your customisation to `Makefile` instead.
```
sabaini: Hm, that comment seems out of place given that this is the Makefile?

docs/conf.py
```diff
@@ -1,140 +1,348 @@
...
-]
-extensions.extend(custom_extensions)
+#######################
+# Project information #
+#######################
 
-### Configuration for extensions
+# Project name
+#
+# TODO: Update with the official name of your project or product
```
sabaini: Guess this `TODO: ` stanzas could be deleted given that we already did actually do the customisation, and just leave the `Update with the official name of your project or product` part as an explanation

docs/conf.py
```diff
@@ -1,140 +1,348 @@
...
+
+
+# Copyright string; shown at the bottom of the page
+#
+# Now, the starter pack uses CC-BY-SA as the license
+# and the current year as the copyright year.
+#
+# TODO: If your docs need another license, specify it instead of 'CC-BY-SA'.
+#
+# TODO: If your documentation is a part of the code repository of your project,
```
sabaini: These TODOs as well

docs/conf.py
```diff
@@ -1,140 +1,348 @@
...
+#
+# TODO: Update with the official URL of your docs or leave empty if unsure.
+#
+# NOTE: The Open Graph Protocol (OGP) enhances page display in a social graph
+#       and is used by social media platforms; see https://ogp.me/
 
-# Used for related links
-if not 'discourse_prefix' in html_context and 'discourse' in html_context:
-    html_context['discourse_prefix'] = html_context['discourse'] + '/t/'
+ogp_site_url = "https://canonical-starter-pack.readthedocs-hosted.com/"
```
sabaini: This should point to the MicroCeph RTD website I guess?

docs/conf.py
```diff
@@ -1,140 +1,348 @@
...
+html_context = {
+    # Product page URL; can be different from product docs URL
+    #
+    # TODO: Change to your product website URL,
+    #       dropping the 'https://' prefix, e.g. 'ubuntu.com/lxd'.
+    #
+    # TODO: If there's no such website,
+    #       remove the {{ product_page }} link from the page header template
+    #       (usually .sphinx/_templates/header.html; also, see README.rst).
+    "product_page": "",
```
sabaini, line 104: Think we might want to point this to http://microceph.com ?

docs/conf.py
```diff
@@ -1,140 +1,348 @@
...
+    # Docs location in the repo; used in links for viewing the source files
+    #
+
+
+    # TODO: To customise the directory, uncomment and update as needed.
+    "repo_folder": "/docs/",
+    # TODO: To enable or disable the Previous / Next buttons at the bottom of pages
+    # Valid options: none, prev, next, both
+    # "sequential_nav": "both",
+    # TODO: To enable listing contributors on individual pages, set to True
```
sabaini: Delete these TODOs as well

docs/conf.py
```diff
@@ -1,140 +1,348 @@
...
+
+if os.path.exists('./reuse/substitutions.yaml'):
+    with open('./reuse/substitutions.yaml', 'r') as fd:
+        myst_substitutions = yaml.safe_load(fd.read())
+
+# Add configuration for intersphinx mapping
+
+intersphinx_mapping = {
+    'starter-pack': ('https://canonical-example-product-documentation.readthedocs-hosted.com/en/latest', None)
+}
```
sabaini: Hm, I wonder if we want to have that intersphinx mapping to an example project here?

docs/reuse/links.txt
sabaini, line 1: Are these links / substitutions below needed for the MicroCeph documentation -- I don't see them referenced anywhere? If not needed I'd suggest deleting them

sabaini (changes requested):
Thanks @skoech for this, left some commentary

docs/Makefile
skoech: Ah, good catch. Should be `conf.py`.

docs/conf.py
skoech: Done.

docs/conf.py
skoech: Yes. Done.

docs/conf.py
skoech: Done.

docs/conf.py
skoech: I'm not sure this qualifies as the product website.

docs/conf.py
skoech: Good catch.

.github/workflows/markdown-style-checks.yml
skoech: Huh, didn't think of that. How would I do this?

docs/reuse/links.txt
skoech: We are not referencing it yet, but I think it would be nice if we started doing so, so that in case a link changes in the docs, we only modify this file.

.github/workflows/markdown-style-checks.yml
sabaini: You'd check the git repo of the action and note the commit corresponding to the v16 tag -- here b4c9feab76d8025d1e83c653fa3990936df0e6c8 

Then reference it like this (opt. comment to note where the hash comes from)

```
    - uses: DavidAnson/markdownlint-cli2-action@b4c9feab76d8025d1e83c653fa3990936df0e6c8  # v16
```

For context, the motivation here is that an attacker that takes control of the above repo could easily point the v16 tag to malicious code, but having this commit hash pointed at another (malicious) code is near-impossible.

docs/conf.py
UtkarshBhatthere: the domain is owned by IS and under canonical. If we want to we could propose this to be used as the landing page. However, I think that should be a separate task.

skoech:
Wonderful!

PR #574, opened 2025-06-24
Adds NFS support

claudiubelu:
# Description

Adds support for enabling NFS clusters in microceph.

Adds the following commands:
- `microceph enable nfs --cluster-id <cluster-id> [--bind-address <ip-address>] [--port <port-num>] [--v4-min-version 0/1/2] [--target <server>] [--wait <bool>]`
- `microceph disable nfs --cluster-id <cluster-id> [--target <server>]`

Adds the `services/nfs` API endpoint (`PUT` / `DELETE`).

Adds the following tables:
- `service_groups`: used to track microceph service clusters.
- `grouped_services`: used to track clustered services running on a particular server.

Added dependencies in the snap:
- `nfs-ganesha`
- `nfs-ganesha-ceph`
- `nfs-ganesha-rados-grace`

## Type of change

## How has this been tested?

> [!NOTE]

## Contributor checklist

Please check that you have:

### DIFF
--- a/microceph/api/types/services.go
+++ b/microceph/api/types/services.go
@@ -1,6 +1,10 @@
 // Package types provides shared types and structs.
 package types
 
+import (
+	"regexp"
+)
+
 // Services holds a slice of services
 type Services []Service
 
@@ -26,6 +30,9 @@ type EnableService struct {
 type NFSService struct {
 	ClusterID string `json:"cluster_id" yaml:"cluster_id"`
 }
+
+// NFSClusterIDRegex is a regex for acceptable ClusterIDs.
+var NFSClusterIDRegex = regexp.MustCompile(`^[\w][\w.-]{1,61}[\w]$`)
 
 // RGWService holds a port number and enable/disable flag
 type RGWService struct {
--- a/microceph/ceph/nfs.go
+++ b/microceph/ceph/nfs.go
@@ -245,28 +245,33 @@ func createNFSKeyring(path, clusterID, userID string) error {
 
 // ensureNFSPools creates the NFS data and metadata Pools for Ganesha if they do not exist.
 func ensureNFSPools(clusterID string) error {
+	logger.Debugf("Creating '.nfs' rados pool if it doesn't exist.")
 	_, err := radosRun("ls", "--pool", ".nfs", "--all", "--create")
 	if err != nil && !strings.Contains(err.Error(), "File exists") {
 		return fmt.Errorf("failed to create .nfs pool: %w", err)
 	}
 
+	logger.Debugf("Creating '.nfs.metadata' rados pool if it doesn't exist.")
 	_, err = radosRun("ls", "--pool", ".nfs.metadata", "--all", "--create")
 	if err != nil && !strings.Contains(err.Error(), "File exists") {
 		return fmt.Errorf("failed to create .nfs.metadata pool: %w", err)
 	}
 
 	// the command is idempotent.
+	logger.Debugf("Enabling cephfs application on '.nfs' pool.")
 	_, err = osdEnablePoolApp(".nfs", "cephfs")
 	if err != nil {
 		return fmt.Errorf("failed to enable .nfs pool: %w", err)
 	}
 
 	object := fmt.Sprintf("conf-nfs.%s", clusterID)
+	logger.Debugf("Creating '%s' rados object in pool '.nfs' in namespace '%s'.", object, clusterID)
 	_, err = radosRun("create", "--pool", ".nfs", "-N", clusterID, object)
 	if err != nil && !strings.Contains(err.Error(), "File exists") {
 		return fmt.Errorf("failed to create object for Ganesha: %w", err)
 	}
 
+	logger.Debugf("Finished creating rados pools and objects for NFS with ClusterID %s.", clusterID)
 	return nil
 }
 
--- a/microceph/ceph/service_placement_nfs.go
+++ b/microceph/ceph/service_placement_nfs.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"syscall"
 
+	"github.com/canonical/microceph/microceph/api/types"
 	"github.com/canonical/microceph/microceph/database"
 	"github.com/canonical/microceph/microceph/interfaces"
 )
@@ -25,8 +26,8 @@ func (nfs *NFSServicePlacement) PopulateParams(s interfaces.StateInterface, payl
 		return err
 	}
 
-	if len(nfs.ClusterID) == 0 {
-		return fmt.Errorf("expected cluster_id to be non-empty")
+	if !types.NFSClusterIDRegex.MatchString(nfs.ClusterID) {
+		return fmt.Errorf("expected cluster_id to be valid (regex: '%s')", types.NFSClusterIDRegex.String())
 	}
 
 	if nfs.V4MinVersion > 2 {
--- a/microceph/ceph/service_placement_nfs_test.go
+++ b/microceph/ceph/service_placement_nfs_test.go
@@ -2,6 +2,8 @@ package ceph
 
 import (
 	"context"
+	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/canonical/microceph/microceph/api/types"
@@ -38,7 +40,22 @@ func (s *servicePlacementNFSSuite) TestInvalidPayload() {
 	}
 
 	err := ServicePlacementHandler(context.Background(), s.TestStateInterface, payload)
-	assert.ErrorContains(s.T(), err, "expected cluster_id to be non-empty")
+	assert.ErrorContains(s.T(), err, "expected cluster_id to be valid")
+
+	payload.Payload = "{\"cluster_id\":\".foo\"}"
+
+	err = ServicePlacementHandler(context.Background(), s.TestStateInterface, payload)
+	assert.ErrorContains(s.T(), err, "expected cluster_id to be valid")
+
+	payload.Payload = fmt.Sprintf("{\"cluster_id\":\"%s\"}", strings.Repeat("a", 64))
+
+	err = ServicePlacementHandler(context.Background(), s.TestStateInterface, payload)
+	assert.ErrorContains(s.T(), err, "expected cluster_id to be valid")
+
+
+	payload.Payload = "{\"cluster_id\":\"foo\",\"v4_min_version\":10}"
+
+	err = ServicePlacementHandler(context.Background(), s.TestStateInterface, payload)
 
 	payload.Payload = "{\"cluster_id\":\"foo\",\"v4_min_version\":10}"
 
--- a/microceph/cmd/microceph/enable_nfs.go
+++ b/microceph/cmd/microceph/enable_nfs.go
@@ -32,7 +32,7 @@ func (c *cmdEnableNFS) Command() *cobra.Command {
 		Short: "Enable the NFS Ganesha service on the --target server (default: this server)",
 		RunE:  c.Run,
 	}
-	cmd.PersistentFlags().StringVar(&c.flagClusterID, "cluster-id", "", "NFS Cluster ID")
+	cmd.PersistentFlags().StringVar(&c.flagClusterID, "cluster-id", "", fmt.Sprintf("NFS Cluster ID (must match regex: '%s'", types.NFSClusterIDRegex.String()))
 	cmd.PersistentFlags().StringVar(&c.flagBindAddr, "bind-address", "0.0.0.0", "Bind address to use for the NFS Ganesha service")
 	cmd.PersistentFlags().UintVar(&c.flagBindPort, "bind-port", 2049, "Bind port to use for the NFS Ganesha service")
 	cmd.PersistentFlags().UintVar(&c.flagV4MinVersion, "v4-min-version", 1, "Minimum supported version")
@@ -43,8 +43,8 @@ func (c *cmdEnableNFS) Command() *cobra.Command {
 
 // Run handles the enable nfs command.
 func (c *cmdEnableNFS) Run(cmd *cobra.Command, args []string) error {
-	if len(c.flagClusterID) == 0 {
-		return fmt.Errorf("please provide a cluster ID using the `--cluster-id` flag")
+	if !types.NFSClusterIDRegex.MatchString(c.flagClusterID) {
+		return fmt.Errorf("please provide a valid cluster ID using the `--cluster-id` flag (regex: '%s')", types.NFSClusterIDRegex.String())
 	}
 
 	if c.flagV4MinVersion > 2 {
--- a/microceph/api/services.go
+++ b/microceph/api/services.go
@@ -180,12 +180,12 @@ func cmdNFSDeleteService(s state.State, r *http.Request) response.Response {
 
 	err := json.NewDecoder(r.Body).Decode(&svc)
 	if err != nil {
-		logger.Errorf("Failed decoding disable service request: %v", err)
+		logger.Errorf("failed decoding disable service request: %v", err)
 		return response.InternalError(err)
 	}
 
-	if len(svc.ClusterID) == 0 {
-		err := fmt.Errorf("Expected cluster_id to not be empty.")
+	if !types.NFSClusterIDRegex.MatchString(svc.ClusterID) {
+		err := fmt.Errorf("expected cluster_id to be valid (regex: '%s')", types.NFSClusterIDRegex.String())
 		return response.SmartError(err)
 	}
 
--- a/microceph/cmd/microceph/disable_nfs.go
+++ b/microceph/cmd/microceph/disable_nfs.go
@@ -23,15 +23,15 @@ func (c *cmdDisableNFS) Command() *cobra.Command {
 		Short: "Disable the NFS Ganesha service on the --target server (default: this server)",
 		RunE:  c.Run,
 	}
-	cmd.PersistentFlags().StringVar(&c.flagClusterID, "cluster-id", "", "NFS Cluster ID")
+	cmd.PersistentFlags().StringVar(&c.flagClusterID, "cluster-id", "", fmt.Sprintf("NFS Cluster ID (must match regex: '%s'", types.NFSClusterIDRegex.String()))
 	cmd.PersistentFlags().StringVar(&c.flagTarget, "target", "", "Server hostname (default: this server)")
 	return cmd
 }
 
 // Run handles the disable nfs command.
 func (c *cmdDisableNFS) Run(cmd *cobra.Command, args []string) error {
-	if len(c.flagClusterID) == 0 {
-		return fmt.Errorf("please provide a cluster ID using the `--cluster-id` flag")
+	if !types.NFSClusterIDRegex.MatchString(c.flagClusterID) {
+		return fmt.Errorf("please provide a valid cluster ID using the `--cluster-id` flag (regex: '%s')", types.NFSClusterIDRegex.String())
 	}
 
 	m, err := microcluster.App(microcluster.Args{StateDir: c.common.FlagStateDir})
--- a/microceph/cmd/microceph/disable_nfs_test.go
+++ b/microceph/cmd/microceph/disable_nfs_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/canonical/microceph/microceph/tests"
@@ -22,8 +23,16 @@ func (s *disableNFSSuite) SetupTest() {
 	s.BaseSuite.SetupTest()
 }
 
-func (s *disableNFSSuite) TestCmdDisableNFSEmpty() {
+func (s *disableNFSSuite) TestCmdDisableNFSInvalid() {
 	cmd := cmdDisableNFS{}
 	err := cmd.Run(nil, []string{})
-	assert.ErrorContains(s.T(), err, "please provide a cluster ID using the `--cluster-id` flag")
+	assert.ErrorContains(s.T(), err, "please provide a valid cluster ID using the `--cluster-id` flag")
+
+	cmd.flagClusterID = ".foo"
+	err = cmd.Run(nil, []string{})
+	assert.ErrorContains(s.T(), err, "please provide a valid cluster ID using the `--cluster-id` flag")
+
+	cmd.flagClusterID = strings.Repeat("a", 64)
+	err = cmd.Run(nil, []string{})
+	assert.ErrorContains(s.T(), err, "please provide a valid cluster ID using the `--cluster-id` flag")
 }
--- a/microceph/cmd/microceph/enable_nfs_test.go
+++ b/microceph/cmd/microceph/enable_nfs_test.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"strings"
 	"testing"
 
 	"github.com/canonical/microceph/microceph/tests"
@@ -22,10 +23,18 @@ func (s *enableNFSSuite) SetupTest() {
 	s.BaseSuite.SetupTest()
 }
 
-func (s *enableNFSSuite) TestCmdEnableNFSEmpty() {
+func (s *enableNFSSuite) TestCmdEnableNFSInvalidClusterID() {
 	cmd := cmdEnableNFS{}
 	err := cmd.Run(nil, []string{})
-	assert.ErrorContains(s.T(), err, "please provide a cluster ID using the `--cluster-id` flag")
+	assert.ErrorContains(s.T(), err, "please provide a valid cluster ID using the `--cluster-id` flag")
+
+	cmd.flagClusterID = ".foo"
+	err = cmd.Run(nil, []string{})
+	assert.ErrorContains(s.T(), err, "please provide a valid cluster ID using the `--cluster-id` flag")
+
+	cmd.flagClusterID = strings.Repeat("a", 64)
+	err = cmd.Run(nil, []string{})
+	assert.ErrorContains(s.T(), err, "please provide a valid cluster ID using the `--cluster-id` flag")
 }
 
 func (s *enableNFSSuite) TestCmdEnableNFSInvalidV4() {
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -378,6 +378,7 @@ parts:
     prime:
       - bin/*
       - lib/*/ganesha
+      - lib/*/libevent_core-*.so*
       - lib/*/libganesha_*.so*
       - lib/*/libnfsidmap.so*
       - lib/*/libntirpc.so*


microceph/cmd/microceph/enable_nfs.go
```diff
@@ -0,0 +1,71 @@
...
+func (c *cmdEnableNFS) Command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "nfs --cluster-id <cluster-id> [--v4-min-version 0/1/2] [--target <server>] [--wait <bool>]",
+		Short: "Enable the NFS Ganesha service on the --target server (default: this server)",
+		RunE:  c.Run,
+	}
+	cmd.PersistentFlags().StringVar(&c.flagClusterID, "cluster-id", "", "NFS Cluster ID")
+	cmd.PersistentFlags().UintVar(&c.flagV4MinVersion, "v4-min-version", 1, "Minimum supported version")
+	cmd.PersistentFlags().StringVar(&c.flagTarget, "target", "", "Server hostname (default: this server)")
+	cmd.Flags().BoolVar(&c.wait, "wait", true, "Wait for rgw service to be up.")
```
UtkarshBhatthere: typo s/rgw/nfs

microceph/ceph/service_placement_nfs.go
```diff
@@ -0,0 +1,43 @@
+package ceph
+
+import (
+	"context"
+	"encoding/json"
+
+	"github.com/canonical/microceph/microceph/interfaces"
+)
+
+type NFSServicePlacement struct {
```
UtkarshBhatthere, line 16: Service IP is missing that should be used as a `BindAddr` value.

microceph/ceph/service_placement_nfs.go
```diff
@@ -0,0 +1,43 @@
...
+
+	return EnableNFS(s, nfs.ClusterID, nfs.V4MinVersion, monitors)
+}
+
+func (nfs *NFSServicePlacement) PostPlacementCheck(s interfaces.StateInterface) error {
+	return genericPostPlacementCheck("nfs-ganesha")
+}
+
+func (nfs *NFSServicePlacement) DbUpdate(ctx context.Context, s interfaces.StateInterface) error {
+	return genericDbUpdate(ctx, s, "nfs-ganesha")
```
UtkarshBhatthere: I don't think a generic db update here would work. As the generic update only creates an entry in the services table but `nfs` is implemented as a clustered service which will have to have a new table (as per the spec: section schema updates).

microceph/ceph/run.go
```diff
@@ -3,3 +3,23 @@ package ceph
 func cephRun(args ...string) (string, error) {
 	return processExec.RunCommand("ceph", args...)
 }
+
```
UtkarshBhatthere, line 6: I am afraid this file is not the correct place for these helpers (except `radosRun` to go.

microceph/ceph/run.go
```diff
@@ -3,3 +3,23 @@ package ceph
...
+func osdEnablePool(pool, app string) (string, error) {
```
UtkarshBhatthere: s/osdEnablePool/osdEnablePoolApp

microceph/ceph/run.go
```diff
@@ -3,3 +3,23 @@ package ceph
...
+	return cephRun("osd", "pool", "application", "enable", pool, app)
+}
+
+func radosRun(args ...string) (string, error) {
+	return processExec.RunCommand("rados", args...)
+}
+
+func radosCreatePool(pool string) (string, error) {
```
UtkarshBhatthere: Please add a comment string for method (// radosCreatePool used to create dot pools i.e. pools that start with the "." prefix.)

microceph/ceph/service_placement_nfs.go
```diff
@@ -0,0 +1,43 @@
...
+func (nfs *NFSServicePlacement) PopulateParams(s interfaces.StateInterface, payload string) error {
+	err := json.Unmarshal([]byte(payload), &nfs)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func (nfs *NFSServicePlacement) HospitalityCheck(s interfaces.StateInterface) error {
```
UtkarshBhatthere, line 55: A hospitality check here may also verify that the required port (default 2049) is available. As the host may have some service running outside microceph that consumes it. This would allow us to fail gracefully.

UtkarshBhatthere: or if the required bind address is not available on the host.

microceph/ceph/run.go
```diff
@@ -3,3 +3,23 @@ package ceph
...
+	return radosRun("ls", "--pool", pool, "--all", "--create")
+}
+
+func radosCreateObject(pool, namespace, object string) (string, error) {
+	return radosRun("create", "-p", pool, "-N", namespace, object)
+}
+
+func ganeshaRadosGrace(cephConfPath, pool, namespace, userID, node string) (string, error) {
```
UtkarshBhatthere: this should live in the nfs.go file ?

UtkarshBhatthere: the method name also does not express what it does.

UtkarshBhatthere: maybe you want ganeshaRadosGraceRun ?

microceph/ceph/run.go
claudiubelu: Done

microceph/ceph/run.go
claudiubelu: Done

microceph/ceph/run.go
claudiubelu: No longer needed.

microceph/ceph/run.go
claudiubelu: Done

microceph/cmd/microceph/enable_nfs.go
claudiubelu: Done

microceph/ceph/service_placement_nfs.go
claudiubelu: Done

microceph/ceph/service_placement_nfs.go
claudiubelu: Done

UtkarshBhatthere:
1. Not providing a target name does not default to the local host.
2. The service name should default to nfs rather than nfs-ganesha (as we enable `nfs` not `nfs-ganesha`).
```
root@microceph1:~# microceph enable nfs --cluster-id zuzu
Error: failed placing service nfs: Failed to get cluster member for request target name "0.0.0.0:2049": CoreClusterMember not found
root@microceph1:~# microceph enable nfs --cluster-id zuzu --target microceph1
root@microceph1:~# microceph status
MicroCeph deployment summary:
- microceph1 (10.17.213.183)
  Services: mds, mgr, mon, nfs-ganesha, osd
  Disks: 1
- microceph2 (10.17.213.150)
  Services: mds, mgr, mon, osd
  Disks: 1
- microceph3 (10.17.213.215)
  Services: mds, mgr, mon, osd
  Disks: 1
```

gboutry:
@UtkarshBhatthere Any updates? Can you review please?

UtkarshBhatthere:
@gboutry I will do a detailed review today

microceph/ceph/config.go
```diff
@@ -93,6 +93,7 @@ func GetConfigTableServiceSet() common.Set {
 		"mgr": struct{}{},
 		"osd": struct{}{},
 		"mds": struct{}{},
+		"nfs": struct{}{},
```
UtkarshBhatthere: I think this entry is not needed as it is only for the cluster configuration API.

microceph/ceph/service_placement_nfs.go
claudiubelu: Done

microceph/ceph/config.go
claudiubelu: Indeed, this is no longer required, it was needed before in `cmdDeleteService`, but we added `cmdNFSDeleteService`.

microceph/ceph/configwriter.go
```diff
@@ -97,3 +97,64 @@ rgw frontends = beast{{if or (ne .rgwPort 0) (not .sslCertificatePath) (not .ssl
...
+
+RADOS_URLS {
+	ceph_conf = "{{.confDir}}/ceph.conf";
+	UserId = "nfs.ganesha";
+	watch_url = "rados://.nfs/{{.clusterID}}/conf-nfs.{{.clusterID}}";
+}
+
+LOG {
+	Components {
+		ALL = FULL_DEBUG;
```
UtkarshBhatthere: For production environments deploying a service on full debug is not suggested.

microceph/ceph/configwriter_test.go
```diff
@@ -188,3 +188,59 @@ func (s *configWriterSuite) TestWriteCephKeyring() {
...
+	_, err = os.Stat(config.GetPath())
+	assert.Equal(s.T(), nil, err)
+	// Check contents of the file
+	data, err := os.ReadFile(config.GetPath())
+	assert.Equal(s.T(), nil, err)
+
+	dataStr := string(data)
+	assert.Contains(s.T(), dataStr, "Bind_Addr = 10.20.30.40;")
+	assert.Contains(s.T(), dataStr, "NFS_Port = 9999;")
+	assert.Contains(s.T(), dataStr, "Minor_Versions = 2;")
```
UtkarshBhatthere, line 221: A more intricate check could be to set minor version to 1 and check if both 1 and 2 are present in the minor version config.

microceph/ceph/configwriter_test.go
```diff
@@ -188,3 +188,59 @@ func (s *configWriterSuite) TestWriteCephKeyring() {
...
+}
+
+// Test ceph config writing for NFS Ganesha
+func (s *configWriterSuite) TestWriteGaneshaCephConfig() {
+	config := newGaneshaCephConfig(s.Tmp)
+
+	err := config.WriteConfig(
+		map[string]any{
+			"monitors": "foo",
+			"confDir":  "/foo/lish",
```
UtkarshBhatthere, line 238: haha

microceph/api/types/services.go
```diff
@@ -20,6 +20,11 @@ type EnableService struct {
 	// Enable Service passes all additional data as a json payload string.
 }
 
+// NFSService holds the Cluster ID of the NFS Service.
```
UtkarshBhatthere, line 29: The `EnableService` structure is sufficient for moving payload to the service placement interface. (as used in the cmd package's enable_nfs.go). This structure therefore can be removed (along with the RGW structure below as it is remnant of the past.

UtkarshBhatthere: aah, I see that this is being used in `disable` API calls.

UtkarshBhatthere: This can stay for the time being. We will have to come up with a cleaner unified approach for service disablement as well, but that is for later.

microceph/ceph/service_placement_nfs.go
```diff
@@ -0,0 +1,114 @@
...
+	if nfs.V4MinVersion > 2 {
+		return fmt.Errorf("expected v4_min_version to be in the interval [0, 2]")
+	}
+
+	if len(nfs.BindAddress) == 0 {
+		nfs.BindAddress = "0.0.0.0"
+	} else {
+		ip := net.ParseIP(nfs.BindAddress)
+		if ip == nil {
+			return fmt.Errorf("bind_address could not be parsed")
```
UtkarshBhatthere: The error string should also contain the user provided `bind_address` for clarity. something like `"bind_address %s could not be parsed"`

microceph/ceph/service_placement_nfs.go
```diff
@@ -0,0 +1,114 @@
...
+	return genericHospitalityCheck("nfs-ganesha")
+}
+
+func (nfs *NFSServicePlacement) ServiceInit(ctx context.Context, s interfaces.StateInterface) error {
+	monitors, err := GetMonitorAddresses(ctx, s)
+	if err != nil {
+		return err
+	}
+
+	return EnableNFS(s, nfs.ClusterID, nfs.BindAddress, nfs.BindPort, nfs.V4MinVersion, monitors)
```
UtkarshBhatthere: In case we add more parameters to the NFSServicePlacement structure, it may make sense to directly pass the structure to the `EnableNFS` function call.

microceph/ceph/service_placement_nfs.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func (nfs *NFSServicePlacement) DbUpdate(ctx context.Context, s interfaces.StateInterface) error {
+	bytes, err := json.Marshal(map[string]any{
+		"v4_min_version": nfs.V4MinVersion,
+	})
+	if err != nil {
+		return err
+	}
+	config := string(bytes)
```
UtkarshBhatthere: consider renaming `config` -> `group_config` and `info` -> `service_config`

microceph/ceph/service_placement_nfs.go
```diff
@@ -0,0 +1,114 @@
...
+
+	return EnableNFS(s, nfs.ClusterID, nfs.BindAddress, nfs.BindPort, nfs.V4MinVersion, monitors)
+}
+
+func (nfs *NFSServicePlacement) PostPlacementCheck(s interfaces.StateInterface) error {
+	return genericPostPlacementCheck("nfs-ganesha")
+}
+
+func (nfs *NFSServicePlacement) DbUpdate(ctx context.Context, s interfaces.StateInterface) error {
+	bytes, err := json.Marshal(map[string]any{
```
UtkarshBhatthere: I have my reservations against using string to string maps because they don't inspire confidence when we unmarshal them, being prone to case/spelling mismatches for keys. I find it worthwhile to create config structures (in the respective go database files) to 
A. Enforce a structure to this information. 
B. Take advantage of golang's struct <-> json features.

microceph/ceph/service_placement_nfs.go
```diff
@@ -0,0 +1,114 @@
...
+		"v4_min_version": nfs.V4MinVersion,
+	})
+	if err != nil {
+		return err
+	}
+	config := string(bytes)
+
+	bytes, err = json.Marshal(map[string]any{
```
UtkarshBhatthere: same as above.

microceph/ceph/nfs.go
```diff
@@ -0,0 +1,316 @@
...
+		}
+
+		return nil
+	})
+
+	return err
+}
+
+// createNFSServiceGroupRecord creates a record in the grouped_services database.
+func createNFSServiceGroupRecord(ctx context.Context, s interfaces.StateInterface, clusterID, info string) error {
```
UtkarshBhatthere: The presence of `ServiceGroup` creates a little confusion that this method is creating a record in the service_group table. While I am not great at naming things myself, I would suggest go with something like `createNFSGroupedServiceRecord` or something else which clearly suggests the information being recorded.

microceph/ceph/nfs.go
```diff
@@ -0,0 +1,316 @@
...
+	ganeshaConfDir := filepath.Join(pathConsts.ConfPath, "ganesha")
+
+	// Stop the NFS Ganesha service.
+	err := stopNFS()
+	if err != nil {
+		return err
+	}
+
+	// Remove the NFS Ganesha Ceph keyring.
+	err = os.Remove(filepath.Join(ganeshaConfDir, "keyring"))
```
UtkarshBhatthere: I think we should also remove the NFS-Ganesha ceph client keyring record from the `ceph auth` keys.

microceph/ceph/nfs.go
```diff
@@ -0,0 +1,316 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to remove NFS Ganesha Ceph configuration: %w", err)
+	}
+
+	err = os.Remove(filepath.Join(ganeshaConfDir, "ganesha.conf"))
+	if err != nil {
+		return fmt.Errorf("failed to remove NFS Ganesha configuration: %w", err)
+	}
+
+	err = deleteNFSServiceGroupRecord(ctx, s, clusterID)
```
UtkarshBhatthere, line 110: So here, we are first removing the Ser

microceph/ceph/nfs.go
```diff
@@ -0,0 +1,316 @@
...
+		}
+
+		return nil
+	})
+
+	return err
+}
+
+// deleteNFSServiceGroupRecord removes a record in the grouped_services database.
+func deleteNFSServiceGroupRecord(ctx context.Context, s interfaces.StateInterface, clusterID string) error {
```
UtkarshBhatthere: same as above, the function name does not clearly express the kind of record it removes.

microceph/ceph/nfs.go
UtkarshBhatthere: So in here, we first remove the grouped_service record and then (if necessary) remove the unreferenced service_groups record. While the order of the operation and the logic is correct, these should not be done in 2 separate transactions. The cleanup logic should run in a single transaction, thereby ensuring atomicity.

microceph/ceph/nfs.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+	// Create the keyring.
+	keyringPath := filepath.Join(path, "keyring")
+	if _, err := os.Stat(keyringPath); err == nil {
+		return nil
+	}
+
+	err := genAuth(
+		keyringPath,
+		"client.nfs.ganesha",
```
UtkarshBhatthere: The client keyring is not parametrized with `hostname` or `cluster-id`. This means that  we will get the same client key for all nfs-ganesha daemons in all nfs clusters. Should be of the form `client.nfs-<clusterid>.<hostname>` ensuring each deamon gets a unique ceph client key.

microceph/ceph/nfs.go
```diff
@@ -0,0 +1,316 @@
...
+}
+
+// osdEnablePoolApp enables the use of an application on the given pool.
+func osdEnablePoolApp(pool, app string) (string, error) {
+	return cephRun("osd", "pool", "application", "enable", pool, app)
+}
+
+// addNodeToSharedGraceMgmtDb adds the given node into the Shared Grace Management Database, which
+// is used by the rados_cluster recovery backend.
+func addNodeToSharedGraceMgmtDb(cephConfPath, clusterID, node string) error {
```
UtkarshBhatthere: tldr; This comment is pure nit-pickery

`addNodeToSharedGraceMgmtDb` should prepare the command parameters and pass to something like `ganeshaRadosGrace` which should be a thin wrapper around `ganesha-rados-grace` subprocess call.

This allows for the reuse of `ganeshaRadosGrace` method for other grace commands as well.

microceph/ceph/service_placement_nfs.go
```diff
@@ -0,0 +1,114 @@
...
+		"bind_address": nfs.BindAddress,
+		"bind_port":    nfs.BindPort,
+	})
+	if err != nil {
+		return err
+	}
+	info := string(bytes)
+
+	err = ensureNFSServiceGroupRecord(ctx, s, nfs.ClusterID, config)
```
UtkarshBhatthere: Similar to another comment about using 2 different transactions, an unwanted error  in the later record update may result in mismatched DB records. recommended is to perform both in a single transaction for atomicity.

microceph/database/schema.go
```diff
@@ -186,3 +187,28 @@ CREATE TABLE remote (
 
 	return err
 }
+
```
UtkarshBhatthere, line 190: For complicated DB transactions, there is a pattern of having `table_mapper_extras.go` (refer to database/client_config_extras.go) which abstract all the database logical complexities in the database module itself. This results in cleaner code in the `ceph` module as it is only responsible for calling them.

microceph/ceph/configwriter.go
claudiubelu: Done

microceph/ceph/configwriter_test.go
claudiubelu: Hmm, so, if the user specifies the `--v4-min-version=1`, the expectation is that we set `Minor_Versions = 1, 2;` here?

microceph/ceph/nfs.go
claudiubelu: Done

microceph/ceph/service_placement_nfs.go
claudiubelu: Done

microceph/ceph/service_placement_nfs.go
claudiubelu: Done

microceph/ceph/service_placement_nfs.go
claudiubelu: Done

microceph/ceph/nfs.go
claudiubelu: Done

microceph/ceph/nfs.go
claudiubelu: Ye, I've juggled `ServiceGroup` and `GroupedService` names around enough to confuse them. :) The name was meant to be `createNFSGroupedServiceRecord`.

microceph/ceph/nfs.go
claudiubelu: Same, it should have been `deleteNFSGroupedServiceRecord`.

microceph/ceph/nfs.go
claudiubelu: Done

microceph/ceph/nfs.go
claudiubelu: Done

microceph/ceph/service_placement_nfs.go
claudiubelu: Done

microceph/ceph/service_placement_nfs.go
claudiubelu: Done

microceph/ceph/service_placement_nfs.go
claudiubelu: Done

microceph/database/schema.go
claudiubelu: Done

microceph/ceph/configwriter.go
```diff
@@ -97,3 +97,69 @@ rgw frontends = beast{{if or (ne .rgwPort 0) (not .sslCertificatePath) (not .ssl
...
+
+NFSv4 {
+	Delegations = false;
+	RecoveryBackend = 'rados_cluster';
+	Minor_Versions = {{.minorVersions}};
+	UseGetpwnam = true;
+}
+
+NFS_KRB5 {
+	CCacheDir = /var/snap/microceph/current/run/ganesha;
```
sabaini: Please use a variable for the /var/snap/microceph/current/run path, ie. the PathConst.RunPath -- similar to how EnableRGW()

microceph/ceph/nfs.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to stop NFS Ganesha service: %w", err)
+	}
+
+	return nil
+}
+
+// createNFSKeyring creates the NFS keyring.
+func createNFSKeyring(path, clusterID, userID string) error {
+	if err := os.MkdirAll(path, 0770); err != nil {
```
sabaini: Minor nit: please use the sep. assign / conditional style for consistency

```
_, err := os.Stat(keyringPath)
if err != nil { ... }
```

Same below

microceph/ceph/nfs_test.go
```diff
@@ -0,0 +1,352 @@
...
+	clientUser := fmt.Sprintf("client.%s", userID)
+	r.On("RunCommand", "ceph", "auth", "del", clientUser).Return("ok", nil).Once()
+
+	// patch processExec
+	processExec = r
+
+	// function call
+	err = DisableNFS(context.Background(), s.TestStateInterface, clusterID)
+
+	assert.EqualError(s.T(), err, "no server certificate")
```
sabaini: I believe `no server certificate` is an error message from the uninitialized database. 
Would it be possible to mock the DB so we can get a clean test run?

microceph/ceph/service_placement_nfs_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+	nfs := NFSServicePlacement{
+		ClusterID:    "foo",
+		V4MinVersion: 2,
+		BindAddress:  "42.42.42.42",
+		BindPort:     9999,
+	}
+
+	err := nfs.DbUpdate(context.Background(), s.TestStateInterface)
+	assert.EqualError(s.T(), err, "no server certificate")
```
sabaini: Similar to above, ideally we'd be able to mock the database here

microceph/ceph/nfs.go
```diff
@@ -0,0 +1,230 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/canonical/microceph/microceph/constants"
+	"github.com/canonical/microceph/microceph/database"
+	"github.com/canonical/microceph/microceph/interfaces"
+)
+
+// EnableNFS enables the NFS Ganesha service on the cluster and adds initial configuration.
+func EnableNFS(s interfaces.StateInterface, nfs *NFSServicePlacement, monitorAddresses []string) error {
```
sabaini, line 20: It would be great if we could make use of the lxd revert module here. This would allow us to rollback changes to the system and cluster in case of faults.

E.g. [similar to the AddOSD() func](https://github.com/canonical/microceph/blob/4910cfbd937860690fba764bd9c85ace5707f969/microceph/ceph/osd.go#L531)

microceph/database/service_group.mapper.go
```diff
@@ -0,0 +1,390 @@
...
+		}
+
+		objects = append(objects, s)
+
+		return nil
+	}
+
+	err := query.SelectObjects(ctx, stmt, dest, args...)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to fetch from \"services_groups\" table: %w", err)
```
sabaini: Typo: service_groups

microceph/database/service_group.mapper.go
```diff
@@ -0,0 +1,390 @@
...
+		}
+
+		objects = append(objects, s)
+
+		return nil
+	}
+
+	err := query.Scan(ctx, tx, sql, dest, args...)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to fetch from \"services_groups\" table: %w", err)
```
sabaini: Typo: service_groups

snapcraft/commands/nfs-ganesha.start
```diff
@@ -0,0 +1,17 @@
...
+
+wait_for_config
+
+conf="${SNAP_DATA}/conf/ganesha/ganesha.conf"
+logs="${SNAP_COMMON}/logs/ganesha/ganesha.log"
+pid_file="${SNAP_DATA}/run/ganesha/ganesha.pid"
+
+mkdir -p "${SNAP_DATA}/run/ganesha"
+
+if [ -n "${conf}" ]; then
```
sabaini, line 15: Hm, when would the $conf not be available?

tests/scripts/actionutils.sh
```diff
@@ -98,6 +98,39 @@ function add_lvm_vol() {
...
+
+    # Create a new NFS FS with the microceph NFS rados pools.
+    local fsname="${1}"
+    sudo microceph.ceph fs new "${fsname}" .nfs.metadata .nfs
+}
+
+function create_nfs_export() {
+    set -x
+
+    # Create a new NFS FS with the microceph NFS rados pools.
```
sabaini, line 162: Hm, is that comment accurate?

microceph/ceph/nfs.go
claudiubelu: Done

tests/scripts/actionutils.sh
claudiubelu: Missed this comment, thanks.

snapcraft/commands/nfs-ganesha.start
claudiubelu: If nfs is not enabled (`microceph enable nfs`), then the conf file shouldn't exist.

microceph/database/service_group.mapper.go
claudiubelu: Indeed. This code is autogenerated though, but will fix it.

microceph/database/service_group.mapper.go
claudiubelu: ditto

microceph/ceph/configwriter.go
claudiubelu: done

microceph/ceph/configwriter_test.go
UtkarshBhatthere: yes, the flag is to get the minimum minor version that should be supported by the client. Hence if they mention 0, it should support 0,1,2... if 1 then 1,2 and only 2 if they request 2.

microceph/ceph/nfs.go
claudiubelu: Done

.github/workflows/tests.yml
```diff
@@ -813,13 +813,6 @@ jobs:
     - name: Install and setup
       run: ~/actionutils.sh install_microceph
 
-    - name: Add loopback file OSDs
```
UtkarshBhatthere: Can you please describe why a test step is removed from an unrelated testcase ?

claudiubelu: Oups. :)

microceph/ceph/nfs_test.go
claudiubelu: Done

microceph/ceph/service_placement_nfs_test.go
claudiubelu: Done

microceph/ceph/configwriter_test.go
claudiubelu: I see. Updated the code in `nfs.go` for this, but this test only verifies that we're indeed writing in the file what we're supposed to.

.github/workflows/tests.yml
```diff
@@ -895,3 +888,50 @@ jobs:
...
+
+    - name: Enable NFS
+      run: ~/actionutils.sh enable_nfs foo
+
+    - name: Create NFS FS
+      run: ~/actionutils.sh create_nfs_fs testfs
+
+    - name: Create NFS export
+      run: ~/actionutils.sh create_nfs_export foo testfs
+
```
UtkarshBhatthere, line 942: As a test step we should also mount the nfs export created (`foo`)  on the previous step and perform IO to verify that it works.

This will require:
1. `nfs-common` package to be installed on the host.
2. `sudo mount -t nfs 10.10.0.10:/<psuedo> /path/to/mountpoint`

microceph/ceph/nfs.go
```diff
@@ -0,0 +1,287 @@
...
+	hostname, err := os.Hostname()
+	if err != nil {
+		return err
+	}
+
+	pathConsts := constants.GetPathConst()
+	ganeshaConfDir := filepath.Join(pathConsts.ConfPath, "ganesha")
+
+	// Stop the NFS Ganesha service.
+	logger.Debugf("Stopping NFS Ganesha service (ClusterID '%s')", clusterID)
```
UtkarshBhatthere, line 152: The method `DisableNFS` does not verify whether the nfs daemon runnning on the host belongs to the provided cluster-id or not. This may cause incorrect service disablement if an operator (by mistake) passes the wrong cluster-id on the wrong host.

.github/workflows/tests.yml
claudiubelu: Done

microceph/database/grouped_service_extras.go
```diff
@@ -0,0 +1,120 @@
...
+	"net/http"
+
+	"github.com/canonical/microceph/microceph/interfaces"
+
+	"github.com/canonical/lxd/shared/api"
+)
+
+var _ = api.ServerEnvironment{}
+
+//go:generate mockery --name GroupedServiceQueryIntf
```
UtkarshBhatthere, line 17: There should also be a GetGroupedServices method as well to fetch a slice of configured Grouped services. This would be used to populate the `microceph status` output. The API for this is `1.0/services`. The output should contain the service configurations (like bind address and ports as well) as that will be used by clients like `manila` eventually.

UtkarshBhatthere (changes requested):
some comments

microceph/ceph/nfs.go
```diff
@@ -56,29 +72,66 @@ func EnableNFS(s interfaces.StateInterface, nfs *NFSServicePlacement, monitorAdd
...
+	}
+
+	revert.Success() // Added revert functions are not run on return.
+
+	logger.Debugf("Enabled NFS on node with ClusterID '%s'", nfs.ClusterID)
+
+	return nil
+}
+
+func nfsVersionsStr(minVersion uint) string {
```
UtkarshBhatthere, line 123: I think it should be:
```

func nfsVersionsStr(minVersion uint) string {
	var versions []string

	for i := minVersion; i <= 2; i++ {
		versions = append(versions, strconv.FormatUint(uint64(i), 10))
	}

	return strings.Join(versions, ",")
}
```

UtkarshBhatthere: given that min version value is under the specified range is already been checked.

UtkarshBhatthere: i.e. 0 <= minVersion <= 2

claudiubelu: Oups. :) Done

microceph/ceph/nfs.go
claudiubelu: Done

microceph/database/grouped_service_extras.go
claudiubelu: Added

UtkarshBhatthere:
# Enabling a second daemon (on a multinode microceph cluster fails)
@claudiubelu 

## Node 1
```
root@secondary:~# microceph enable nfs --cluster-id manila
root@secondary:~#
```

## Node 2
```
# enablement on second cluster fails
root@primary:~# microceph enable nfs --cluster-id manila
Error: failed placing service nfs: failed to add DB record for nfs: conflicting service group configurations
root@primary:~#
```

## Database Dump
```
root@primary:~# microceph cluster sql "select * from grouped_services"
+----+------------------+-----------+---------------------------------------------+
| id | service_group_id | member_id |                    info                     |
+----+------------------+-----------+---------------------------------------------+
| 1  | 1                | 2         | {"bind_address":"0.0.0.0","bind_port":2049} |
+----+------------------+-----------+---------------------------------------------+
root@primary:~# microceph cluster sql "select * from service_groups"
+----+---------+----------+----------------------+
| id | service | group_id |        config        |
+----+---------+----------+----------------------+
| 1  | nfs     | manila   | {"v4_min_version":1} |
+----+---------+----------+----------------------+
root@primary:~#
```

microceph/ceph/service_placement_nfs.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (nfs *NFSServicePlacement) PopulateParams(s interfaces.StateInterface, payload string) error {
+	err := json.Unmarshal([]byte(payload), &nfs)
+	if err != nil {
+		return err
+	}
+
+	if len(nfs.ClusterID) == 0 {
+		return fmt.Errorf("expected cluster_id to be non-empty")
+	}
```
sabaini, line 31: While the ceph docs say ClusterID can be an arbitrary string, I believe the clusterid ends up in pool and user names, also it gets rendered into the config file and cmdlines, so there's some risk here (ie. the bobby tables effect). 

I would suggest to constrain which chars to accept e.g. the POSIX portable filename set, ensure it doesn't start with a period `.` as that is for internal pools, and constrain length to something sensible like 63 chars

microceph/cmd/microceph/enable_nfs.go
```diff
@@ -0,0 +1,92 @@
...
+	client           *microclusterclient.Client
+}
+
+func (c *cmdEnableNFS) Command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "nfs --cluster-id <cluster-id> [--bind-address <ip-address>] [--port <port-num>] [--v4-min-version 0/1/2] [--target <server>] [--wait <bool>]",
+		Short: "Enable the NFS Ganesha service on the --target server (default: this server)",
+		RunE:  c.Run,
+	}
+	cmd.PersistentFlags().StringVar(&c.flagClusterID, "cluster-id", "", "NFS Cluster ID")
```
sabaini: Possibly should add some doc on which names are allowed

microceph/ceph/nfs.go
```diff
@@ -0,0 +1,305 @@
...
+		[]string{"osd", fmt.Sprintf("allow rw pool=.nfs namespace=%s", clusterID)})
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// ensureNFSPools creates the NFS data and metadata Pools for Ganesha if they do not exist.
+func ensureNFSPools(clusterID string) error {
```
sabaini, line 247: May I ask for a logger.Debugf() call here too as this is a p. crucial func

microceph/ceph/nfs.go
```diff
@@ -0,0 +1,305 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to enable .nfs pool: %w", err)
+	}
+
+	object := fmt.Sprintf("conf-nfs.%s", clusterID)
+	_, err = radosRun("create", "--pool", ".nfs", "-N", clusterID, object)
+	if err != nil && !strings.Contains(err.Error(), "File exists") {
+		return fmt.Errorf("failed to create object for Ganesha: %w", err)
+	}
+
```
sabaini, line 273: ...and logger.Debugf() on exit too please

sabaini (changes requested):
Thanks @claudiubelu ! Added some commentary

microceph/ceph/nfs.go
claudiubelu: Done

microceph/ceph/nfs.go
claudiubelu: Done

microceph/ceph/service_placement_nfs.go
claudiubelu: Done

microceph/cmd/microceph/enable_nfs.go
claudiubelu: Done

Issue #547, opened 2025-04-25
Enable pool level replication using type snapshot


roosterfish:
Hi, from the [docs](https://canonical-microceph.readthedocs-hosted.com/en/squid-stable/how-to/configure-rbd-mirroring/) it was not clear how to enable replication using type snapshot so I ended up trying to enable it on pool level which (according to upstream [docs](https://docs.ceph.com/en/latest/rbd/rbd-mirroring/)) looks to be not supported.

However MicroCeph allows running the following command to enable replication using type snapshot on pool level:

`microceph replication enable rbd foo --type snapshot --remote secondary`

But instead the command seems to trigger the setup of `--type journal` instead as pre-existing volumes of the pool `foo` now have the `journaling` feature enabled and their mirroring mode (`rbd info foo/test`) also displays `mirroring mode: journal`.

I have tested using MicroCeph `squid/stable`. Those are the reproducer steps on the two Ceph clusters `micro01` and `micro02`:

```bash
# micro01

microceph cluster bootstrap
microceph disk add /dev/sdb /dev/sdc /dev/sdd
microceph remote import secondary <token from micro02> --local-name primary
microceph cluster export secondary

microceph enable rbd-mirror
ceph osd pool create foo 32
rbd pool init foo

rbd create foo/test --size 10M
rbd create foo/test2 --size 10M

root@micro01:~# rbd info foo/test
rbd image 'test':
	size 10 MiB in 3 objects
	order 22 (4 MiB objects)
	snapshot_count: 0
	id: 375c50cac2b3
	block_name_prefix: rbd_data.375c50cac2b3
	format: 2
	features: layering, exclusive-lock, object-map, fast-diff, deep-flatten
	op_features: 
	flags: 
	create_timestamp: Fri Apr 25 07:15:51 2025
	access_timestamp: Fri Apr 25 07:15:51 2025
	modify_timestamp: Fri Apr 25 07:15:51 2025

microceph replication enable rbd foo --type snapshot --remote secondary # fails without error

root@micro01:~# rbd info foo/test
rbd image 'test':
	size 10 MiB in 3 objects
	order 22 (4 MiB objects)
	snapshot_count: 0
	id: 375c50cac2b3
	block_name_prefix: rbd_data.375c50cac2b3
	format: 2
	features: layering, exclusive-lock, object-map, fast-diff, deep-flatten, journaling
	op_features: 
	flags: 
	create_timestamp: Fri Apr 25 07:15:51 2025
[...]

PR #575, opened 2025-06-25
fix: prevent enabling snapshot replication on RBD pools

panagiotisevaggelou:
# Description

This change adds a validation check to prevent users from enabling snapshot-based replication on RBD pools, which is currently unsupported. If a user attempts this, the system now returns a clear and descriptive error message.

This improves user experience by avoiding silent failures and preventing unexpected fallback to journal mode.

Fixes: #547

## Type of change

- ✅ Bug fix (non-breaking change which fixes an issue)

## How has this been tested?

I tested this manually using my local build of MicroCeph (custom snap package) installed in `--devmode` in two clean Multipass VMs:

- **micro01** and **micro02** were set up as separate Ceph clusters using `microceph cluster bootstrap`.
- Fake disks were created using loopback devices on both VMs (`fallocate` + `losetup`) and added using `microceph disk add`.
- The two clusters were linked via exported tokens using `microceph remote import` and `cluster export`.
- On `micro01`, I created a pool (`foo`) and two RBD images (`test`, `test2`).
- Before replication, I confirmed the images had no journaling features via `microceph.rbd info foo/test`.

Then I ran:
```bash
sudo microceph replication enable rbd foo --type snapshot --remote secondary
```
✅ The command failed as expected with the message:

```bash
Error: failed to process enable_replication request for rbd: Snapshot-based replication is only supported for individual RBD images, not pools
```
✅ After the command, `microceph.rbd info foo/test` confirmed that journaling was not silently enabled — verifying the bug is now prevented.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -878,6 +878,9 @@ jobs:
     - name: Enable RBD Mirror Daemon
       run : ~/actionutils.sh remote_enable_rbd_mirror_daemon
 
+    - name: Verify snapshot replication on pool fails
+      run: ~/actionutils.sh remote_verify_snapshot_pool_replication_fails
+
     - name: Configure RBD mirror
       run : ~/actionutils.sh remote_configure_rbd_mirroring
 
--- a/microceph/ceph/replication_rbd.go
+++ b/microceph/ceph/replication_rbd.go
@@ -132,7 +132,11 @@ func (rh *RbdReplicationHandler) EnableHandler(ctx context.Context, args ...any)
 
 	logger.Infof("REPRBD: Local(%s) Remote(%s)", dbRec[0].LocalName, dbRec[0].Name)
 	if rh.Request.ResourceType == types.RbdResourcePool {
-		return handlePoolEnablement(rh, dbRec[0].LocalName, dbRec[0].Name)
+		if rh.Request.ReplicationType == types.RbdReplicationSnapshot {
+			return fmt.Errorf("Snapshot-based replication is only supported for individual RBD images, not pools")
+		} else {
+			return handlePoolEnablement(rh, dbRec[0].LocalName, dbRec[0].Name)
+		}
 	} else if rh.Request.ResourceType == types.RbdResourceImage {
 		return handleImageEnablement(rh, dbRec[0].LocalName, dbRec[0].Name)
 	}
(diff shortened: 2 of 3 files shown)


panagiotisevaggelou:
Forgot to sign-off my commit, I will after pipeline finishes.🙂

panagiotisevaggelou:
Also don't know why [Test reef upgrades](https://github.com/canonical/microceph/actions/runs/15878746904/job/44774155016?pr=575#logs) failed previously.

UtkarshBhatthere:
@panagiotisevaggelou ```55f3fca19c010f8df7c2f8fcfbb585e794baa547    Expected "panagiotisevaggelou <panagiotisevaggelou02@gmail.com>", but got "Panagiotis Evaggelou <panagiotisevaggelou02@gmail.com>".```

.github/workflows/tests.yml
```diff
@@ -878,6 +878,10 @@ jobs:
     - name: Enable RBD Mirror Daemon
       run : ~/actionutils.sh remote_enable_rbd_mirror_daemon
 
+    - name: Ensure snapshot replication on pool fails
+      run: |
+        ! microceph replication enable rbd foo --type snapshot --remote site-b
```
UtkarshBhatthere:
```
Run ! microceph replication enable rbd foo --type snapshot --remote site-b
  ! microceph replication enable rbd foo --type snapshot --remote site-b
  shell: /usr/bin/bash -e {0}
/home/runner/work/_temp/62fd9cdf-69c9-43a2-9380-64cc615b4def.sh: line 1: microceph: command not found
```
Since we use LXD based virtual machines to run 2 ceph clusters (for replication) runnning the microceph command directly on the runner host will fail. Please check script `actionutils.sh` (functions like remote_enable_rbd_mirror_daemon may give a hint) for how this could be done from an LXD machine.

UtkarshBhatthere (changes requested):
Left a comment on the testing strategy, also, it is unwise to just check if the command failed (many reasons). It would better if we grep the failure string for `Snapshot-based replication is only supported for individual RBD images, not pools` just to be sure that the failure is as expected.

.github/workflows/tests.yml
panagiotisevaggelou: @UtkarshBhatthere, thanks for the feedback!
Updated the test to run microceph inside the LXD container via `lxc exec` since the command isn’t available on the runner host.
Also added a `grep` check to verify the specific error message.
Let me know if any changes are needed!

PR #576, opened 2025-06-26
fix: prevent using the same name for local alias and remote cluster

### LINKED ISSUE #559: Fault enabling replication
Two test clusters on nodes tst-0 and tst-1

I have imported a remote on tst-1 into my cluster, giving it a name like this:

```
ubuntu@tst-1:~$ sudo microceph remote import --local-name tst0 tst0 <token>
```

I can see the remote from the tst-1 machine:

```
ubuntu@tst-1:~$ sudo ceph -s --cluster tst0 --id tst0
  cluster:
    id:     28260466-321a-44c2-b2aa-d8998d3c0fd7
    health: HEALTH_WARN
            1 pool(s) do not have an application enabled
 
  services:
    mon: 1 daemons, quorum tst-0 (age 25m)
    mgr: tst-0(active, since 24m)
    osd: 3 osds: 3 up (since 24m), 3 in (since 24m)
 
  data:
    pools:   2 pools, 33 pgs
    objects: 3 objects, 577 KiB
    usage:   81 MiB used, 5.9 GiB / 6 GiB avail
    pgs:     33 active+clean
```

I have created a remote on the other end as well:

```
ubuntu@tst-0:~$ sudo ceph -s --cluster tst1 --id tst1
  cluster:
    id:     0a429727-8cf3-412f-b850-90343906716a
    health: HEALTH_WARN
            1 pool(s) do not have an application enabled
 
  services:
    mon: 1 daemons, quorum tst-1 (age 28m)
    mgr: tst-1(active, since 28m)
    osd: 3 osds: 3 up (since 27m), 3 in (since 27m)
 
  data:
    pools:   2 pools, 33 pgs
    objects: 3 objects, 577 KiB
    usage:   81 MiB used, 5.9 GiB / 6 GiB avail
    pgs:     33 active+clean

```
But I can't enable rbd replication:

```
ubuntu@tst-1:~$ sudo microceph replication enable rbd rbd --remote tst0
Error: failed to process enable_replication request for rbd: failed to import peer bootstrap token: Failed to run: rbd mirror pool peer bootstrap import --site-name tst0 --direction rx-tx rbd /var/snap/microceph/1355/conf/rbd_mirror/tst0_peer_keyring --cluster tst0 --id tst0: exit status 22 (2025-05-15T13:39:19.814+0000 7f14775ff680 -1 librbd::api::Mirror: peer_bootstrap_import: cannot import token for duplicate site name
rbd: failed to import peer bootstrap token)
```

panagiotisevaggelou:
# Description

This PR introduces a validation in the remote import CLI command to prevent users from specifying the same value for the --local-name flag and the remote cluster name (positional argument). This resolves an issue where Ceph replication fails due to duplicate site names during peer bootstrap.

Additionally, the relevant documentation (Import a remote MicroCeph cluster) was updated to include a note that clarifies this requirement for users.

Fixes #559

## Type of change

## How has this been tested?

The validation was tested locally by building the microceph binary and running the remote import command with:
- Identical --local-name and remote name: correctly triggers a validation error
- Distinct values: proceeds past validation as expected

### DIFF
--- a/microceph/cmd/microceph/remote_import.go
+++ b/microceph/cmd/microceph/remote_import.go
@@ -39,6 +39,10 @@ func (c *cmdRemoteImport) Run(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("please provide a local name using `--local-name` flag")
 	}
 
+	if c.localName == args[0] {
+		return fmt.Errorf("local alias (--local-name) and remote name must be different to avoid site conflicts")
+	}
+
 	m, err := microcluster.App(microcluster.Args{StateDir: c.common.FlagStateDir})
 	if err != nil {
 		return err
--- a/docs/how-to/import-remote-cluster.rst
+++ b/docs/how-to/import-remote-cluster.rst
@@ -26,6 +26,11 @@ At the primary cluster, this token can be imported to create the remote record.
 
    sudo microceph remote import simple eyJmc2lkIjoiN2FiZmMwYmItNjIwNC00M2FmLTg4NDQtMjg3NDg2OGNiYTc0Iiwia2V5cmluZy5jbGllbnQubWFnaWNhbCI6IkFRQ0hJdmRtNG91SUNoQUFraGsvRldCUFI0WXZCRkpzUC92dDZ3PT0iLCJtb24uaG9zdC5zaW1wbGUtcmVpbmRlZXIiOiIxMC40Mi44OC42OSIsInB1YmxpY19uZXR3b3JrIjoiMTAuNDIuODguNjkvMjQifQ== --local-name magical
 
+.. note::
+
+   The value of ``--local-name`` must be different from the remote cluster name (``simple`` in this example).
+   Using the same name for both will result in an error to avoid conflicts during replication.
+
 This will create the required $simple.conf and $simple.keyring files.
 Note: Importing a remote cluster is a uni-directional operation. For symmetric
 relations both clusters should be added as remotes at each other.


microceph/cmd/microceph/remote_import.go
```diff
@@ -39,6 +39,10 @@ func (c *cmdRemoteImport) Run(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("please provide a local name using `--local-name` flag")
 	}
 
+	if c.localName == args[0] {
```
UtkarshBhatthere: this check is client sided (CLI), which still leaves the backend API open to such mistakes from a user. Please check `cmdOpsReplication` in `microceph/api/ops_replication.go` to add this validation in the API.

UtkarshBhatthere (changes requested):
Thanks for the PR. suggestions.

microceph/cmd/microceph/remote_import.go
panagiotisevaggelou: @UtkarshBhatthere , I've added the validation at the API level as requested. Do you think it's worth keeping the CLI-side check as well for quicker feedback to CLI users?

panagiotisevaggelou: @UtkarshBhatthere, I've added the validation at the API level as requested. The empty check prevents errors from cases where the remote name is empty, which caused a test to fail previously. Also included LocalAlias in the struct to explicitly track and use the local cluster name.

PR #580, opened 2025-07-02
Add our product page to header

skoech:
# Description

The docs starter pack gives us an option to include our product page in the header of of our docs site. We agreed as a team to point to microceph.com because it's more of a homepage, and it's suitable for a broader audience - including for marketing reasons, which are not satisfied by our docs site.

## Type of change

## How has this been tested?

- I have built the docs locally to make sure it works.

## Contributor checklist

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -101,7 +101,7 @@ html_context = {
     # TODO: If there's no such website,
     #       remove the {{ product_page }} link from the page header template
     #       (usually .sphinx/_templates/header.html; also, see README.rst).
-    "product_page": "",
+    "product_page": "microceph.com",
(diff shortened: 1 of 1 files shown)


UtkarshBhatthere (approved):
LGTM

UtkarshBhatthere:
So if you open [microceph.com] today it opens the documentation homepage which is good. But having the same link then in the product page makes you jump back to doc homepage when you click on the product link ?

Maybe use the snapcraft landing page as the product page as MicroCeph is a snap only artefact ?

opinions and discussion welcome.

UtkarshBhatthere (changes requested):
some things need to be cleared out.

PR #583, opened 2025-07-16
docs: Updates documentation to include information about enabling NFS

claudiubelu:
# Description

Adds information about how to `enable nfs` in MicroCeph.

Depends-On: https://github.com/canonical/microceph/pull/574

## Type of change

## How has this been tested?

N/A

## Contributor checklist

Please check that you have:

### DIFF
--- a/docs/how-to/enable-service-instances.rst
+++ b/docs/how-to/enable-service-instances.rst
@@ -16,6 +16,7 @@ Cluster designs that call for extra service instances, however, can be
 satisfied by manual means. In addition to the above-listed services, the
 following service can be added manually to a node:
 
+* NFS
 * RGW (`RADOS Gateway service`_)
 
 This is the purpose of the :command:`enable` command. It manually enables a
@@ -27,9 +28,9 @@ The syntax is:
 
    sudo microceph enable <service> --target <destination> ...
 
-Where the service value is one of 'mon', 'mds', 'mgr', and 'rgw'. The
-destination is a node name as discerned by the output of the :command:`status`
-command:
+Where the service value is one of 'mon', 'mds', 'mgr', 'nfs-<cluster-id>' and
+'rgw'. The destination is a node name as discerned by the output of the
+:command:`status` command:
 
 .. code-block:: none
 
@@ -43,8 +44,11 @@ service:
 
    sudo microceph enable <service> --help
 
-Example: enable an RGW service
-------------------------------
+Let's take an example of enabling RGW and NFS services, viewing the possible
+extra parameters for both services.
+
+Enable an RGW service
+---------------------
 
 First check the status of the cluster to get node names and an overview of
 existing services:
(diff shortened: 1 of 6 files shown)


UtkarshBhatthere:
@skoech please take a look.

docs/how-to/enable-service-instances.rst
```diff
@@ -88,6 +89,74 @@ Finally, view cluster status again and verify expected changes:
      Services: mds, mgr, mon
      Disks: 0
 
+Example: enable a NFS service
```
skoech: Prefacing the title with "Example:" seems to me like it is continuing an unfinished thought from the previous paragraph. I think the title should stand on its own, i.e. tell the user firmly what to expect next after a concluded thought, even if they are related.
I realise that the previous title was framed this way (Example: enable an RGW service), but I think this should change, too, and be reframed to "Enable a RGW service". The paragraph before this title should prepare the user for examples, so that the titles dive straight into their topic.
E.g. For a given service, the enable command may support extra parameters. These can be discovered by querying for help for the respective service:

`sudo microceph enable <service> --help`

Let's take an example of enabling RGW and NFS services, viewing the possible extra parameters for both service.

## Enable an RGW service
text

## Enable an NFS service
text

Also, since the commands and output is mostly the same for enabling both services, you may want to leave out the second example; no need for both unless there's a something different I am missing that you'd like the user to see.

docs/how-to/enable-service-instances.rst
```diff
@@ -88,6 +89,74 @@ Finally, view cluster status again and verify expected changes:
...
+
+   MicroCeph deployment summary:
+   - node1-2c3eb41e-14e8-465d-9877-df36f5d80922 (10.111.153.78)
+     Services: mds, mgr, mon, nfs.foo osd
+     Disks: 3
+   - workbook (192.168.29.152)
+     Services: mds, mgr, mon
+     Disks: 0
+
+For the next steps (creating a share and mounting it), consult the
```
skoech: This should go to the very end of the guide, after everything important to this guide (including the cautions and warning have been mentioned.

docs/reference/commands/disable.rst
```diff
@@ -15,6 +15,7 @@ Available Commands:
 
 .. code-block:: none
 
+   nfs         Disable the NFS Ganesha service on the --target server (default: this server)
```
skoech, line 18: Perhaps format `--target` like this to show that it is a flag?

docs/how-to/enable-service-instances.rst
claudiubelu: Done

docs/how-to/enable-service-instances.rst
claudiubelu: Done

skoech (changes requested):
Thank you, @claudiubelu. Other than some minor formatting inconsistencies, LGTM! 
Refer to https://canonical-starter-pack.readthedocs-hosted.com/latest/reference/rst-syntax-reference/#inline-formatting

PR #584, opened 2025-07-17
feat: Added microceph orchestrator module to snap build

UtkarshBhatthere:
```
ubuntu@juju-64bf41-test-1:~$ sudo ceph orch host ls
HOST                ADDR          LABELS  STATUS
juju-64bf41-test-0  10.149.6.228          Online
juju-64bf41-test-1  10.149.6.176          Online
juju-64bf41-test-2  10.149.6.149          Online
3 hosts in cluster
ubuntu@juju-64bf41-test-1:~$ sudo ceph orch ls
NAME        PORTS  RUNNING  REFRESHED  AGE  PLACEMENT
mds                    3/0  -          -    juju-64bf41-test-2;juju-64bf41-test-1;juju-64bf41-test-0;count:3
mgr                    3/0  -          -    juju-64bf41-test-2;juju-64bf41-test-1;juju-64bf41-test-0;count:3
mon                    3/0  -          -    juju-64bf41-test-2;juju-64bf41-test-1;juju-64bf41-test-0;count:3
nfs.manila             3/0  -          -    juju-64bf41-test-2;juju-64bf41-test-1;juju-64bf41-test-0;count:3
rbd-mirror             1/0  -          -    juju-64bf41-test-1;count:1
ubuntu@juju-64bf41-test-1:~$ sudo ceph nfs cluster ls
[
  "manila"
]
ubuntu@juju-64bf41-test-1:~$ sudo ceph nfs cluster info manila
{
  "manila": {
    "backend": [
      {
        "hostname": "juju-64bf41-test-2",
        "ip": "10.149.6.149",
        "port": 2049
      },
      {
        "hostname": "juju-64bf41-test-1",
        "ip": "10.149.6.176",
        "port": 2049
      },
      {
        "hostname": "juju-64bf41-test-0",
        "ip": "10.149.6.228",
        "port": 2049
      }
    ],
    "virtual_ip": null
  }
}
```


microceph-orch/src/microceph/client/service.py
```diff
@@ -0,0 +1,218 @@
...
+    """BaseService is the base service class for sunbeam clusterd services."""
+
+    def __init__(
+        self,
+        session: Session,
+        endpoint: str,
+        certs=None,
+        timeout: int | float | None = None,
+    ):
+        """Creates a new BaseService for the sunbeam clusterd API.
```
wolsen: In initial draft, this client was taken from the sunbeam code base, but the references to sunbeam clusterd will feel out of place to future authors. I propose we remove the references and potentially get rid of some of the token/node management pieces. They can always be added back later if / when it is necessary.

snap/snapcraft.yaml
```diff
@@ -179,6 +179,23 @@ apps:
...
+      - microceph
+      - ceph
+    override-prime: |
+      set -eux
+      a=$(pwd)
+      craftctl default
+      mkdir -p share/ceph/mgr/microceph
+      # Force copy to overwrite.
+      cp -rf lib/python3.12/site-packages/microceph share/ceph/mgr/
+      cp -rf lib/python3.12/site-packages/orchestrator share/ceph/mgr/
```
wolsen: I'm a little wary of including / copying the orchestrator module here. I believe that this is included due to the need to enable microceph orchestrator as a valid orchestrator. Instead, I propose we apply a patch so that we don't have to refresh the orchestrator module routinely - otherwise we may miss out or remove upstream fixes to this python package.

wolsen (changes requested):
Some small comments in line

snap/snapcraft.yaml
UtkarshBhatthere: The diff is actually very simple, (added `microceph` to an allowed enum), a patch here makes a lot of sense. Thanks.

UtkarshBhatthere: https://github.com/canonical/microceph/pull/584/files#diff-9f7543f993148cf2fef1b5d3905f48427441a692c5b21910e2b43bc7fc81e082R354

microceph-orch/src/microceph/client/service.py
UtkarshBhatthere: ack, done. Dropped unrelated references and token methods.

snap/snapcraft.yaml
UtkarshBhatthere: for the moment, we have moved to consuming the modified orchestrator from a ppa.

wolsen: I think the PPA works (as you implied its for the moment) - but a patch file being applied is probably 100x simpler to maintain for the time being.

I have a patch with a proposal for this [here](https://github.com/wolsen/microceph/commit/847c1b4da3c180eb1064f7c379f2ad3f17354650) - though it's not a PR because I haven't had time to do more than verify the resulting snap file is correct.

@UtkarshBhatthere thoughts?

wolsen:
Reviewed and proposed a patching approach in a separate branch.

I think we need at least a unit test that enables the orchestrator and runs a command.

snap/snapcraft.yaml
UtkarshBhatthere: well I must say, using patches in snapcraft build itself is kinda cool. I have verified your commit. It worked! so i stole it :D

microceph-orch/src/microceph/client/client.py
```diff
@@ -0,0 +1,116 @@
...
+def to_file_path_certs(certificate: str, private_key: str) -> tuple[str, str]:
+    """Template certpair in tmp files, return the path."""
+    cert_fd, cert_path = tempfile.mkstemp(suffix=".crt")
+    pk_fd, pk_path = tempfile.mkstemp(suffix=".key")
+    with os.fdopen(cert_fd, "w") as cert_file:
+        cert_file.write(certificate)
+    with os.fdopen(pk_fd, "w") as pk_file:
+        pk_file.write(private_key)
+    atexit.register(lambda: os.remove(cert_path))
+    atexit.register(lambda: os.remove(pk_path))
```
sabaini: I feel a bit uneasy about this one. atexit() is not run when the program crashes or is killed, and this might leave sensitive key material longer around than intended. Using a context manager resp. try / finally would be safer (the program can still crash but the time window is shorter)

microceph-orch/src/microceph/client/client.py
```diff
@@ -0,0 +1,116 @@
...
+        return cls("http+unix://" + escaped_socket_path)
+
+    @classmethod
+    def from_http(
+        cls,
+        endpoint: str,
+        certificate_authority: str | None = None,
+        certificate: str | None = None,
+        private_key: str | None = None,
+    ) -> "Client":
```
sabaini: Do we actually use Client.from_http() anywhere? Asking because if we don't need it we could bypass the pesky ssl key handling altogether, cf. above

microceph-orch/src/microceph/client/client.py
```diff
@@ -0,0 +1,116 @@
...
+        """Return a client initiliazed to the clusterd http endpoint.
+
+        If both certificate and private_key are provided, the client will
+        use them to authenticate to the server.
```
sabaini: The comment seems to imply that ssl certs are optional but from the usage above they seem to be mandatory if http is being used?

microceph-orch/src/microceph/module.py
```diff
@@ -0,0 +1,277 @@
...
+        return True, "", {}
+
+    def notify(self, notify_type: NotifyType, notify_id: str) -> None:
+        """
+
+        :param notify_type:
+        :param notify_id:
+        :return:
+        """
+        logger.info(f"notify called with notify_type: {notify_type} and notify_id: {notify_id}")
```
sabaini: Would it make sense to log the fact we're not actually doing anything on notify? 

Sth like
```
logger.info(f"noop: notify called with notify_type: {notify_type} and notify_id: {notify_id}")
```

microceph-orch/src/microceph/module.py
```diff
@@ -0,0 +1,277 @@
...
+    DaemonDescription,
+    CLICommandMeta,
+    handle_orch_error,
+    OrchResult,
+)
+
+from .client.client import Client
+from .client.service import RemoteException
+
+logger = logging.getLogger(__name__)
```
sabaini, line 39: nit: in previous modules we were using `LOG = logging.getLogger(__name__)` 
Think both styles are ok but would opt for consistent usage

microceph-orch/src/microceph/client/client.py
```diff
@@ -0,0 +1,116 @@
...
+                )
+            self._certs = to_file_path_certs(certificate, private_key)
+            self._session.mount(
+                "https://",
+                MTLSAdapter(certificate_authority=certificate_authority),
+            )
+
+        self.cluster = MicroClusterService(self._session, self._endpoint, self._certs)
+        self.status = StatusService(self._session, self._endpoint, self._certs)
+        self.services = ExtendedAPIService(self._session, self._endpoint, self._certs)
```
sabaini, line 39: Suggest to add a debug logging call here to say we created a client and log the endpoint

sabaini (changes requested):
Hey @UtkarshBhatthere thanks for this :pray: 
Some commentary inline

microceph-orch/src/microceph/module.py
UtkarshBhatthere: ack

microceph-orch/src/microceph/module.py
UtkarshBhatthere: I see this in other ceph mgr modules (cephadm in this case).
image

microceph-orch/src/microceph/client/client.py
UtkarshBhatthere: ack

microceph-orch/src/microceph/client/client.py
UtkarshBhatthere: removed from http and certificate dance.

microceph-orch/src/microceph/client/client.py
UtkarshBhatthere: removed from_http

microceph-orch/src/microceph/client/client.py
UtkarshBhatthere: removed from_http

microceph-orch/src/microceph/module.py
sabaini: Works for me, can we use `logger = logging.getLogger(__name__)` then in microceph-orch/src/microceph/client/cluster.py etc. too for consistency?

sabaini: Works for me, can we use `logger = logging.getLogger(__name__)` then in microceph-orch/src/microceph/client/cluster.py etc. too for consistency?

sabaini (changes requested):
Hey, thanks for this, one nit inline.

Do you think it would make sense to have some unit tests here too?

microceph-orch/src/microceph/module.py
UtkarshBhatthere: oh, you meant the client modules (i assumed the fellow ceph mgr modules)

sabaini (approved):
lgtm, cheers!

Issue #507, opened 2025-02-09
`microceph disk add` can run mkfs against unavailable and in-use disks


nobuto-m:
microceph  19.2.0+snap9ecc402e1e  1280  squid/edge  canonical✓  -

microceph allows users to run `disk add` action and start running mkfs **without** --wipe option. A typo in a command like mixing up PCI IDs (/dev/disk/by-path/pci-0000:04:00.0-nvme-2 vs /dev/disk/by-path/pci-0000:05:00.0-nvme-3) could lead to a disaster although mkfs actually failed luckily.

How to reproduce:

```
uvt-simplestreams-libvirt sync release=noble arch=amd64
uvt-kvm create --cpu 4 --memory 4096 foo release=noble

uvt-kvm wait foo
uv-kvm ssh foo
```

```
$ lsblk 
NAME    MAJ:MIN RM   SIZE RO TYPE MOUNTPOINTS
loop0     7:0    0  44.4M  1 loop /snap/snapd/23545
loop1     7:1    0  66.2M  1 loop /snap/core24/716
loop2     7:2    0 111.5M  1 loop /snap/microceph/1280
vda     253:0    0     8G  0 disk 
├─vda1  253:1    0     7G  0 part /
├─vda14 253:14   0     4M  0 part 
├─vda15 253:15   0   106M  0 part /boot/efi
└─vda16 259:0    0   913M  0 part /boot
vdb     253:16   0   372K  0 disk 
```

```
sudo snap install microceph --channel squid/edge

sudo microceph cluster bootstrap

$ sudo microceph disk list --json
{"ConfiguredDisks":[],"AvailableDisks":[]}
```

As above, there is no available disks. However, the disk add commands allows to start formatting `/dev/vda`, which is the root volume with the OS without requesting `--wipe` option.

```
$ sudo microceph disk add /dev/vda 

+----------+---------+
|   PATH   | STATUS  |
+----------+---------+
| /dev/vda | Failure |
+----------+---------+
Error: failed to bootstrap OSD: Failed to run: ceph-osd --mkfs --no-mon-config -i 1: exit status 250 (2025-02-09T13:58:31.743+0000 7a0636cce600 -1 bluestore(/var/lib/ceph/osd/ceph-1/block) _read_bdev_label unable to decode label /var/lib/ceph/osd/ceph-1/block at offset 4096: End of buffer [buffer:2]
[...]

PR #588, opened 2025-07-23
fix: add checks before adding OSD, WAL, or DB devices

sabaini:
# Description

Add checks for mounted devices and partitions. Also add/update some unit tests

Fixes: #507

## Type of change

## How has this been tested?

Unit tests 

## Contributor checklist

Please check that you have:

### DIFF
--- a/microceph/ceph/osd.go
+++ b/microceph/ceph/osd.go
@@ -18,6 +18,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/canonical/microceph/microceph/common"
 	"github.com/canonical/microceph/microceph/constants"
 	"github.com/canonical/microceph/microceph/interfaces"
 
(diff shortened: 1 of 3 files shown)


microceph/ceph/osd.go
```diff
@@ -463,6 +520,17 @@ func (m *OSDManager) bootstrapOSD(osdDataPath string, nr int64, wal, db *types.D
...
 
+		// Check for partitions on WAL device unless wipe is enabled
+		if !wal.Wipe {
+			hasPartitions, err := m.checkDeviceHasPartitions(storage, wal.Path)
+			if err != nil {
+				return fmt.Errorf("failed to check partitions on WAL device %s: %w", wal.Path, err)
+			}
+			if hasPartitions {
+				return fmt.Errorf("WAL device %s has partitions - use --wipe to override", wal.Path)
+			}
```
lmlg: Why not make this a function that takes a `types.DiskParameter` instead of doing this for the data, WAL and DB devices separatedly? Something like the following:

```
func checkPartitions (disk *types.DiskParameter, m *OSDManager, storage *api.ResourcesStorage) (bool, error) {
  if disk.Wipe {
    hasPartitions, err := m.checkDeviceHasPartitions(storage, disk.Path)
    if err != nil {
      return fmt.Errorf("failed to check partitions on device %s: %w", disk.Path, err)
    } else if hasPartitions {
      return fmt.Errorf("device %s has partitions", disk.Path)
    }
  }
}
```

sabaini: Ack, makes sense

lmlg (approved):

PR #590, opened 2025-07-24
fix: check if disks are pristine before attempting to use them

sabaini:
# Description

Check for pristineness before attempting to use a block device. We read the first 2kb off a block device and deem the disk pristine if it contains all zeroes.

## Type of change

## How has this been tested?

Unit tests

## Contributor checklist

Please check that you have:


microceph/common/storage.go
```diff
@@ -19,33 +19,33 @@ func IsMounted(device string) (bool, error) {
...
+		// Check if it's an exit status error (device not mounted)
+		// Handle both direct exec.ExitError and wrapped errors
+		if exitError, ok := err.(*exec.ExitError); ok {
+			if exitError.ExitCode() == 1 {
+				// Exit code 1 means device not found/not mounted
+				return false, nil
+			}
+		} else if strings.Contains(err.Error(), "exit status 1") {
+			// Handle wrapped exit status 1 from shared.RunCommand
+			return false, nil
```
lmlg: can this actually happen? That is, the path where the return code is 1 is already being handled above. Unless the program lies with this message but its exit code is not 1, then I don't see how this path can be taken.

sabaini: You're not wrong -- just uploaded a change

lmlg (approved):
lgtm!

PR #593, opened 2025-07-29
ci: Add workflow for tagging stable commit

skoech:
# Description

When a MicroCeph snap revision is promoted to stable, the corresponding Git commit is tagged manually.This PR adds a GitHub Actions workflow to help us automate updating Git tags (which we also use for marking our doc versions - and hopefully for release notes using the GitHub release feature). The workflow uses the `github-script` action, and will:
1. Install the MicroCeph snap,
2. Parse the snap info output,
3. Get the latest stable channel,
4. Find its commit ID,
5. Go to the repo and find that commit hash,
6. and tagging it.

 The workflow will run on a regular cadence (twice a week).

## Type of change

## How has this been tested?

This workflow hasn't been tested. I am hoping to test it by triggering it manually first, before enabling the `schedule` event and cron expressions.

## Contributor checklist

Please check that you have:

### DIFF
--- a/.github/workflows/tag-stable-commit.yml
+++ b/.github/workflows/tag-stable-commit.yml
@@ -5,31 +5,39 @@ name: Tag stable MicroCeph commit
 # Controls when the action will run. Workflow runs when there is a new stable channel
 # promoted on Snapcraft
 on:
-# Allows you to run this workflow manually from the Actions tab
-  workflow_dispatch: null
-#  schedule:
-#    - cron: '0 0 * * MON,THU' # Runs biweekly on Tuesdays and Thursdays
+  pull_request:
+    types: [opened, synchronize, reopened]
+    branches:
+      - main
+  # Allow manual trigger from Actions UI
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * MON,THU' # Runs on Tuesdays and Thursdays at midnight UTC
+
 jobs:
   tag-stable-commit:
-   # The type of runner that the job will run on
+    # The type of runner that the job will run on
     runs-on: ubuntu-latest
     permissions:
       contents: write # Needed for creating tags
     steps:
-     - name: Checkout repository
-       uses: actions/checkout@v4
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
       # Install the MicroCeph snap
-     - name: Install MicroCeph snap
-       run: |
-        sudo snap install microceph
-    # Find the first name under "channels:" that includes "/stable:"
-    # and parse <code name>, <release version> and <commit ID>
-     - name: Extract channel information
-       id: snap
-       uses: actions/github-script@v7
-       with:
-         script: |
-           const { execSync } = require('child_process');
+      - name: Install MicroCeph snap
+        run: |
+          sudo snap install microceph
+
+      # Find the first name under "channels:" that includes "/stable:"
+      # and parse <code name>, <release version> and <commit ID>
+      - name: Extract channel information
+        id: snap
+        uses: actions/github-script@v7
+        with: 
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const { execSync } = require('child_process');
             // Run `snap info microceph`
             const info = execSync('snap info microceph', { encoding: 'utf-8' });
             const lines = info.split('\n');
@@ -64,17 +72,17 @@ jobs:
             // Add minimum length requirement to validate output variables          
             if (!codeName || codeName.length < minCodeNameLength) {
               core.setFailed(`Invalid codeName: "${codeName}"`);
-                return;
-                }        
+              return;
+            }        
             if (!version || version.length < minVersionLength) {
               core.setFailed(`Invalid version: "${version}"`);
-                return;
-                }
+              return;
+            }
             if (!commit || commit.length < minCommitLength) {
               core.setFailed(`Invalid commit: "${commit}"`);
-                return;
-                }
-          
+              return;
+            }
+
             core.setOutput('codeName', codeName);
             core.setOutput('version', version);
             core.setOutput('commit', commit);
@@ -83,9 +91,11 @@ jobs:
             core.info(`commit=${commit}`);
 
       # Verify commit exists in the repo and print commit message first line
-     - name: Verify commit exists
-       uses: actions/github-script@v7
-       with: 
+      - name: Verify commit exists
+        id: verify
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
           script: |
             const target = '${{ steps.snap.outputs.commit }}'.slice(0, 7);  // 7-char prefix
             core.info(`Looking for a commit starting with "${target}" ...`);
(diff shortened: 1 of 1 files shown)


.github/workflows/tag-stable-commit.yml
```diff
@@ -0,0 +1,146 @@
...
+
+name: Tag stable MicroCeph commit
+
+# Controls when the action will run. Workflow runs when there is a new stable channel
+# promoted on Snapcraft
+on:
+# Allows you to run this workflow manually from the Actions tab
+  workflow_dispatch: null
+#  schedule:
+#    - cron: '0 0 * * MON,THU' # Runs biweekly on Tuesdays and Thursdays
```
sabaini: Eventually we will want to enable the cron line

.github/workflows/tag-stable-commit.yml
```diff
@@ -0,0 +1,146 @@
...
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+# This workflow contains a few jobs; the first one is called "tag-stable-commit"
+jobs:
+  tag-stable-commit:
+   # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read # Needed for commit search via the API
+  # Steps represent a sequence of tasks that will be executed as part of the job
```
sabaini: Aestethic/readability nitpickery: in general it would be good to line up comments similarly to the workflow statements (here as well as below)

.github/workflows/tag-stable-commit.yml
```diff
@@ -0,0 +1,146 @@
...
+# ------------------------------------------------------------
+      # 1. Install the MicroCeph snap
+# ------------------------------------------------------------
+     - name: Install MicroCeph snap
+       run: |
+        sudo snap install microceph
+# -----------------------------------------------------------
+    # 2. Find the first name under “channels:”
+    # 3. and then parse <release version> and <commit ID>
+# -----------------------------------------------------------
```
sabaini: Is this comment accurate? Its not looking for the first but the first line that contains stable right?

.github/workflows/tag-stable-commit.yml
```diff
@@ -0,0 +1,146 @@
...
+  
+     - name: Extract channel information
+       id: snap
+       uses: actions/github-script@v7
+       with:
+         script: |
+           const { execSync } = require('child_process');
```
sabaini: Indent needs one more whitespace :-)

.github/workflows/tag-stable-commit.yml
```diff
@@ -0,0 +1,146 @@
...
+            core.info(`Channel line: "${channelRaw}"`);
+            // Parse version, and commit ID
+            const m = channelRaw.match(
+              /^\s*([^/]+)\/stable:\s+([^+\s]+)\+snap([a-f0-9]+)\s/
+            );
+            if (!m) {
+              core.setFailed('Failed to parse channel line');
+              return;
+            }
+            const [, flavour, version, commit] = m;
```
sabaini: Naming nit: not sure about `flavour`, maybe `channel`?

.github/workflows/tag-stable-commit.yml
```diff
@@ -0,0 +1,146 @@
...
+            core.setOutput('flavour', flavour);
+            core.setOutput('version', version);
+            core.setOutput('commit', commit);
+            core.info(`flavour=${flavour}`);
+            core.info(`version=${version}`);
+            core.info(`commit=${commit}`);
```
sabaini, line 91: For safety, would suggest some quick sanity checks for output vars, e.g. minimum length requirements

.github/workflows/tag-stable-commit.yml
```diff
@@ -0,0 +1,146 @@
...
+            const channelRaw = lines.slice(headerIdx + 1)
+                                    .find(l => l.includes('/stable:'));
+            if (!channelRaw) {
+              core.setFailed('Could not find a "/stable:" channel line');
+              return;
+            }
+            core.info(`Channel line: "${channelRaw}"`);
+            // Parse version, and commit ID
+            const m = channelRaw.match(
+              /^\s*([^/]+)\/stable:\s+([^+\s]+)\+snap([a-f0-9]+)\s/
```
sabaini: Suggest to tighten RE a bit:
- channel name  `[a-z]+`
- version: `[0-9.]+`

.github/workflows/tag-stable-commit.yml
```diff
@@ -0,0 +1,146 @@
...
+            core.info(`flavour=${flavour}`);
+            core.info(`version=${version}`);
+            core.info(`commit=${commit}`);
+      # ------------------------------------------------------------------------
+      # 4. Verify commit exists in the repo and print commit message first line
+      # ------------------------------------------------------------------------
+     - name: Verify commit exists
+       uses: actions/github-script@v7
+       with: 
+           script: |
```
sabaini: Indent one less whitespace

.github/workflows/tag-stable-commit.yml
```diff
@@ -0,0 +1,146 @@
...
+            const target = '${{ steps.snap.outputs.commit }}'.slice(0, 7);  // 7-char prefix
+            core.info(`Looking for a commit starting with "${target}" …`);
+            const commits = await github.paginate(
+              github.rest.repos.listCommits,
+              {
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                per_page: 100
```
sabaini, line 107: Hm, does this mean we're only looking at the first 100 commits? May want to inc. this

.github/workflows/tag-stable-commit.yml
```diff
@@ -0,0 +1,146 @@
...
+              }
+            );
+            const hit = commits.find(c => c.sha.startsWith(target));
+            if (hit) {
+              core.info(`✅ Found commit: ${hit.sha} — ${hit.html_url}`);
```
sabaini: Personally I try to keep source code ASCII just for simplicity

.github/workflows/tag-stable-commit.yml
```diff
@@ -0,0 +1,146 @@
...
+       with:
+         script: |
+           const target = '${{ steps.snap.outputs.commit }}'.slice(0, 7);
+           core.info(`Looking for a commit starting with "${target}" …`);
+           const commits = await github.paginate(
+           github.rest.repos.listCommits,
+           {
+            owner: context.repo.owner,
+            repo: context.repo.repo,
+            per_page: 100
```
sabaini: Same q as above, may want to incr. this limit

.github/workflows/tag-stable-commit.yml
UtkarshBhatthere: why not tag every push to release branch with the snap artefact they be building ?

.github/workflows/tag-stable-commit.yml
skoech: No, it looks through all the commits; the `per_page` variable means that the commits found are being grouped into buckets of 100 to increase readability (part of the GitHub API used in pagination - the number of results returned per page is 100 max).

.github/workflows/tag-stable-commit.yml
skoech: Haha, I'll get rid of the emojis. :D

.github/workflows/tag-stable-commit.yml
skoech: Left a comment.

.github/workflows/tag-stable-commit.yml
skoech: > Eventually we will want to enable the cron line

Yes, I only disabled it for testing purposes.

skoech: > why not tag every push to release branch with the snap artefact they be building ?

I don't quite understand what you mean by this, @UtkarshBhatthere . Could you elaborate some more? :)
The main aim of this job is only to tag the stable commit, how would tagging every push to the release branch work?

.github/workflows/tag-stable-commit.yml
AnneCYH: "flavour" here is grabbing the unique "name" assigned to the release; which is not technically the full "channel" name. :)

skoech: @AnneCYH, you're right. It's taking the code name, i.e. Squid.

.github/workflows/tag-stable-commit.yml
skoech: Right, first line that includes `/stable:`, thanks!

.github/workflows/tag-stable-commit.yml
```diff
@@ -0,0 +1,175 @@
...
+# promoted on Snapcraft
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    branches:
+      - main
+  # Allow manual trigger from Actions UI
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 * * MON,THU' # Runs on Tuesdays and Thursdays at midnight UTC
```
sabaini, line 15: Comment not accurate

.github/workflows/tag-stable-commit.yml
```diff
@@ -0,0 +1,175 @@
...
+    permissions:
+      contents: write # Needed for creating tags
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      # Install the MicroCeph snap
+      - name: Install MicroCeph snap
+        run: |
+          sudo snap install microceph
```
sabaini, line 30: For debugging purposes I'd suggest to run `snap info microceph` right here

.github/workflows/tag-stable-commit.yml
```diff
@@ -0,0 +1,175 @@
...
+            // Run `snap info microceph`
+            const info = execSync('snap info microceph', { encoding: 'utf-8' });
+            const lines = info.split('\n');
+            // Find the "channels:" header (ignoring indentation)
+            const headerIdx = lines.findIndex(l => l.trim().startsWith('channels:'));
+            if (headerIdx === -1) {
+              core.setFailed('Could not find "channels:" in snap info output');
+              return;
+            }
+            // Get the first non-empty channel line after the header
```
sabaini, line 50: Comment not accurate

.github/workflows/tag-stable-commit.yml
```diff
@@ -0,0 +1,175 @@
...
+            const channelRaw = lines.slice(headerIdx + 1)
+                                    .find(l => l.includes('/stable:'));
```
sabaini, line 52: This finds the first line that has "/stable" in it. Can we assume that this is the correct stable release though, can we rely on the ordering? 

When I run snap info microceph I get 

```
...
channels:
  squid/stable:     19.2.0+snapab139d4a1f 2025-06-26 (1393) 117MB -
  squid/candidate:  19.2.0+snap4910cfbd93 2025-07-15 (1407) 117MB -
  squid/beta:       ↑                                             
  squid/edge:       19.2.1+snap54a98b3cec 2025-08-07 (1464) 118MB -
  latest/stable:    18.2.4+snapc9f2b08f92 2024-09-19 (1139) 102MB -
  latest/candidate: ↑                                             
  latest/beta:      ↑                                             
  latest/edge:      19.2.1+snap54a98b3cec 2025-08-07 (1464) 118MB -
  reef/stable:      18.2.4+snapc9f2b08f92 2024-09-02 (1139) 102MB -
  reef/candidate:   18.2.4+snapa97ae91192 2024-11-12 (1234) 102MB -
  reef/beta:        ↑                                             
  reef/edge:        ↑                                             
  quincy/stable:    0+git.4a608fc         2024-01-10  (793)  96MB -
...
```

So here the first line with squid/stable is indeed correct. Can we rely on that ordering though?

The other issue I see here is that we're only handling the /stable releases in the main branch (ie. the latest major release). It would be great if we could also auto-tag other stable releases -- at the moment this would be squid, reef, quincy. Those would correspond to the branches main, reef and quincy respectively.  

Maybe the workflow could loop through all the foo/stable channels, exclude the one named "latest/stable" and look to tag in the main and named branches.

.github/workflows/tag-stable-commit.yml
```diff
@@ -0,0 +1,175 @@
...
+            core.info(`Channel line: "${channelRaw}"`);
+            // Parse version, and commit ID
+            const m = channelRaw.match(
+              /^\s*([a-z]+)\/stable:\s+([0-9.]+)\+snap([a-f0-9]+)\s/
+            );
+            if (!m) {
+              core.setFailed('Failed to parse channel line');
+              return;
+            }
+            const [, codeName, version, commit] = m;
```
UtkarshBhatthere, line 66: is codename here referring to the release channel name ? (like `squid`, or `quincy` etc ?).

.github/workflows/tag-stable-commit.yml
UtkarshBhatthere: 😢

.github/workflows/tag-stable-commit.yml
UtkarshBhatthere: I think we can keep the "tag all pushes" thing to another PR. For documentation purposes, tagging the stable release commit alone is sufficient.

.github/workflows/tag-stable-commit.yml
```diff
@@ -0,0 +1,175 @@
...
+      # If it already exists but points to the correct commit, do nothing     
+      - name: Create or update stable tag for verified commit
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const commitSha = '${{ steps.verify.outputs.full_sha }}';
+            const codeName = '${{ steps.snap.outputs.codeName }}';
+            const version = '${{ steps.snap.outputs.version }}';
+            const stableTag = `v${version}+${codeName}`;
```
UtkarshBhatthere, line 133: So from what I've read above, if you use `v${Version}+${codeName}` as the tag, an example tag could look like: 
-> `v19.2.0+squid` right ?

This has a slight problem that the tag will change everytime the underlying ceph point release changes. for example: in future the tag could become:
-> `v19.2.2+squid`. 

Now since the tag is changing based on ceph-release, one of us would have to manually go and pull a new version on the RTD dashboard based on the new tag.

I propose we do not use the full semantic version but just the constant bit (something like `v19+squid`) as it will stay constant throughout the squid release life. And every time this workflow re-tags a commit, the documentation will update automatically.

PR #604, opened 2025-08-11
fix: implement logging with slog and persistent config

### LINKED ISSUE #239: microceph spams logs with debug messages
I'm deploying LXD+microceph clusters on top of Raspberry PI instances.
I found that microceph spams **syslog** with debug messages in addition to tons of debug messages in  **/var/snap/microceph/common/logs** as described in issue https://github.com/canonical/microceph/issues/163.
How to make microceph's daemons log level configurable?

sabaini:
# Description

This commit introduces a new logging system based on slog. It also separates backend from cli logging.

- Add new logger package with slog-based implementation
- Legacy level compatibility for API backwards compatibility
- Use new logger instead of legacy lxd-based logger
- Add CLI logger package for command-line utilities with debug/verbose flags
- Add tests

Fixes #239 

## Type of change

## How has this been tested?

Existing and new unit tests

## Contributor checklist

Please check that you have:


microceph/api/ops_maintenance.go
```diff
@@ -26,7 +27,7 @@ type maintenanceResponse struct {
 
 // Render renders a response for /ops/maintenance/{node} endpoint.
 func (r *maintenanceResponse) Render(w http.ResponseWriter, req *http.Request) (err error) {
-	debugLogger := logger.Logger(logger.Log)
+	debugLogger := lxdlogger.Logger(lxdlogger.Log)
```
UtkarshBhatthere: is it not possible to migrate this to the new logger ?

microceph/client/client_configs.go
```diff
@@ -83,15 +83,15 @@ func SendUpdateClientConfRequestToClusterMembers(ctx context.Context, s interfac
 	// Get a collection of clients to every other cluster member, with the notification user-agent set.
 	cluster, err := s.ClusterState().Cluster(false)
 	if err != nil {
-		logger.Errorf("failed to get a client for every cluster member: %v", err)
+		clilogger.Errorf("failed to get a client for every cluster member: %v", err)
```
UtkarshBhatthere, line 86: is there a reason why the same logger could not be used ? so as to have a cliLogger.

UtkarshBhatthere:
@sabaini i did a quick review, will do another one today.

microceph/api/ops_maintenance.go
sabaini: No, this expects the interface of the lxd logger

microceph/client/client_configs.go
sabaini: Reason is that the clilogger logs to the terminal and has different logging levels, also it should not be touched by the api log level -- this is why I made it a separate package

microceph/api/ops_maintenance.go
UtkarshBhatthere: ack.

UtkarshBhatthere: Is this case should we create an issue for "eventually" migrating these instances from older lxd interface to the new logger in future ?

microceph/client/client_configs.go
UtkarshBhatthere: fair and square. Sounds good.

lmlg (approved):
lgtm, good stuff

PR #616, opened 2025-09-03
fix: use ceph-bluestore-tool for wiping

### LINKED ISSUE #617: OSD wipe unreliable
For wiping OSDs we currently use dd to zero out 40M from the target bdev. However ceph-osd when running with the --mkfs switch may place metadata information across the block device, not just in the first blocks. A partially wiped bdev can confuse the OSD bootstrap process though, thinking it actually has a bootstrapped device present

This seems to be more prevalent in the 19.2.1 point release where we see that an incompleteley wiped device seems to succeed in bootstrapping, but the subsequent ceph-osd start then failing with a message `missing 'type' file and unable to infer osd type`

sabaini:
Necessary as plain dd did not reliably wipe the bdev

Fixes #617

### DIFF
--- a/microceph/ceph/osd.go
+++ b/microceph/ceph/osd.go
@@ -1128,9 +1128,9 @@ func (m *OSDManager) timeoutWipe(path string) error {
 
 	_, err := m.runner.RunCommandContext(
 		ctx,
-		"dd", "if=/dev/zero",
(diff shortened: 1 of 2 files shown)


lmlg (approved):
lgtm
(scary that dd isn't enough, though, I wonder how many wrongly-wiped disks there are out there)

PR #621, opened 2025-09-10
fix: improve pristine disk check with ceph-bluestore-tool validation

### LINKED ISSUE #618: Pristine disk check insufficient
We currently look at the first 2k of a block device to check if it's pristine. With upstream changes in 19.2.1 this is not sufficient anymore, as left-over labels deep in the disk can trip up the OSD bootstrap process.

We need to improve the pristine disk check to be more strict about this while keeping it light weight so that we don't incur a huge i/o load for checking pristineness

sabaini:
# Description

Improve the heuristics on what we consider a "pristine" disk. We used to just check if the first 2k of a blockdev were zeroes, but with recent upstream that isn't enough anymore. This change introduces more checks so we're more strict on what we accept as pristine: checking for zero sectors in various strategic locations (begin and end of disk, gpt tables, 64k and 1M) and check if ceph-bluestore-tool returns any ceph labels

If it fixes a reported issue, mention it here, i.e.:
Fixes #618 

## Type of change

## How has this been tested?

Add unit and functional test

## Contributor checklist

Please check that you have:


microceph/common/storage.go
```diff
@@ -94,41 +94,133 @@ func IsCephDeviceWithFs(device string, fs afero.Fs) (bool, error) {
...
+
+	// First, quick check of some likely locations for non-zero data
+	logger.Debugf("Perform zero-byte check on device: %s", devicePath)
+	pristineByZeroCheck, err := checkForZeros(devicePath, fs)
+	if err != nil {
+		logger.Errorf("Zero-byte check failed for device %s: %v", devicePath, err)
+		return false, err
+	}
+	if !pristineByZeroCheck {
+		logger.Infof("Device %s failed zero-byte check - contains non-zero data at strategic locations", devicePath)
```
UtkarshBhatthere, line 115: nit: log level should be warn or err.

microceph/common/storage.go
```diff
@@ -94,41 +94,133 @@ func IsCephDeviceWithFs(device string, fs afero.Fs) (bool, error) {
...
+		return false, nil
+	}
+	logger.Debugf("Device %s passed zero-byte check - all strategic locations contain zeros", devicePath)
```
UtkarshBhatthere: This log print is redundant. As it is followed just after by another log (of same level) and all previous branches have their own logs.

microceph/common/storage.go
```diff
@@ -94,41 +94,133 @@ func IsCephDeviceWithFs(device string, fs afero.Fs) (bool, error) {
...
+		// If ceph-bluestore-tool fails for any reason, treat as pristine for robustness
+		logger.Infof("ceph-bluestore-tool check failed for %s, treating as pristine for robustness: %v", devicePath, err)
+		return true, nil
+	}
 
+	if pristineByCephTool {
+		logger.Infof("Device %s is pristine - passed both zero-byte and ceph-bluestore-tool checks", devicePath)
+	} else {
+		logger.Infof("Device %s is not pristine - ceph-bluestore-tool found existing labels", devicePath)
+	}
```
UtkarshBhatthere: Super Duper NITPICKING: I think a better way is as a single log telling whether device is pristine or not. Since individual causes (whether nonPristine by zeroes or ceph-bluestore-tool) already have their own log prints.

```
logger.Info("Device %s is pristine: %v", deviceId, isPristine)
```

microceph/common/storage.go
```diff
@@ -94,41 +94,133 @@ func IsCephDeviceWithFs(device string, fs afero.Fs) (bool, error) {
...
 	}
+	deviceSize := stat.Size()
+	logger.Debugf("Device %s size: %d bytes", devicePath, deviceSize)
 
-	if readBytes != wantBytes {
-		logger.Warnf("short read from %s: got %d bytes, expected %d", devicePath, readBytes, wantBytes)
-		return false, nil
+	// Define locations to check (in bytes from start)
+	checkPoints := []int64{
+		0,       // Beginning of disk (MBR, GPT primary)
```
UtkarshBhatthere, line 153: Nice!

microceph/common/storage.go
sabaini: Unsure about that tbh. 

ERROR should be for cases where something unexpected happened that impacted a component such that it can't function properly. WARNING is for cases where the system is not impacted but the operator should be warned of some adverse condition that needs acting upon.

Users trying to (re)using disks (and those disks being rejected w/o wiping) doesn't really impact the system

microceph/common/storage.go
sabaini: Fair

microceph/common/storage.go
UtkarshBhatthere: hmm, fair. The operator gets an error for the disk add command anyways so their is clear information about failures. For system, having this log at info level seems alright.

PR #628, opened 2025-09-23
feat: Adds support for enable/disable CephFS replication

UtkarshBhatthere:
# Description

Please include a summary of the changes and the related issue. Also include relevant
motivation and context. List any dependencies that are required for this change.

If it fixes a reported issue, mention it here, i.e.:
Fixes #<issue-number>

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How has this been tested?

> [!NOTE]

## Contributor checklist

Please check that you have:


microceph/ceph/cephfs_snapshot.go
```diff
@@ -0,0 +1,140 @@
...
+		"peer_list", volume, "--format=json",
+	})...)
+	if err != nil {
+		err = fmt.Errorf("failed to get CephFS snapshot mirror list: %w", err)
+		logger.Error(err.Error())
+		return false, err
+	}
+
+	peers := map[string]CephFSSnapshotMirrorPeer{}
+	err = json.Unmarshal([]byte(output), &peers)
```
sabaini, line 101: Hm, stupid question but is the output from above not a list instead of map?

microceph/ceph/cephfs_snapshot.go
```diff
@@ -0,0 +1,140 @@
...
+
+func cephFSSnapshotMirrorCmd(args []string) []string {
+	cmd := []string{
+		"fs", "snapshot", "mirror",
+	}
+
+	return append(cmd, args...)
+}
+
+func dropEscapedChars(input string, dropStr string) string {
```
sabaini: Still needed?

microceph/ceph/cephfs_volume.go
```diff
@@ -56,7 +56,25 @@ func ListCephFSVolumes() ([]Volume, error) {
 		response = append(response, Volume(volume.String()))
 	}
 
-	return response, nil
+	return parseListVolOutputJson(output)
+}
+
+func ListRemoteCephFSVolumes(remote string, local string) ([]Volume, error) {
```
sabaini, line 64: Please add a func doc comment for public functions (here and below)

microceph/ceph/replication_cephfs.go
```diff
@@ -260,6 +327,183 @@ func (rh *CephfsReplicationHandler) GetCephFSMirrorStatus(ctx context.Context) e
...
 }
 
+func verifyEnableRequestData(ctx context.Context, s interfaces.CephState, request types.CephfsReplicationRequest) error {
+	if len(request.Volume) == 0 {
+		return fmt.Errorf("missing CephFS volume name")
+	}
+
+	if len(request.RemoteName) == 0 {
+		return fmt.Errorf("missing remote cluster name")
+	} else {
```
sabaini: could lose the `else` statement as the return above will shortcut anyway

microceph/ceph/cephfs_volume.go
```diff
@@ -155,14 +190,30 @@ func GetCephFSSubvolumes(volume Volume, subvolumegroup string) ([]Subvolume, err
...
-	var args []string
-
+func GetCephFSSubvolumePath(subvolumegroup string, subvolume string) string {
+	var retval string
 	if len(subvolumegroup) != 0 {
-		args = []string{"fs", "subvolume", "getpath", volume, subvolume, subvolumegroup}
+		retval = fmt.Sprintf("/volumes/%s/%s/", subvolumegroup, subvolume)
 	} else {
-		args = []string{"fs", "subvolume", "getpath", volume, subvolume}
+		retval = fmt.Sprintf("/volumes/_nogroup/%s/", subvolume)
```
sabaini: Here and above a trailing slash is added to the path -- is that correct? Aiui Ceph by default doesn't use trailing slashes, could provoke mismatches.

microceph/ceph/cephfs_snapshot.go
```diff
@@ -0,0 +1,140 @@
...
+
+// Contains methods for interacting with the Ceph FS mirror snapshot funtionality
+
+type CephFSSnapshotMirrorPeer struct {
+	ClientName string `json:"client_name"`
+	SiteName   string `json:"site_name"`
+	Volume     string `json:"fs_name"`
+}
+
+func cephFSSnapshotMirrorEnableVolume(volume string) error {
```
sabaini, line 22: General question for all the ceph-cmd bindings here -- would it make sense to add a logger.Warn() for those we don't expect an error regularly?

microceph/ceph/cephfs_volume.go
```diff
@@ -98,6 +118,21 @@ func GetCephFSVolume(volume Volume) (CephFSVolume, error) {
 	return response, nil
 }
 
+func CephFSSubvolumeExists(volume string, subvolumegroup string, subvolume string) bool {
```
sabaini, line 124: I feel like this func needs more logger.Debugf()

microceph/ceph/replication_cephfs.go
```diff
@@ -260,6 +327,183 @@ func (rh *CephfsReplicationHandler) GetCephFSMirrorStatus(ctx context.Context) e
...
+		return err
+	}
+
+	for _, volume := range volumes {
+		if strings.Compare(string(volume), vol) == 0 {
+			return nil
+		}
+	}
+
+	return fmt.Errorf("cephfs volume %s not found on remote %s", vol, remote)
```
sabaini: Maybe also output the list of existing vols

microceph/ceph/replication_cephfs.go
```diff
@@ -260,6 +327,183 @@ func (rh *CephfsReplicationHandler) GetCephFSMirrorStatus(ctx context.Context) e
...
+}
+
+func enableCephFSVolumeMirror(ctx context.Context, volume string, remote string, local string) error {
+	peerExists, err := cephFSSnapshotMirrorPeerExists(ctx, volume, remote)
+	if err != nil {
+		// fail if volume mirroring is enabled.
+		if !strings.Contains(err.Error(), constants.VolumeNotMirrored) {
+			logger.Errorf("Failed to check if peer %s exists for CephFS volume %s: %v", remote, volume, err)
```
sabaini, line 428: Hm, not sure I understand the purpose here, maybe could clarify the comment?

sabaini (changes requested):
Hey @UtkarshBhatthere thanks for this patch, quite a piece of work. Some questions / requests inline

microceph/ceph/cephfs_snapshot.go
UtkarshBhatthere: no :D, dropped.

microceph/ceph/cephfs_volume.go
UtkarshBhatthere: ack

microceph/ceph/replication_cephfs.go
UtkarshBhatthere: ack

microceph/ceph/cephfs_volume.go
UtkarshBhatthere: ack, fixed and simplified the logic.

microceph/ceph/cephfs_snapshot.go
UtkarshBhatthere: The typical cmd functions are supposed to be thin wrappers while the responsibility of error handling is for the caller. The functions which go beyond this definition (in the same file) all have error checks, error logs and a debug log of the output.

microceph/ceph/cephfs_volume.go
UtkarshBhatthere: ack, added.

microceph/ceph/replication_cephfs.go
UtkarshBhatthere: ack, added.

microceph/ceph/replication_cephfs.go
UtkarshBhatthere: ack.

microceph/ceph/cephfs_snapshot.go
UtkarshBhatthere: oh not at all, i also expected the output of a list command to be a list :D

but rather the output is of the form:
`{"767bd3d3-80ca-4582-a1aa-5bb444009103": {"client_name": "client.fsmir-vol-secondary", "site_name": "primary", "fs_name": "vol"}}`

Which is more of a dict with peer uuid as the key and a dict of details as value.

microceph/cmd/microceph/replication_disable_cephfs.go
```diff
@@ -0,0 +1,89 @@
...
+	cmd.Flags().BoolVar(&c.flagForce, "force", false, "forcefully disable replication for resource")
+
+	cmd.MarkFlagRequired("volume")
+	cmd.MarkFlagsMutuallyExclusive("dir-path", "subvolumegroup")
+	cmd.MarkFlagsMutuallyExclusive("dir-path", "subvolume")
+
+	return cmd
+}
+
+func (c *cmdReplicationDisableCephFS) PreRun(cmd *cobra.Command, args []string) error {
```
sabaini, line 45: Think this needs to be hooked up to be actually run?

microceph/cmd/microceph/replication_disable_cephfs.go
```diff
@@ -0,0 +1,89 @@
...
+	volume         string
+	subvolume      string
+	subvolumegroup string
+	dirpath        string
+	flagForce      bool
+}
+
+func (c *cmdReplicationDisableCephFS) Command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "cephfs <resource>",
```
sabaini: Is this Use string accurate?

microceph/cmd/microceph/replication_enable_cephfs.go
```diff
@@ -0,0 +1,104 @@
...
+	remoteName     string
+	volume         string
+	dirpath        string
+	subvolume      string
+	subvolumegroup string
+}
+
+func (c *cmdReplicationEnableCephFS) Command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:     "cephfs <resource>",
```
sabaini: Is this Use string accurate?

sabaini (changes requested):
Hey hey some nits, otherwise shaping up nicely!

microceph/cmd/microceph/replication_enable_cephfs.go
UtkarshBhatthere: nice catch, all the resources are named so this is not correct.

microceph/cmd/microceph/replication_disable_cephfs.go
UtkarshBhatthere: fixed, thanks :D

microceph/cmd/microceph/replication_disable_cephfs.go
UtkarshBhatthere: aaaah, i missed the disable one. Thanks for the catch.

microceph/ceph/cephfs_volume.go
```diff
@@ -66,6 +86,8 @@ func CephFsSubvolumePathDeconstruct(path string) (subvolumegroup string, subvolu
 		return "", "", fmt.Errorf("invalid CephFS subvolume path: %s", path)
 	}
 
+	logger.Debugf("FSVOL: %+v", parts)
+
 	subvolumegroup = parts[CephFsSubVolumeGroupIndex]
 	subvolume = parts[CephFsSubVolumeIndex]
```
sabaini, line 92: I realize this is not part of the actual PR but with the new workflow we can now have /dir1/ directories paths passed in here as well right? ths would split into ["", "dir1", ""] and pass the guard but would trip up on parts[3] here

sabaini: As a follow up I think a possible way to make this more robust is to canonicalize (no pun intended) all paths e.g. strip off any trailing slashes preferably as soon as the system receives them either via api or cli

UtkarshBhatthere: So this path deconstruction is only done for subvolume paths which have a formal structure (/volumes/svgroup/sv) so unless cephfs changes the path structure for subvolumes itself (which would be a breaking change) we should be OK.

For standard filesystem directory paths we pass them as is and ceph decides whether they are valid or not (should always start with '/')

UtkarshBhatthere: the trailing slash is not a big concern for us because we split at slashes to pick out group name and subvolume name. so `/volumes/abc/def/` and `/volumes/abc/def` would both still return abc and def at same indexes.

sabaini: Hm, can't a path "/dir1/" come in from Ceph and be passed to to CephFsSubvolumePathDeconstruct(path) via GetCephFsPerVolumeListResponse()? This would result in https://go.dev/play/p/8YlSpNrMH2I

UtkarshBhatthere: I see what you mean, have added a prefix check for subvolumes path to prevent this further.

sabaini (approved):
lgtm!

PR #643, opened 2025-10-29
doc: Add HTML meta descriptions

skoech:
# Description

This PR adds HTML meta descriptions to a few pages (more to come) as part of our Traffic/SEO/analytics objectives. We need to generate description metadata for pages, and not rely on Sphinx to create it (or search engines to extract summaries from the content).
The meta description is a type of HTML metadata that influences SEO. It provides a summary of the page, and often, search engines will use it as the search result snippet for that page, if they think it gives users a more accurate description than would be possible generating one purely from the on-page content. 

The HTML files used by browsers to render our documentation are generated from non-HTML sources, i.e. rST. This means that we never directly edit the HTML source files, including the meta description and any other HTML metadata. Therefore, we cannot set the meta description directly in the raw HTML. Our [Sphinx starter pack](https://github.com/canonical/sphinx-docs-starter-pack) automatically generates the meta description for each HTML page from the first 200 visible characters of its main text, excluding the top-level heading, using the [sphinxext-opengraph extension](https://sphinxext-opengraph.readthedocs.io/en/latest/).

But we can manually set custom meta description to override the auto-generated one, using the [meta](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#html-meta)[ directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#html-meta) at the top of the file, with the description option.

## Type of change

## How has this been tested?

I've tested the changes locally.

> [!NOTE]

## Contributor checklist

Please check that you have:

### DIFF
--- a/docs/explanation/index.rst
+++ b/docs/explanation/index.rst
@@ -1,3 +1,6 @@
+..meta::
+   :description: Explanations of MicroCeph concepts and topics like, the MicroCeph charm, MicroCeph architecture, security, network configuration, remote replication, and cluster scaling.
+
 Explanation
 ===========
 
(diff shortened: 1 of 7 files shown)


skoech:
Thank you, @SecondSkoll!

PR #650, opened 2025-11-20
feat: add stripingv2 to default rbd featureset

sabaini:
# Description

Add stripingv2 to the rbd features enabled by default. Client kernels > 5.3 should be ok with this. Add some docs for older clients on how to downgrade rbd features in case of incompat.

Add tests, refactor a bit.

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How has this been tested?

Updated unit test, add functest verif.

## Contributor checklist

Please check that you have:

### DIFF
--- a/docs/how-to/mount-block-device.rst
+++ b/docs/how-to/mount-block-device.rst
@@ -177,7 +177,7 @@ RBD features and older clients
 
 MicroCeph sets the default RBD features to 63 which corresponds to ``layering + exclusive-lock + object-map + fast-diff + deep-flatten + stripingv2``. This is done to ensure better performance and features for the RBD images. However, older kernels might not support all these features.
 
-If you face issues mapping RBD images on older kernels, you can either disable the unsupported features for the specific image or change the default behaviour for the cluster.
+If you face issues mapping RBD images on older kernels, you can either disable the unsupported features for the specific image, change the default behaviour for the cluster, or for a specific pool.
 
 To disable unsupported features for a specific image:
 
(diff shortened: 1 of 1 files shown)


microceph/ceph/bootstrap.go
```diff
@@ -50,6 +50,16 @@ func BootstrapCephConfigs(cn string, pn string) error {
 		return err
 	}
 
+	// Default RBD features
+	// 63 = layering + exclusive-lock + object-map + fast-diff + deep-flatten + stripingv2
+	err = SetConfigItem(apiTypes.Config{
+		Key:   "rbd_default_features",
+		Value: "63",
```
lmlg, line 59: I'm generally OK with these changes, but I wanted to know first _how_ you reached the magic number 63, and whether it's a stable value or if we have to maintain different constants for different Ceph versions.

sabaini: Ack, added some comments

docs/how-to/mount-block-device.rst
```diff
@@ -171,3 +171,36 @@ Ceph cluster state post IO:
...
+
+If you face issues mapping RBD images on older kernels, you can either disable the unsupported features for the specific image or change the default behaviour for the cluster.
+
+To disable unsupported features for a specific image:
+
+.. code-block:: none
+
+    $ sudo rbd feature disable block_pool/bd_foo object-map fast-diff deep-flatten
+
+To change the default behaviour for the cluster, you can update the ``rbd_default_features`` config option:
```
skoech: Here (and in a few sections below), I'd suggest a bit more authoritative language, i.e. remove the "you can" to give users more confidence and direction/not feel like there are other options to consider (unless there are).

PR #668, opened 2026-01-19
feat: add DSL-based device matching for OSD selection

sabaini:
# Description

This implements the snap part of phase 1 of the CE127 spec, adding DSL-based device matching to the microceph disk add command.

New features:
- New DSL package (microceph/dsl/) with lexer, parser, and evaluator
- --osd-match flag for selecting devices using DSL expressions
- --dry-run flag to preview matched devices without adding
- Server-side DSL evaluation against available devices

Fix: use sysfs to get block device size for pristine check

## Type of change

## How has this been tested?

Added 
- unit tests
- functests

## Contributor checklist

Please check that you have:

### DIFF
--- a/microceph/dsl/eval.go
+++ b/microceph/dsl/eval.go
@@ -285,45 +285,20 @@ func compareValues(a, b Value) (int, error) {
 		return 0, nil
 	}
 
-	// If one is a number and the other is a string, try to parse the string
-	if a.Type() == ValueTypeNumber || b.Type() == ValueTypeNumber {
-		an := a.Number()
-		bn := b.Number()
-
-		// If one value is already a number, try to use the other's Number() conversion
-		// Note: StringValue.Number() returns 0, so this only works if at least one is a number
-		if a.Type() == ValueTypeNumber && b.Type() == ValueTypeString {
-			// Try to parse b as a number with unit
-			parsed, _, err := ParseNumberWithUnit(b.String())
-			if err != nil {
-				return 0, fmt.Errorf("cannot compare number with non-numeric string '%s'", b.String())
-			}
-			bn = parsed
-		} else if b.Type() == ValueTypeNumber && a.Type() == ValueTypeString {
-			// Try to parse a as a number with unit
-			parsed, _, err := ParseNumberWithUnit(a.String())
-			if err != nil {
-				return 0, fmt.Errorf("cannot compare number with non-numeric string '%s'", a.String())
-			}
-			an = parsed
-		}
-
-		if an < bn {
+	// If both are strings, compare lexicographically
+	if a.Type() == ValueTypeString && b.Type() == ValueTypeString {
+		as, bs := a.String(), b.String()
+		if as < bs {
 			return -1, nil
 		}
-		if an > bn {
+		if as > bs {
 			return 1, nil
 		}
 		return 0, nil
 	}
 
-	// Compare as strings
-	as, bs := a.String(), b.String()
-	if as < bs {
-		return -1, nil
-	}
-	if as > bs {
-		return 1, nil
-	}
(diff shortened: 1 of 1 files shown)


tests/scripts/test_dsl_functest.sh
```diff
@@ -0,0 +1,535 @@
...
+    echo "=== Running DSL Tests ==="
+
+    test_dsl_disk_list
+    test_dsl_available_disks
+    test_dsl_type_match
+    test_dsl_size_comparison
+    test_dsl_combined_conditions
+    test_dsl_no_match
+    test_dsl_invalid_expression
+    test_dsl_mutual_exclusivity
```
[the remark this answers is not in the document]
sabaini: As there's quite a bit of overhead in test setup involved I'd tend to accept the dep

microceph/dsl/eval.go
```diff
@@ -0,0 +1,357 @@
...
+}
+
+// matchWithTimeout executes a regex match with a timeout to prevent ReDoS attacks.
+func matchWithTimeout(re *regexp.Regexp, s string, timeout time.Duration) (bool, error) {
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+
+	resultCh := make(chan bool, 1)
+	go func() {
+		resultCh <- re.MatchString(s)
```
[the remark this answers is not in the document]
sabaini: We use a goroutine to address the fact Gos RE lib can't be cancelled. As we're limiting the pattern length and the RE lib doesnt have > linear time complexity this should be safe and go routines should be able to finish. Uploaded change w/ doc comments

microceph/dsl/eval.go
```diff
@@ -0,0 +1,366 @@
...
+// matchWithTimeout executes a regex match with a timeout as a safety measure.
+//
+// Note on goroutine behavior: Go's regexp package does not support cancellation,
+// so on timeout the spawned goroutine will continue until MatchString completes.
+// The buffered channel prevents the goroutine from blocking on send. This is
+// acceptable because:
+//  1. Go's RE2-based regexp guarantees linear time complexity O(n*m), preventing
+//     catastrophic backtracking that causes ReDoS in PCRE-based engines.
+//  2. Pattern length is already limited to MaxRegexPatternLength (1000 chars).
+//  3. Input strings (device attributes) are typically short (<256 chars).
```
lmlg: Given these reasons, I'm frankly not sure that we need to set timeouts, since it's rather unlikely that a regexp (malicious or otherwise) could result in a ReDoS

microceph/dsl/eval.go
```diff
@@ -0,0 +1,366 @@
...
+			an = parsed
+		}
+
+		if an < bn {
+			return -1, nil
+		}
+		if an > bn {
+			return 1, nil
+		}
+		return 0, nil
```
lmlg: Why do we need this? Couldn't we just use the fallback mechanism that converts both values to strings and compares those?

microceph/dsl/eval.go
sabaini: Good point ig... I've simplified this now, ptal? @lmlg

microceph/dsl/eval.go
sabaini: If we use the fallback and convert both to String we end up comparing lexicographically e.g. `"5GiB"` then being bigger than `"10GiB"`. Not sure there's a good use case for this. Alternatively we could just make it an error to try to compare Numbers/Strings although we might then need to introduce type casting functionality.

sabaini: FTR. as per our discussion uploading a change so that mixed type comparisons are throwing an error

PR #670, opened 2026-01-25
Clarify RGW and RADOS terminology in tutorial

anantone:
Added a note explaining RGW and RADOS terms.

Since this tutorial can take a new user to having a test installation of MicroCeph, it would help knowing what RGW stands for, as the acronym is introduced. Whoever came up with an acronym partly made of another acronym is responsible for the fact that RADOS has to be explained too, consequently.


UtkarshBhatthere:
@anantone can you please resolve the merge conflict on this PR ? the change overall lgtm.

PR #671, opened 2026-01-28
Add guide for enabling RGW with TLS certificate

### LINKED ISSUE #596: Doc: howto SSL/TLS for RGW
We currently have reference-level documentation on how to enable SSL/TLS for Ceph Rados Gateway (RGW), but it is terse and assumes prior knowledge. This makes it difficult for users to set up a secure RGW endpoint confidently.

To address this, we propose creating a dedicated How-To guide for enabling SSL/TLS for the RGW component in MicroCeph, following the [Diataxis documentation framework](https://diataxis.fr/how-to-guides/). The focus should be on practical, step-by-step instructions that users can follow with minimal assumptions, and that cover:

* Required preconditions (e.g. presence of RGW, valid certificates, MicroCeph deployment)
* Options for obtaining or generating TLS certificates (self-signed vs Lets Encrypt vs external CA)
* Configuration steps for MicroCeph to enable and verify SSL/TLS on RGW
* How to check that HTTPS is correctly set up and used
* Troubleshooting common issues (e.g. certificate mismatch, browser warnings, config key mistakes)

The guide should use concrete MicroCeph examples and commands. 

The guide should be consistent with the existing documentation here: https://canonical-microceph.readthedocs-hosted.com/stable/

anantone:
This guide details the steps to enable RGW with a TLS certificate for a MicroCeph single node installation, including prerequisites, certificate acquisition using Certbot, and configuration verification.

Fixes #596


anantone:
@skoech @sabaini Here is a proposal, which I am of course more than happy to improve!

anantone:
I've now signed the CLA.

docs/how-to/enable-rgw-with-tls.rst
```diff
@@ -0,0 +1,133 @@
...
+How to enable RGW with a TLS certificate
+========================================
+
+This guide will demonstrate how to enable RGW with a TLS certificate, in order to access a MicroCeph single node installation through a https endpoint.
+
+Prerequisites
+=============
+
+- `a MicroCeph single node installation <https://canonical-microceph.readthedocs-hosted.com/latest/tutorial/get-started/>`_. This will have RGW enabled.
+- `a valid TLS certificate`_. We will use Certbot to obtain a TLS certificate from Let's Encrypt. You can also use a self-signed certificate, or request one from an external/commercial CA.
```
sabaini, line 11: Hm, we're listing this under `Prerequisites` but at the same time say we're going to create one which seems contradictory. Personally I'd lean towards asking for a cert as a prereq. and only providing pointers on how to obtain one (LetsEncrypt or other means).

docs/how-to/enable-rgw-with-tls.rst
```diff
@@ -0,0 +1,133 @@
...
+
+    Please deploy a DNS TXT record under the name:
+
+    _acme-challenge.s3.yourdomain.com.
+
+    with the following value:
+
+    MKc2mNJmrOuZ5-6zcxnD3NUCb_0w_mRG8bOPIA8K66w
+
+Your certificate should be automatically issued and downloaded.
```
sabaini, line 56: This assumes that users have control over the DNS of the domain they want to host it on. I'm a bit worried that creating TLS certs will get out of scope for this tutorial. I'd lean towards asking for certs as a preqrequisite, provide pointers to instructions on how to get one (certbot if they can, self-signed or bought otherwise)

docs/how-to/enable-rgw-with-tls.rst
```diff
@@ -0,0 +1,133 @@
...
+Enable RGW with your certificate
+================================
+
+The ``enable`` command expects the actual base 64 certificate and key, not just the file path. This can be done with the following command, where you will need to substitute the actual path to your certificate and key.
+
+.. code::
+
+    sudo microceph enable rgw \
+    --ssl-certificate "$(base64 -w0 ./domain.crt)" \
+    --ssl-private-key "$(base64 -w0 ./domain.key)"
```
sabaini, line 84: Nit: typically certbot will create a 0600 root owned key, which means the base64 which is running with non-sudo privs here won't be able to read it

docs/how-to/enable-rgw-with-tls.rst
```diff
@@ -0,0 +1,133 @@
...
+    If you used Let's Encrypt to obtain your certificate, the paths should look like ``/etc/letsencrypt/live/s3.yourdomain.com/fullchain.pem`` for the certificate and ``/etc/letsencrypt/live/s3.yourdomain.com/privkey.pem`` for the key.
+
+If your port 443 is already in use, you can specify a different SSL port:
+
+.. code::
+
+    sudo microceph enable rgw \
+    --ssl-port 7443 \
+    --ssl-certificate "$(base64 -w0 ./domain.crt)" \
+    --ssl-private-key "$(base64 -w0 ./domain.key)"
```
sabaini, line 96: See above

sabaini (changes requested):
Hey @anantone thanks for that, left some comments/worries inline.
cc: @skoech

docs/how-to/enable-rgw-with-tls.rst
anantone: Fair point. The problem is that we don't have a good tutorial for how to get certificates, that we could link to. @skoech and myself were actually discussing writing one, but in the meantime decided to provide one possible way here, so that users of this guide can still go forward even if they don't know the whole certificate story.

docs/how-to/enable-rgw-with-tls.rst
anantone: Again this is just one option. But yes, it may be better to skip it entirely, and just link to a full ressource on the topic.

docs/how-to/enable-rgw-with-tls.rst
anantone: I'm not sure how to resolve that. This worked in my testing with a non root, sudo privileged user. Would you have a technical suggestion?

docs/how-to/enable-rgw-with-tls.rst
anantone: Noted. Same fix to apply when determined.

docs/how-to/enable-rgw-with-tls.rst
```diff
@@ -0,0 +1,133 @@
...
+Get a certificate
+-----------------
+
+In the following command, replace the placeholders with your domain and valid email address (for certificate renewals).
+
+.. code::
+
+    sudo certbot certonly --manual \
+        --preferred-challenges dns \
+        -d s3.yourdomain.com \
```
astrojuanlu, line 40: I came to this guide looking for a way to set up a _local_ TLS certificate for development purposes, so I don't have a domain name to use. Should that be listed in the Prerequisites?

PR #673, opened 2026-02-06
Feat: MicroCeph Tentacle

UtkarshBhatthere:
For the tentacle release of microceph snap we need:

1. Core26 as tentacle release will match with ubuntu 26.04 support cycle.
2. ceph packages (for stable we need these packages to come from the distro pkgs).
3. microcluster v3 (LTS for micro)


UtkarshBhatthere:
stale

PR #676, opened 2026-02-14
fix: Resolve loop device unexpected behaviour

### LINKED ISSUE #674: Encryption flag silently does nothing on loop files
When requesting a loop backed disk with `--encrypt` no encryption occurs and no error or warning is emitted. 

Loop files are intended for development, and not working with FDE is intentional, however the experience of requesting an encrypted loop file, and the result being that you get an unencrypted device is unexpected. The user may expect that FDE is available on the loop file. 

If the `--encrypt` flag is passed to `disk add`, an error should occur.

johnramsden:
# Description

Fixes #674

Loop devices can exhibit unexpected behavior where they are passed encrypt/wal/db flags but it does nothing. FDE/WAL/DB is intended for production usage, so we do not support it on loop devices, but the user experience should be better in case they are trying to test encryption on loop devices and wondering why it's not working.

Fix silent failures when adding loop-backed OSDs:
  - Error on --encrypt with loop devices instead of silently ignoring it
  - Fix missing encryptDevice error handling (err was not assigned)
  - Apply validateAddOSDArgs to single-disk bulk path which previously skipped validation
  - Consolidate loop device detection into isLoopDevice helper
  - Pass WAL/DB through to add disk so we can validate and not ignore WAL/DB
  - Amend old docs

## Type of change

## How has this been tested?

* Unit tests adjusted to test WAL/DB/encrypt

Manual test commands:

    sudo microceph cluster bootstrap

    # Should error: encryption on loop device
    sudo microceph disk add loop,4G,3 --encrypt

Validation Error found
Error: encryption is not supported on loop devices

    # Should error: loop device with WAL
    sudo microceph disk add loop,4G,3 --wal-device /dev/sdx

Validation Error found
Error: WAL/DB is not supported on loop devices

    # Should error: loop device with DB
    sudo microceph disk add loop,4G,3 --db-device /dev/sdx

Validation Error found
Error: WAL/DB is not supported on loop devices

    # Should succeed: loop device without encrypt/WAL/DB
    sudo microceph disk add loop,1G,3

```
+-----------+---------+
|   PATH    | STATUS  |
+-----------+---------+
| loop,1G,3 | Success |
+-----------+---------+
```

## Contributor checklist

Please check that you have:

### DIFF
--- a/microceph/ceph/osd.go
+++ b/microceph/ceph/osd.go
@@ -191,7 +191,7 @@ func (m *OSDManager) setupEncryptedOSD(devicePath string, osdDataPath string, os
 	}
 
 	// Encrypt the device
-	encryptDevice(devicePath, key)
+	err = encryptDevice(devicePath, key)
 	if err != nil {
 		return "", fmt.Errorf("failed to encrypt: %w", err)
 	}
@@ -678,6 +678,11 @@ func (m *OSDManager) addBulkDisks(ctx context.Context, disks []types.DiskParamet
 	ret := types.DiskAddResponse{}
 
 	if len(disks) == 1 {
+		// Validate
+		err := m.validateAddOSDArgs(disks[0], wal, db)
+		if err != nil {
+			return prepareValidationFailureResp(disks, err)
+		}
 		// Add single disk with requested WAL/DB devices.
 		resp := m.addSingleDisk(ctx, disks[0], wal, db)
 		ret.Reports = append(ret.Reports, resp)
@@ -705,7 +710,7 @@ func (m *OSDManager) addBulkDisks(ctx context.Context, disks []types.DiskParamet
 
 // addSingleDisk is a wrapper around AddOSD which logs disk addition failures and returns a formatted response.
 func (m *OSDManager) addSingleDisk(ctx context.Context, disk types.DiskParameter, wal *types.DiskParameter, db *types.DiskParameter) types.DiskAddReport {
-	if strings.Contains(disk.Path, constants.LoopSpecId) {
+	if isLoopDevice(disk) {
 		// Add file based OSDs.
 		err := m.addLoopBackOSDs(ctx, disk.Path)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


sabaini (commented):
Thanks @johnramsden lgtm! Need to get CI green then it's gtg

microceph/ceph/osd.go
```diff
@@ -191,7 +191,7 @@ func (m *OSDManager) setupEncryptedOSD(devicePath string, osdDataPath string, os
 	}
 
 	// Encrypt the device
-	encryptDevice(devicePath, key)
+	err = encryptDevice(devicePath, key)
```
UtkarshBhatthere, line 194: nice catch

microceph/ceph/osd.go
```diff
@@ -739,9 +744,20 @@ func (m *OSDManager) addOSD(ctx context.Context, data types.DiskParameter, wal *
...
+}
+
 func (m *OSDManager) validateAddOSDArgs(data types.DiskParameter, wal *types.DiskParameter, db *types.DiskParameter) error {
-	if data.LoopSize != 0 && (wal != nil || db != nil) {
-		return fmt.Errorf("loopback and WAL/DB are mutually exclusive")
+	if isLoopDevice(data) {
+		if wal != nil || db != nil {
+			return fmt.Errorf("loopback and WAL/DB are mutually exclusive")
+		}
+		if data.Encrypt {
```
UtkarshBhatthere, line 758: is this a limitation that we are imposing ? Because this will make it harder for us to test encryption with loop devices (most of our CI)

UtkarshBhatthere (commented):
lgtm, left some comments/ questions

microceph/ceph/osd.go
johnramsden: It says in the [documentation](https://canonical-microceph.readthedocs-hosted.com/stable/reference/commands/disk/): 

> WAL and DB devices can only be used with data devices that reside on a block device, not with loop files. Loop files do not support encryption.

Additionally, I believe the CI [uses](https://github.com/canonical/microceph/blob/afb6a3a0cb2a28edc1c5d10ef0a16b77ccd56dd7/.github/workflows/tests.yml#L125) manually created loop devices, not microceph managed "loop files" - these will not be identified as different from a block device, and do support encryption.

UtkarshBhatthere:
@johnramsden have triggered the tests.

microceph/ceph/osd.go
UtkarshBhatthere: ack, that sounds good to me.

PR #680, opened 2026-02-26
fix: Resolve missing unlock encrypted wal/db at osd start 

### LINKED ISSUE #655: OSD fails to start after reboot if using encrypted WAL or DB devices
### Description

The latest version of microceph supports encrypting DB and WAL devices, if provided separately:
```
Usage:
  microceph disk add <SPEC> [flags]

Flags:
      --all-available       add all available devices as OSDs
      --db-device string    The device used for the DB
      --db-encrypt          Encrypt the DB device prior to use
      --db-wipe             Wipe the DB device prior to use
      --encrypt             Encrypt the disk prior to use
      --wal-device string   The device used for WAL
      --wal-encrypt         Encrypt the WAL device prior to use
      --wal-wipe            Wipe the WAL device prior to use
      --wipe                Wipe the disk prior to use
```

While OSD with encrypted DB and WAL gets created successfully, these devices do not get unlocked upon the next reboot as can be seen from the [source code](https://github.com/canonical/microceph/blob/9c99d97567fe3c5707e3a0f42d8aac9a9bdf03c5/snapcraft/commands/osd.start#L52).

Consequently, the `microceph.osd` service fails to start producing errors similar to the following ones:
```
Dec 02 10:12:07 node1 systemd[1]: Started snap.microceph.osd.service - Service for snap application microceph.osd.
Dec 02 10:12:09 node1 microceph.osd[104948]: 2025-12-02T10:12:09.829+0000 7996b2994640 -1 bluestore(/var/lib/ceph/osd/ceph-10) _minimal_open_bluefs /var/lib/ceph/osd/ceph-10/block.wal symlink exists but target unusable: (2) No such file or directory
Dec 02 10:12:12 node1 microceph.osd[104948]: 2025-12-02T10:12:12.595+0000 7996b2994640 -1 bluestore(/var/lib/ceph/osd/ceph-10) _minimal_open_bluefs /var/lib/ceph/osd/ceph-10/block.wal symlink exists but target unusable: (2) No such file or directory
Dec 02 10:12:12 node1 microceph.osd[104948]: 2025-12-02T10:12:12.619+0000 7996b2994640 -1 bluestore(/var/lib/ceph/osd/ceph-10) _open_db failed to prepare db environment:
Dec 02 10:12:12 node1 microceph.osd[104948]: 2025-12-02T10:12:12.855+0000 7996b2994640 -1 osd.10 0 OSD:init: unable to mount object store
Dec 02 10:12:12 node1 microceph.osd[104948]: 2025-12-02T10:12:12.855+0000 7996b2994640 -1  ** ERROR: osd init failed: (5) Input/output error
Dec 02 10:12:12 node1 systemd[1]: snap.microceph.osd.service: Main process exited, code=exited, status=250/n/a
```

### Steps to Reproduce
1.  Create a MicroCeph cluster.
2.  Add an OSD with encryption and a separate WAL device (e.g., using LVM paths):
    ```bash
    microceph disk add /dev/mapper/data-lv --wal-device /dev/mapper/wal-lv --wal-encrypt
    ```
3.  Observe that the OSD starts successfully (`systemctl status snap.microceph.osd`)
4.  Reboot the node (restarting `snap.microceph.osd` is not enough).
5.  Observe that the service `microceph.osd` fails to start.

### Code Analysis / Root Cause
There is a discrepancy between the OSD Creation logic (Go) and the OSD Startup logic (Shell).

1.  **Creation (Correct):**
    In `microceph/ceph/osd.go`, the function `setupEncryptedOSD` correctly handles Block, WAL, and DB using the corresponding "suffix". It formats and **opens** (unlocks) all of them explicitly.
    *   [Reference: microceph/ceph/osd.go](https://github.com/canonical/microceph/blob/9c99d97567fe3c5707e3a0f42d8aac9a9bdf03c5/microceph/ceph/osd.go#L167)

2.  **Startup (Bug):**
    In `snapcraft/commands/osd.start`, the `spawn()` function iterates through the OSD directories.
    *   It checks for `${i}/unencrypted` (Main Block) and calls `maybe_unlock`.
    *   **It fails to check for or unlock** `${i}/unencrypted.wal` or `${i}/unencrypted.db`.
    *   [Reference: snapcraft/commands/osd.start](https://github.com/canonical/microceph/blob/9c99d97567fe3c5707e3a0f42d8aac9a9bdf03c5/snapcraft/commands/osd.start#L52)

Because the shell script ignores the `unencrypted.wal` symlink, the mapper is never created, and BlueStore cannot access its write-ahead log.

### Proposed Fix
Update `snapcraft/commands/osd.start` to check for and unlock separate WAL/DB devices if they exist, mirroring the logic for the main block device.

```bash
# Pseudocode logic missing in osd.start:
if [ -b "${i}/unencrypted.db" ] ; then
    maybe_unlock "${i}/unencrypted.db" "${nr}" "$( get_key "${nr}" )"
fi
if [ -b "${i}/unencrypted.wal" ] ; then
    maybe_unlock "${i}/unencrypted.wal" "${nr}" "$( get_key "${nr}" )"
fi
```

### Manual Workaround
The OSD can be started manually by retrieving the key from the config store and unlocking the volume via `cryptsetup` before starting the service.

1. Identify the missing symlink to the unloked device, e.g.:
```
ls -l /var/snap/microceph/common/data/osd/ceph-X/block.wal
```
Assume the result is, e.g.: `/dev/mapper/luksosd.wal-X`
2. Dump microceph keys using the command `microceph.ceph config-key dump`
3. Copy the key for the encrypted device, e.g. `microceph:osd.wal.X/key`
4. Unlock the decrypted device using the output from Step 1:
```
cryptsetup open --type luks /var/snap/microceph/common/data/osd/ceph-X/unencrypted.wal luksosd.wal-X
```
5. Restart the OSD service:
```
systemctl start snap.microceph.osd
```

### Environment
*   **MicroCeph Version:** `ceph-version: 19.2.1-0ubuntu0.24.04.2; microceph-git: 74c0060321`
*   **OS:** `Ubuntu 24.04 LTS` 
*   **Storage Backend:** LVM

johnramsden:
# Description

On node reboot, or if devices are closed encrypted wal/dbs will not be unlocked correctly and will lead to issues on startup. 

The relevant issue #655 describes the problem very well:

> There is a discrepancy between the OSD Creation logic (Go) and the OSD Startup logic (Shell).
> 
> 1. **Creation (Correct):**
>    In `microceph/ceph/osd.go`, the function `setupEncryptedOSD` correctly handles Block, WAL, and DB using the corresponding "suffix". It formats and **opens** (unlocks) all of them explicitly.
>    
>    * [Reference: microceph/ceph/osd.go](https://github.com/canonical/microceph/blob/9c99d97567fe3c5707e3a0f42d8aac9a9bdf03c5/microceph/ceph/osd.go#L167)
> 2. **Startup (Bug):**
>    In `snapcraft/commands/osd.start`, the `spawn()` function iterates through the OSD directories.
>    
>    * It checks for `${i}/unencrypted` (Main Block) and calls `maybe_unlock`.
>    * **It fails to check for or unlock** `${i}/unencrypted.wal` or `${i}/unencrypted.db`.
>    * [Reference: snapcraft/commands/osd.start](https://github.com/canonical/microceph/blob/9c99d97567fe3c5707e3a0f42d8aac9a9bdf03c5/snapcraft/commands/osd.start#L52)
> 
> Because the shell script ignores the `unencrypted.wal` symlink, the mapper is never created, and BlueStore cannot access its write-ahead log.
> 
> ### Proposed Fix
> Update `snapcraft/commands/osd.start` to check for and unlock separate WAL/DB devices if they exist, mirroring the logic for the main block device.
> 
> # Pseudocode logic missing in osd.start:
> if [ -b "${i}/unencrypted.db" ] ; then
>     maybe_unlock "${i}/unencrypted.db" "${nr}" "$( get_key "${nr}" )"
> fi
> if [ -b "${i}/unencrypted.wal" ] ; then
>     maybe_unlock "${i}/unencrypted.wal" "${nr}" "$( get_key "${nr}" )"
> fi

Add tests and fix the unlock.

Fixes #655

## Type of change

## How has this been tested?

This has been tested locally by executing: 

```
INITIAL_OSD_COUNT=$(microceph.ceph -s -f json | jq -r '.osdmap.num_in_osds // 0')
EXPECTED=$((INITIAL_OSD_COUNT + 1))
tests/scripts/actionutils.sh test_encrypted_wal_db_startup "${EXPECTED}" 
```

The test which is also been added to CI creates three loop back devices, and then creates an OSD with an encrypted WAL & DB. This happens successfully. 

The test then shuts down the OSD, closes the relevant encrypted devices and then starts the OSD back up. This should lead to the OSD being opened successfully. Instead it fails.

## Contributor checklist

Please check that you have:

- [x] added tests to verify effectiveness of this change reply with my

### DIFF
--- a/tests/scripts/actionutils.sh
+++ b/tests/scripts/actionutils.sh
@@ -99,6 +99,7 @@ function add_encrypted_osds() {
 }
 
 function test_encrypted_wal_db_startup() {
+    set -eux
     sudo snap connect microceph:dm-crypt
     sudo snap restart microceph.daemon
     local expected_osds="${1?missing}"


tests/scripts/actionutils.sh
```diff
@@ -98,6 +98,52 @@ function add_encrypted_osds() {
...
+
+    # Find the OSD ID just created
+    local osd_id
+    osd_id=$(sudo ls -1 /var/snap/microceph/common/data/osd/ | sort -t- -k2 -n | tail -1 | sed 's/ceph-//')
+
+    # Simulate reboot: stop service and close all LUKS volumes
+    sudo snap stop microceph.osd
+    sudo cryptsetup close "luksosd-${osd_id}"
+    sudo cryptsetup close "luksosd.wal-${osd_id}"
+    sudo cryptsetup close "luksosd.db-${osd_id}"
```
sabaini, line 131: Nit: we're not checking if closing succeeded and aren't running `set -e` so if `cryptsetup close` failed we could get a false pos. for the open check below I think? suggest to run this in a ( set -e ... ) block and/or check if mappers are really absent

johnramsden:
I believe the current errors are not related to this PR

PR #683, opened 2026-02-27
fix: Add waitready command to verify cluster ready

### LINKED ISSUE #653: `microceph wait-ready` a way to wait for the microceph cluster untill it's up
In some CI runs (possibly be the strained ones) race conditions with sibling projects (LXD, MicroCloud) issuing commands before the cluster was ready to accept the operations. Refer to the [original issue](https://github.com/canonical/microcloud/issues/653).

johnramsden:
# Description

When an operator attempts to do something before the cluster is up they can receive unexpected failures because bootstrap is not finished or microcluster is not yet available. This can be particularly problematic in CI or scripting.

Add an additional subcommand (similar to lxd waitready) https://manpages.debian.org/unstable/lxd/lxd.waitready.1

To confirm the cluster is up we check for the microcluster daemon to be ready, and for ceph to be ready (ceph -s)

On failure we get a message like the following if we haven't bootstrapped for example:

```
microceph waitready --timeout 30
Error: ceph not ready: timed out waiting for Ceph to become ready: context deadline exceeded
```

Running the following you should expect it to wait before running status, and it should succeed

```
sudo microceph cluster bootstrap &
sudo microceph waitready
sudo microceph status
[1] 35966
MicroCeph deployment summary:
- microceph (10.56.203.112) Services: mds, mgr, mon Disks: 0
```

Also add --storage flag:

When --storage is passed, after daemon and monitor readiness, poll until enough OSDs are up to satisfy pool replication requirements.

The required count is max(pool.Size) across all pools, falling back to osd_pool_default_size if no pools exist.

Update GetOSDPools to accept a context allowing us to reuse functionality

Fixes #653 
Fixes: #683

## Type of change

## How has this been tested?

Added tests demonstrating waiting and timeout prior to bootstrap, and waiting succeeding post bootstrap.

## Contributor checklist

Please check that you have:


microceph/ceph/monitor.go
```diff
@@ -125,3 +127,23 @@ func getActiveMons() ([]string, error) {
 
 	return activeMons, nil
 }
+
+// WaitForCephReady polls "ceph -s" until it succeeds, indicating that
+// the Ceph can accept commands.
+// It retries every second until success or the context is cancelled/expired.
+func WaitForCephReady(ctx context.Context) error {
+	for {
+		_, err := common.ProcessExec.RunCommand("ceph", "-s")
```
sabaini: Hm, a hanging `ceph -s` could hang this forever, not sure how often this occurs in practice but just for robustness could use the cephRunContext() function and pass in the ctx

microceph/cmd/microceph/waitready.go
```diff
@@ -0,0 +1,57 @@
...
+}
+
+func (c *cmdWaitready) Run(cmd *cobra.Command, args []string) error {
+	m, err := microcluster.App(microcluster.Args{StateDir: c.common.FlagStateDir})
+	if err != nil {
+		return err
+	}
+
+	ctx := context.Background()
+	if c.flagTimeout > 0 {
```
sabaini, line 43: Minor nit: should we be erroring out if operators pass in a neg. timeout value?

johnramsden: Good catch. I set it to unsigned and modified the error message to include a bit of context. I think it's nicer to have the type safety and have it unsigned but it means we receive a bit of a messy message: 

> Error: invalid argument "-1" for "--timeout" flag: strconv.ParseUint: parsing "-1": invalid syntax: timeout must be a positive number of seconds 

The alternative is to switch back to an integer and have a nicer message that is just 

> Error: invalid argument "-1" for "--timeout" flag:  timeout must be a positive number of seconds 

But I think this tradeoff is worth it fine for the type safety

microceph/ceph/monitor.go
johnramsden: See https://github.com/canonical/microceph/pull/683/commits/c49678a471849053d58c8b6da900f9ead85ddbfd

UtkarshBhatthere (commented):
overall lgtm, that being said, imo a `--storage` flag would make this even better. Since microceph has opinions on what should be a minimum storage setup it should wait untill enough OSDs to spawn a storage pool are available (i.e. if rf==1 - > one OSD and if rf==3 -> 3 OSD). IDK what should the criteria be for EC pools.

microceph/cmd/microceph/waitready.go
```diff
@@ -0,0 +1,60 @@
...
+func (c *cmdWaitready) Command() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "waitready",
+		Short: "Wait until the daemon is ready and Ceph is operational",
+		RunE:  c.Run,
+	}
+
+	cmd.Flags().Uint64Var(&c.flagTimeout, "timeout", 0, "Number of seconds to wait before giving up (0 = indefinitely)")
+	cmd.SetFlagErrorFunc(func(cmd *cobra.Command, err error) error {
+		return fmt.Errorf("%w: timeout must be a positive number of seconds", err)
```
sabaini: Totally OCD nit, suggest `%w: timeout must be a positive number of seconds or zero` or similar

.github/workflows/tests.yml
```diff
@@ -245,6 +248,12 @@ jobs:
           ~/actionutils.sh hurl tests/hurl/services-mon.hurl
           ~/actionutils.sh hurl tests/hurl/maintenance-put-failed.hurl
 
+      - name: Test microceph waitready not bootstrapped
+        run: |
+          sudo snap remove microceph --purge
+          sudo snap install --dangerous ~/microceph_*.snap
```
sabaini: Suggest to connect the interfaces as well to ensure the waitready check fails *only* because of the missing bootstrap for future proofing the test

```
    sudo snap install --dangerous ~/microceph_*.snap
    sudo snap connect microceph:block-devices
    sudo snap connect microceph:hardware-observe
    sudo snap connect microceph:mount-observe
    sudo snap connect microceph:load-rbd
    sudo snap connect microceph:microceph-support
    sudo snap connect microceph:network-bind
    sudo snap connect microceph:process-control
```

.github/workflows/tests.yml
```diff
@@ -245,6 +248,12 @@ jobs:
...
+          ~/actionutils.sh test_waitready_before_bootstrap
```
sabaini: After this test the cluster is unbootstrapped and non-operational; just as a warning for future travellers might add a comment to that effect here.

microceph/ceph/monitor.go
```diff
@@ -147,3 +149,74 @@ func WaitForCephReady(ctx context.Context) error {
...
 }
+
+// WaitForOSDsReady polls until enough OSDs are up to satisfy pool replication
+// requirements. The required count is max(pool.Size) across all pools, falling
+// back to osd_pool_default_size if no pools exist.
+func WaitForOSDsReady(ctx context.Context) error {
+	required, err := getRequiredOSDCount(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to determine required OSD count: %w", err)
+	}
```
sabaini: Not a blocker but I wonder if we should run `getRequiredOSDCount(ctx)` inside the for{} loop to give it more chances in case the cluster is unstable right after deployment

microceph/ceph/monitor.go
```diff
@@ -125,3 +129,94 @@ func getActiveMons() ([]string, error) {
 
 	return activeMons, nil
 }
+
+// WaitForCephReady polls "ceph -s" until it succeeds, indicating that
+// the Ceph can accept commands.
+// It retries every second until success or the context is cancelled/expired.
+func WaitForCephReady(ctx context.Context) error {
```
UtkarshBhatthere: `monitor.go` is usually for monitor orchestration code only. Can we place organise this somewhere else ? I am sure there are `x_helper.go` files as well.

microceph/ceph/monitor.go
```diff
@@ -125,3 +129,94 @@ func getActiveMons() ([]string, error) {
...
+			return fmt.Errorf("timed out waiting for Ceph to become ready: %w", ctx.Err())
+		case <-time.After(time.Second):
+		}
+	}
+}
+
+// WaitForOSDsReady polls until enough OSDs are up to satisfy pool replication
+// requirements. The required count is max(pool.Size) across all pools, falling
+// back to osd_pool_default_size if no pools exist.
+func WaitForOSDsReady(ctx context.Context) error {
```
UtkarshBhatthere: Similarly for OSD, having it under expected places makes it easier to discover code.

.github/workflows/tests.yml
```diff
@@ -245,6 +254,20 @@ jobs:
           ~/actionutils.sh hurl tests/hurl/services-mon.hurl
           ~/actionutils.sh hurl tests/hurl/maintenance-put-failed.hurl
 
+      # Note: After this test the cluster is unbootstrapped and non-operational.
```
UtkarshBhatthere: why are we testing the before bootstrap scenario like this ? Wouldn't it be more suitable to execute the pre bootstrap validation and post bootstrap validation in the same test (single system test) ?

johnramsden: See https://github.com/canonical/microceph/pull/683/commits/add14daf286ffaa574992cddcff3fc95dd51b930

Moved it into a single test, also split install_microceph into {install,bootstrap}_microceph, and added install_and_bootstrap

PR #687, opened 2026-03-05
Add Availability Zone Support

johnramsden:
# Description

Add support for availability zones to MicroCeph.

* Add a bootstrap and join flag `--availability-zone` that adds the AZ to the database.
* If a host is added without an AZ, and AZs exist, the join will be rejected. 
* If a host is added with an AZ and AZs are empty on other hosts, reject the join. In this scenario the operator must recreate the cluster.
* CRUSH map topology is changed to move each host under its stored AZ as a rack failure domain when they have an AZ
* Upon reaching 3 AZs, a CRUSH rule is added for racks - microceph_auto_rack
  * Reaching 3 AZs here means 3 nodes with at least one OSD each in 3 different AZs
  * The pool also has the relevant rule set (setPoolCrushRule)
* Operator responsibility to ensure balanced AZs

## Type of change

## How has this been tested?

Unit tests:                                                                                
  - AZ join validation: accepts AZ when existing nodes have AZs, rejects AZ when existing    
  nodes don't have one (and vice versa), allows no AZ when none exist                        
  - updateTopology no-op when no AZ is configured                                          
  - updateTopology creates CRUSH buckets but skips rack rule with <3 AZs
  - updateTopology switches to rack rule when 3+ AZs all have OSDs
  - updateTopology skips rack rule when 3+ AZs exist but not all have OSDs
  - updateFailureDomain skips host-based scaling when AZs are configured
  - updateFailureDomain works normally (osd→host at 3+ nodes) without AZs
  - Adopt bootstrapper Prefill correctly sets AvailabilityZone

Integration test:
  - 4-node cluster across 3 AZs, adding OSDs one at a time and verifying the CRUSH rule stays
   on OSD until all 3 AZs have an OSD, then switches to rack

## Contributor checklist

Please check that you have:

### DIFF
--- a/tests/scripts/actionutils.sh
+++ b/tests/scripts/actionutils.sh
@@ -1893,18 +1893,30 @@ function test_az_lifecycle() {
     echo "node-wrk3 has osd.${wrk3_osd}"
 
     # Removing the last OSD in az-c should be blocked (would break rack failure domain)
-    if lxc exec node-wrk3 -- sh -c "microceph disk remove osd.${wrk3_osd}" 2>&1 ; then
+    local remove_out remove_rc
+    remove_out=$(lxc exec node-wrk3 -- sh -c "microceph disk remove osd.${wrk3_osd}" 2>&1) && remove_rc=0 || remove_rc=$?
+    if echo "${remove_out}" | grep -q "rack-level failure domain" ; then
+        echo "Confirmed: disk remove blocked as expected (rack failure domain protection)"
+    elif [[ $remove_rc -eq 0 ]] ; then
         echo "FAIL: expected disk remove to be blocked but it succeeded"
         return 1
-    fi
-    echo "Confirmed: disk remove blocked as expected (rack failure domain protection)"
-
-    # Verify --confirm-failure-domain-downgrade is also a no-op on rack topology
-    if lxc exec node-wrk3 -- sh -c "microceph disk remove osd.${wrk3_osd} --confirm-failure-domain-downgrade" 2>&1 ; then
-        echo "FAIL: expected disk remove to be blocked even with --confirm-failure-domain-downgrade"
+    else
+        echo "FAIL: disk remove failed with unexpected error: ${remove_out}"
         return 1
     fi
-    echo "Confirmed: --confirm-failure-domain-downgrade does not bypass rack protection"
+
+    # Verify --confirm-failure-domain-downgrade is also a no-op on rack topology
+    local downgrade_out downgrade_rc
+    downgrade_out=$(lxc exec node-wrk3 -- sh -c "microceph disk remove osd.${wrk3_osd} --confirm-failure-domain-downgrade" 2>&1) && downgrade_rc=0 || downgrade_rc=$?
+    if echo "${downgrade_out}" | grep -q "rack-level failure domain" ; then
+        echo "Confirmed: --confirm-failure-domain-downgrade does not bypass rack protection"
+    elif [[ $downgrade_rc -eq 0 ]] ; then
+        echo "FAIL: expected disk remove to be blocked even with --confirm-failure-domain-downgrade but it succeeded"
+        return 1
+    else
+        echo "FAIL: disk remove failed with unexpected error: ${downgrade_out}"
+        return 1
+    fi
 
     # Force removal with --bypass-safety-checks
     lxc exec node-wrk3 -- sh -c "microceph disk remove osd.${wrk3_osd} --bypass-safety-checks"


microceph/ceph/osd.go
```diff
@@ -342,11 +342,128 @@ func (m *OSDManager) switchFailureDomain(old string, new string) error {
...
+	// No AZ for this host — non-AZ cluster.
+	if data.hostAZ == "" {
+		return nil
+	}
+
+	logger.Infof("Setting up CRUSH topology for host %s in AZ %s", hostname, data.hostAZ)
+
+	// Create the AZ rack bucket and move it under root default.
+	// Doesn't fail on existing, can repeat safely if failure
+	_, err = cephRunContext(ctx, "osd", "crush", "add-bucket", data.hostAZ, "rack")
```
sabaini: IIUC data.hostAZ is (validated) user input, I wonder if it would be wise to ensure it doesn't collide with e.g. things like 'default', could be for instance prefix it with `az.{data.hostAZ}` or similar

microceph/ceph/join.go
```diff
@@ -100,6 +132,48 @@ func Join(ctx context.Context, s interfaces.StateInterface) error {
...
+	if az == "" {
+		return nil // Empty AZ, dont add
+	}
+
+	if !IsValidCrushName(az) {
+		return fmt.Errorf("invalid availability zone name %q: must match [a-zA-Z0-9_.-]+", az)
+	}
+
+	key := fmt.Sprintf("az.host.%s", hostname)
+	_, err = joinCreateConfigItem(ctx, tx, database.ConfigItem{Key: key, Value: az})
```
sabaini: Hm, I see we're adding az.host.$hostname items but do we also delete them when removing nodes?

microceph/ceph/osd.go
```diff
@@ -938,6 +1055,12 @@ func (m *OSDManager) doAddOSD(ctx context.Context, data types.DiskParameter, wal
 		return err
 	}
 
+	err = m.updateTopology(ctx)
```
sabaini, line 1062: I now realize we have similar risk already -- but we have a risk of partial failures if updateTopology() errors, it could leave the spawned OSD around, similar to the below invocations

microceph/ceph/crush.go
```diff
@@ -145,6 +155,53 @@ func getDefaultCrushRule() (string, error) {
...
+	nodes := gjson.Get(output, "nodes")
+	count := 0
+	for az := range azNames {
+		// The current host's AZ is always active — we're adding an OSD to it right now.
+		if az == currentAZ {
+			count++
+			continue
+		}
+		// Find the rack node for this AZ and get its children (host IDs).
+		children := nodes.Get(fmt.Sprintf(`#(name=="%s").children`, az))
```
sabaini: Hm, might make sense to add a type constraint as well in case of name collisions btw nodes

sabaini (changes requested):
Hey @johnramsden thanks, this looks nice! -- some drive-by comments below

microceph/ceph/osd.go
johnramsden: About the approach:

Do you think we should address this in this PR? It is a little bit complex because how do we approach errors - cleanup can also fail. 

Would you like me to introduce a clean up mechanism here? Would probably be best effort since failures could occur during cleanup.

microceph/ceph/osd.go
```diff
@@ -342,11 +342,130 @@ func (m *OSDManager) switchFailureDomain(old string, new string) error {
...
+		return nil
+	}
+
+	logger.Infof("Setting up CRUSH topology for host %s in AZ %s", hostname, data.hostAZ)
+
+	// Create the AZ rack bucket and move it under root default.
+	// Prefix AZ names with "az-" to avoid collisions with existing CRUSH bucket names.
+	// Doesn't fail on existing, can repeat safely if failure
+	rackBucket := fmt.Sprintf("az.%s", data.hostAZ)
+	_, err = cephRunContext(ctx, "osd", "crush", "add-bucket", rackBucket, "rack")
```
sabaini, line 400: I believe it's correct that re-running `ceph osd crush add-bucket foo ...` won't error -- however might be worthwhile to prove this in the functest should `ceph` decide to throw EEXIST at some point

microceph/ceph/osd.go
```diff
@@ -342,11 +342,130 @@ func (m *OSDManager) switchFailureDomain(old string, new string) error {
...
+	if err != nil {
+		return fmt.Errorf("failed to create rack bucket %s: %w", rackBucket, err)
+	}
+	_, err = cephRunContext(ctx, "osd", "crush", "move", rackBucket, "root=default")
+	if err != nil {
+		return fmt.Errorf("failed to move rack %s under root: %w", rackBucket, err)
+	}
+
+	// Create the host bucket and move it under the AZ rack.
+	_, err = cephRunContext(ctx, "osd", "crush", "add-bucket", hostname, "host")
```
sabaini, line 413: Hm, I believe ceph already would create host buckets, does it not? If it's necessary we can create them here as well for robustness, otoh I don't think I've seen ceph fail to do this.

microceph/cmd/microcephd/adopt_bootstrapper.go
```diff
@@ -176,13 +178,21 @@ func (ab *AdoptBootstrapper) updateCephClusterConfigs() error {
...
 	configs := map[string]string{
 		"fsid":                          ab.FSID,
 		constants.AdminKeyringFieldName: ab.AdminKey,
 		"public_network":                ab.PublicNet,
 	}
 
+	// If AZ present, validate and record
+	if ab.AvailabilityZone != "" {
+		if !ceph.IsValidCrushName(ab.AvailabilityZone) {
+			return nil, fmt.Errorf("invalid availability zone name %q: must match [a-zA-Z0-9_.-]+", ab.AvailabilityZone)
```
sabaini: Could we move this validation to Precheck() to fail as early as possible?

microceph/ceph/join.go
sabaini: joinCreateConfig() will fail if the entry already exists. I wonder if we could make this more robust with upsert semantics i.e. update an entry if it already exists, create it if not. This would help in cases where nodes are dead and couldn't properly be removed or so

microceph/ceph/osd.go
```diff
@@ -342,11 +342,130 @@ func (m *OSDManager) switchFailureDomain(old string, new string) error {
...
+
+	// No AZ for this host — non-AZ cluster.
+	if data.hostAZ == "" {
+		return nil
+	}
+
+	logger.Infof("Setting up CRUSH topology for host %s in AZ %s", hostname, data.hostAZ)
+
+	// Create the AZ rack bucket and move it under root default.
+	// Prefix AZ names with "az-" to avoid collisions with existing CRUSH bucket names.
```
sabaini: typo s/az-/az./

microceph/ceph/osd.go
sabaini: Think we don't need to block ovedr this. I've added CEPH-1690 for tracking

microceph/cmd/microcephd/adopt_bootstrapper.go
sabaini: Similarly in simple_bootstrapper.go too

Also suggest to add a test around validation failure

microceph/cmd/microcephd/adopt_bootstrapper.go
```diff
@@ -176,13 +178,21 @@ func (ab *AdoptBootstrapper) updateCephClusterConfigs() error {
...
+		}
+		configs[fmt.Sprintf("az.host.%s", hostname)] = ab.AvailabilityZone
```
UtkarshBhatthere: The problem with overloading the configs table is that we have to manage the cleanup (in case of host removal) ourselves which we get for free if the table has a `ON DELETE CASCADE` relation.

What are your thoughts about persisting this information in a new table called `HostTags` ? The idea is that we can have various string tags associated with each cluster member. With this we get a flexible tagging system we can use to tag hosts for AZs as well as extendable in future for things like service placement. And we get the cleanup for free from dqlite.

microceph/ceph/crush.go
```diff
@@ -12,6 +13,15 @@ import (
 	"github.com/tidwall/gjson"
 )
 
+// validCrushName matches the validation in Ceph's CrushWrapper::is_valid_crush_name.
+// https://github.com/ceph/ceph/blob/main/src/crush/CrushWrapper.cc
+var validCrushName = regexp.MustCompile(`^[a-zA-Z0-9_.\-]+$`)
```
UtkarshBhatthere: I think the regexp can be a constant string which can be fed into the regexp object right inside the IsValidCrushName method (rather than being a top level item in the module)

microceph/ceph/osd.go
johnramsden: There appears to be a race condition. I added a comment to the code. Without it we get a failure

microceph/ceph/osd.go
johnramsden: Added test case

microceph/cmd/microcephd/simple_bootstrapper.go
```diff
@@ -69,6 +71,13 @@ func (sb *SimpleBootstrapper) Precheck(ctx context.Context, state interfaces.Sta
 		return err
 	}
 
+	// Validate AZ name early, before any cluster state is modified.
```
UtkarshBhatthere, line 74: can we also add `PreInit` hook based checks to ensure AZ is provided for joining node if it was provided for bootstrap node and vice versa.

microceph/ceph/osd.go
```diff
@@ -342,11 +342,131 @@ func (m *OSDManager) switchFailureDomain(old string, new string) error {
...
+
+	// Create the host bucket explicitly before moving it under the AZ rack.
+	// Although Ceph creates host buckets when an OSD is added, there is a race
+	// between OSD registration and this topology update — the host bucket may
+	// not yet exist in the CRUSH map, causing "crush move" to fail with ENOENT.
+	_, err = cephRunContext(ctx, "osd", "crush", "add-bucket", hostname, "host")
+	if err != nil {
+		return fmt.Errorf("failed to create host bucket %s: %w", hostname, err)
+	}
+	_, err = cephRunContext(ctx, "osd", "crush", "move", hostname, fmt.Sprintf("rack=%s", rackBucket))
```
UtkarshBhatthere, line 417: Is a respective removal done when the node is removed from the cluster ?

UtkarshBhatthere: Also, what about the last node on the AZ ? Should we ask for --bypass-safety-checks if the last node is being removed ?

microceph/cmd/microcephd/simple_bootstrapper.go
johnramsden: Unfortunately I don't think we can do that unless we have access to the database which happens after PreInit. I tried breaking it up but I discovered the database is not available until after PreInit so I had to revert it. I kept validation and setting separate to keep it a bit more modular.

microceph/ceph/osd.go
johnramsden: While looking into this I found a potential issue. Right now there's nothing warning the user if they try to remove the last osd in an AZ, which I believe is the real concern. I added something that will block in this scenario, so that we do not remove the last OSD in an AZ if we have exactly 3 AZs. I added a check gated on BypassSafety. I believe this has resolved any of the concerns we might have. 

Your comment about removing buckets - this is just cosmetic, and the operator could clean them up if they chose to. For now I've let them dangling. I think we could clean them up. It does introduce a potential race condition if we get into a situation where we are trying to clean them while someone is trying to add an osd to that AZ, but we could introduce a lock for this purpose. Let me know if you think it makes sense to clean them up.

johnramsden: https://github.com/canonical/microceph/pull/687/commits/efec80c5145d5ee39392a0155abeacd8a77a34a6

johnramsden: @sabaini this was my concern from design hour regarding removing the buckets:    
                                                                                             
1. Node B removes the last OSD from a rack, sees the rack bucket is now empty, and issues `ceph osd crush remove az.foo`
2. Node A is concurrently adding an OSD and calls updateTopology() which does ceph osd `crush move {hostname} rack=az.foo`                                                       
3. The rack gets removed between step 2's bucket creation and the move command, or, after the move succeeds the bucket gets deleted with the host still referencing it.        
                            
I think a reasonable solution for this would be to use a distributed lock via dqlite or forward relevant requests to the leader so that they are serialized.

sabaini: Gotcha, quite right. I think we may want to serialize all topo updates I suspect there is other racey potential. I'd go for forwarding requests to the leader, istr theres a facility to get the leader from the dqlite db

microceph/ceph/join.go
johnramsden: Fixed via db cleanup

microceph/cmd/microcephd/adopt_bootstrapper.go
johnramsden: Fixed

microceph/ceph/osd.go
johnramsden: @sabaini I do question the complexity that will be introduced as a result of trying to forward everything to a leader a little bit here. There's no downside I see to leaving dangling buckets in terms of functionality and correctness. 

I think introducing this change may be out of the scope of this PR. Since we will need to forward all relevant topology changes to the leader. This would involve changes to where requests are routed. We would basically need to modify the flow that leads to topology changes and clean up to always occur on the leader. Do we want to introduce this with this PR? It might make sense as a separate feature: Serialize topology changes to leader in order to avoid race conditions.

sabaini: We can defer, created CEPH-1707 for this

microceph/ceph/join.go
```diff
@@ -60,6 +76,17 @@ func Join(ctx context.Context, s interfaces.StateInterface) error {
 		return fmt.Errorf("failed to generate the configuration: %w", err)
 	}
 
+	// Validate and record the availability zone for this host.
+	err = s.ClusterState().Database().Transaction(ctx, func(ctx context.Context, tx *sql.Tx) error {
+		if err := validateJoinAZ(ctx, tx, jc.AvailabilityZone); err != nil {
+			return err
+		}
+		return setJoinAZ(ctx, tx, s.ClusterState().Name(), jc.AvailabilityZone)
```
sabaini, line 85: Could we make this retry-able? Aiui currently if we fail down later we won't be able to retry as setJoinAz() would fail I think?

johnramsden: See https://github.com/canonical/microceph/pull/687/commits/892dd7f80a906824d221f94514e7e6c6c08de2f8, along with relevant details in the commit message

tests/scripts/adoptutils.sh
```diff
@@ -59,6 +59,18 @@ function bootstrap_cephadm() {
...
+  # Wait for the cluster to settle before adopt can connect
+  echo "Waiting for ceph cluster to become healthy..."
+  for i in $(seq 1 30); do
+    if lxc exec $name -- sh -c "cephadm shell -- ceph health 2>/dev/null" | grep -q "HEALTH_OK"; then
+      echo "Cluster is healthy"
+      break
+    fi
+    echo "Waiting for cluster health... attempt $i/30"
+    sleep 10s
+  done
```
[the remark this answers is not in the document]
UtkarshBhatthere: nack

microceph/database/host_tag.mapper.go
```diff
@@ -0,0 +1,420 @@
...
+UPDATE host_tags
+  SET member_id = (SELECT core_cluster_members.id FROM core_cluster_members WHERE core_cluster_members.name = ?), key = ?, value = ?
+ WHERE id = ?
+`)
+
+// hostTagColumns returns a string of column names to be used with a SELECT statement for the entity.
+// Use this function when building statements to retrieve database entries matching the HostTag entity.
+func hostTagColumns() string {
+	return "hosts_tags.id, core_cluster_members.name AS member, hosts_tags.key, hosts_tags.value"
+}
```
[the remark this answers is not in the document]
johnramsden: Dead code, regenerated, still wrong. Deleted - https://github.com/canonical/microceph/pull/687/commits/43fb20aa78b1fa504f59a11fc40a9605d5d8b8eb

tests/scripts/actionutils.sh
```diff
@@ -1917,6 +1767,213 @@ function test_waitready_storage() {
...
+    lxc exec node-wrk0 -- sh -c "microceph.ceph -s"
+    lxc exec node-wrk0 -- sh -c "microceph.ceph osd tree"
+
+    # Find the OSD number on node-wrk3 (az-c) dynamically
+    local wrk3_osd
+    wrk3_osd=$(lxc exec node-wrk0 -- sh -c "microceph disk list --json" | jq -r '.ConfiguredDisks[] | select(.location == "node-wrk3") | .osd')
+    echo "node-wrk3 has osd.${wrk3_osd}"
+
+    # Removing the last OSD in az-c should be blocked (would break rack failure domain)
+    if lxc exec node-wrk3 -- sh -c "microceph disk remove osd.${wrk3_osd}" 2>&1 ; then
```
UtkarshBhatthere: +1 for the check, however, this if block assumes that ANY failure that occurs during disk remove is caused by blocking. Is it worthwhile to also check against the error message ? (to make sure it is not, timeouts, EOF response (microcephd panic) or anything else)

johnramsden: https://github.com/canonical/microceph/pull/687/commits/d8957bece020756d428830a16ee57b45bd11a233

Issue #525, opened 2025-03-12
snap enable microceph won't enable all services


wideawakening:
spotted it on `squid/stable`

```
root@micro1:~# snap disable microceph
2025-03-11T08:43:29Z INFO Waiting for "snap.microceph.mds.service" to stop.
microceph disabled

root@micro1:~# snap services | grep microceph
Service                          Startup   Current   Notes
microceph.daemon                 disabled  inactive  -
microceph.mds                    disabled  inactive  -
microceph.mgr                    disabled  inactive  -
microceph.mon                    disabled  inactive  -
microceph.osd                    disabled  inactive  -
microceph.rbd-mirror             disabled  inactive  -
microceph.rgw                    disabled  inactive  -


root@micro1:~# snap enable microceph
microceph enabled

root@micro1:~# snap services  | grep microceph
Service                          Startup   Current   Notes
microceph.daemon                 enabled   active    -
microceph.mds                    disabled  inactive  -
microceph.mgr                    disabled  inactive  -
microceph.mon                    disabled  inactive  -
microceph.osd                    disabled  inactive  -
microceph.rbd-mirror             disabled  inactive  -
microceph.rgw                    disabled  inactive  -
```

sabaini:
Thanks for the report!

PR #691, opened 2026-03-18
fix: re-enable services after snap disable/enable cycle

johnramsden:
# Description

When `snap disable microceph` is run followed by `snap enable microceph`, only the main daemon service gets re-enabled. All other services (mon, mgr, mds, osd, rgw, etc.) remain disabled because they use `install-mode: disable` in snapcraft.yaml.

On daemon startup, query the database for services registered on this host and re-enable any that are inactive. OSD is always re-enabled since it is started unconditionally at bootstrap.

A serviceStartMu lock is used to prevent the re-enablement goroutine from racing with bootstrap/join, which also start services via snapctl. Without it, concurrent `snapctl start --enable` calls cause "service-control change in progress" errors.

Note: For consistency I also changed a few close by context deadlines to the slightly more correct pattern: 

```
- time.Sleep(time.Minute)                                              
+ select {                                                             
+     case <-ctx.Done():                                                   
+         return                                                             
+     case <-time.After(time.Minute):                                      
+ }
```

Also fixed bug where `getServicesForHost` overrode a param.

Fixes: #525

## Type of change

## How has this been tested?

Added integration test to confirm services are the same before/after enable/disable.

I decided unit tests did not make sense here since this feature is tightly coupled with the database and other areas. I did not determine it to be useful giving the amount of mocking that would be required.

## Contributor checklist

Please check that you have:

### DIFF
--- a/microceph/ceph/join.go
+++ b/microceph/ceph/join.go
@@ -106,7 +106,7 @@ func Join(ctx context.Context, s interfaces.StateInterface) error {
 }
 
 // getServicesForHost get services needed to be spawned on this machine.
-func getServicesForHost(ctx context.Context, s interfaces.StateInterface, hostname string) ([]database.Service, error) {
+var getServicesForHost = func(ctx context.Context, s interfaces.StateInterface, hostname string) ([]database.Service, error) {
 	var services []database.Service
 	err := s.ClusterState().Database().Transaction(ctx, func(ctx context.Context, tx *sql.Tx) error {
 		var err error
--- a/microceph/ceph/start_test.go
+++ b/microceph/ceph/start_test.go
@@ -1,15 +1,19 @@
 package ceph
 
 import (
+	"context"
 	"errors"
-	"github.com/canonical/microceph/microceph/common"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


tests/scripts/actionutils.sh
```diff
@@ -1630,6 +1630,67 @@ function verify_mount_check() {
...
+    echo "--- snap services before disable ---"
+    snap services microceph
+
+    # Record which services are both enabled and active before disable.
+    # Only these should come back — transiently started services (active
+    # but Startup=disabled) are not DB-registered and won't be restored.
+    local services_before
+    services_before=$(snap services microceph | awk '$2 == "enabled" && $3 == "active" {print $1}')
+    local count_before
+    count_before=$(echo "$services_before" | wc -l)
```
sabaini: Nit: tiny edge case here if there's no services before
```
echo "" | wc -l
1
```
Maybe sth like `count_before=$(printf '%s' "$services_before" | sed '/^$/d' | wc -l)`

johnramsden: added https://github.com/canonical/microceph/pull/691/commits/dfc179f851eb88ac4bb9a6997aba07fdb9807d06

PR #697, opened 2026-03-21
feat: Add support for SSL cert modification

### LINKED ISSUE #421: Adding easy certificate rotation for Rados Gateway in combination with LetsEncrypt Certificates
# Issue report

I first used `19.2.0~git+snap36f71d7700` to try out the rgw ssl features, but it didnt have all I needed, so I switched back to `18.2.4+snapc9f2b08f92`. Since I currently need to manually modify the radosgw.conf file anyways I prefer to be on stable version.

Use this section to describe the channel/revision which produces the unexpected behaviour.
This information can be fetched from the `installed:` section of `sudo snap info microceph` output.

1. Using certbot to create SSL Certificates
2. Configuring rgw to use the certificates with the method from https://github.com/canonical/microceph/pull/355
3. After 60 days the certificates are renewed by Certbot
4. RGW doesnt use the new certs.

The command to add the certificates to rgw uses base64 instead of files: https://github.com/canonical/microceph/pull/355#issuecomment-2294154200

This passes the certificates as `"$(sudo base64 -w0 ~/server.crt)"` which breaks the link with the original Letsencrypt certificate files, since it internally copies the files to the snap directory. Which means upon certbot renew the updated certificates are not updated for rgw.

It would be nice to have a `microceph update rgw ----ssl-certificate=.....` of some sort that can be placed into the certbot deploy hook for using the automated renewal process. Or in the ideal case have microceph handle the certbot setup/renewal as well.

The command I generally use for certbot config is:

```
  certbot certonly \
      --dns-cloudflare \
      --dns-cloudflare-credentials cloudflare.ini \
      --dns-cloudflare-propagation-seconds 90 \
      --non-interactive --agree-tos \
      --email my@email.com \
      -d "your.example.domain" \
      --deploy-hook "certbot_deploy_hook.sh"
```

If it’s considerably long, please paste to https://gist.github.com/ and insert the link here.

## Additional comments.

My current solution is to write a deploy hook that copies the certificates to ssl_certificate=/var/snap/microceph/common/server.crt ssl_private_key=/var/snap/microceph/common/server.key and then restarts the rgw service, but we should not be manually editing snap content, instead it should be going through the microceph api.

johnramsden:
# Description

Add support for SSL cert modification.

Currently there is no proper way to update certificates. This adds support for updating the existing certificates. 

The design includes a new certificates subcommand leaving room for adding additional certificate types. The alternative would be moving the functionality under "enable rgw". 

Unfortunately there is no current way of actually reloading certificates live. SIGHUP has no effect. There is a ssl_reload option for beast but it is only available in tentacle https://github.com/ceph/ceph/pull/65842. 

**Note:** Once we have migrated to tentacle, we can add functionality that supports auto reload. I propose what we should do here is have a reload occurring every 24 hours and issue a warning saying certificates will be updated in the background every 24 hours if the user does not issue the restart flag. I already mapped out this functionality once, it is fairly easy to add it is just not currently supported and the reload flag does nothing. Ill open an issue after merge.

Fixes #421

## Type of change

## How has this been tested?

Added unit and integration tests verifying:                                                
  - Certificate and key files are written with correct content and 0600 permissions
  - Command is rejected when RGW is not running                                              
  - Command is rejected when RGW was not configured with SSL
  - Invalid base64 certificate or private key is rejected with descriptive errors            
  - --restart causes immediate certificate pickup                                            
  - Without --restart, the old certificate continues to be served until manual restart       
  - --target rotates the certificate on a specific node (single-node self-target and         
  multi-node cross-node)

## Contributor checklist

Please check that you have:

### DIFF
--- a/microceph/ceph/rgw.go
+++ b/microceph/ceph/rgw.go
@@ -17,6 +17,13 @@ import (
 // writeSSLFiles decodes base64-encoded SSL certificate and key, and writes them to disk.
 // Returns the paths to the written certificate and key files.
 func writeSSLFiles(sslFilesPath string, sslCertificate string, sslPrivateKey string) (certPath string, keyPath string, err error) {
+	if sslCertificate == "" {
+		return "", "", fmt.Errorf("SSL certificate cannot be empty")
+	}
+	if sslPrivateKey == "" {
+		return "", "", fmt.Errorf("SSL private key cannot be empty")
+	}
+
 	decodedCert, err := base64.StdEncoding.DecodeString(sslCertificate)
 	if err != nil {
 		return "", "", fmt.Errorf("failed to decode SSL certificate: %w", err)
--- a/microceph/cmd/microceph/certificate_set_rgw.go
+++ b/microceph/cmd/microceph/certificate_set_rgw.go
@@ -44,7 +44,28 @@ manually for the change to take effect.`,
 	return cmd
 }
 
+// validateSSLInputs checks that the SSL certificate and key are non-empty and valid base64.
+func (c *cmdCertificateSetRGW) validateSSLInputs() error {
+	if c.flagSSLCertificate == "" {
+		return fmt.Errorf("SSL certificate cannot be empty")
+	}
+	if c.flagSSLPrivateKey == "" {
+		return fmt.Errorf("SSL private key cannot be empty")
+	}
+	if _, err := base64.StdEncoding.DecodeString(c.flagSSLCertificate); err != nil {
+		return fmt.Errorf("failed to decode SSL certificate: %w", err)
+	}
+	if _, err := base64.StdEncoding.DecodeString(c.flagSSLPrivateKey); err != nil {
+		return fmt.Errorf("failed to decode SSL private key: %w", err)
+	}
+	return nil
+}
+
 func (c *cmdCertificateSetRGW) Run(cmd *cobra.Command, args []string) error {
+	if err := c.validateSSLInputs(); err != nil {
+		return err
+	}
+
 	m, err := microcluster.App(microcluster.Args{StateDir: c.common.FlagStateDir})
 	if err != nil {
 		return err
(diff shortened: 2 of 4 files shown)


microceph/ceph/rgw.go
```diff
@@ -14,28 +14,41 @@ import (
...
+
+	certPath = filepath.Join(sslFilesPath, "server.crt")
+	if err := os.WriteFile(certPath, decodedCert, 0600); err != nil {
+		return "", "", fmt.Errorf("failed to write SSL certificate: %w", err)
+	}
+
+	keyPath = filepath.Join(sslFilesPath, "server.key")
+	if err := os.WriteFile(keyPath, decodedKey, 0600); err != nil {
+		return "", "", fmt.Errorf("failed to write SSL private key: %w", err)
+	}
```
sabaini, line 52: For both key/cert suggest to make writing atomic so that a partial failure doesn't leave inconsistent state behind. E.g. write both key/cert to .tmp files, and only once they're written move in place.

microceph/cmd/microceph/certificate_set_rgw.go
```diff
@@ -0,0 +1,74 @@
...
+		RunE: c.Run,
+	}
+
+	cmd.Flags().StringVar(&c.flagSSLCertificate, "ssl-certificate", "", "base64 encoded SSL certificate")
+	cmd.Flags().StringVar(&c.flagSSLPrivateKey, "ssl-private-key", "", "base64 encoded SSL private key")
+	cmd.Flags().StringVar(&c.flagTarget, "target", "", "Server hostname (default: this server)")
+	cmd.Flags().BoolVar(&c.flagRestart, "restart", false, "Restart the RGW service for immediate certificate pickup")
+
+	cmd.MarkFlagRequired("ssl-certificate")
+	cmd.MarkFlagRequired("ssl-private-key")
```
sabaini, line 42: I'm not sure but would MarkFlagRequired() reject an empty arg? Poss. could do some input validation to prevent accidentally setting invalid data

microceph/cmd/microceph/certificate_set_rgw.go
```diff
@@ -0,0 +1,83 @@
...
+		return err
+	}
+
+	// Validate that inputs are valid base64 before sending to the server.
+	if _, err := base64.StdEncoding.DecodeString(c.flagSSLCertificate); err != nil {
+		return fmt.Errorf("failed to decode SSL certificate: %w", err)
+	}
+	if _, err := base64.StdEncoding.DecodeString(c.flagSSLPrivateKey); err != nil {
+		return fmt.Errorf("failed to decode SSL private key: %w", err)
+	}
```
sabaini: Minor nit: afaik an empty string would pass this test 

```
peter@pirx ~ » echo "" | base64 -d
peter@pirx ~ » echo $?
0
```

johnramsden: https://github.com/canonical/microceph/pull/697/commits/bab480038322c78dd23e46cbc2e301aee9cd7e5e

PR #719, opened 2026-04-27
doc: Make links accessible and reliable, replace Sphinx `:doc:` target with `:ref:` role

skoech:
# Description

We need to use links more effectively and accessibly. This PR makes links in the MicroCeph docs accessible and reliable by making them descriptive, and by using the Sphinx `:ref:` role for cross-referencing to make them more resilient and future-proof. It:

- **Make all links descriptive**, i.e. they tell our users exactly where they’ll go if they click on them. They now make sense when taken out of context of the surrounding text, such that people using screen readers can make sense of them. It gets rid of ambiguous links like, “click here” and “read more”.
- **Replaces the current internal cross-referencing implementation** (Sphinx `:doc:` target) with the rST `:ref:` role. The Sphinx `:doc:` target which is fragile and will break when the target is changed (changing the filename, moving or deleting the target page) will break links
- **Gets rid of naked URLs without anchor text**
- **Adds links to all places that ought to have links**, e.g. to upstream or other docs, and other external documentation.
- **Adds intersphinx references to external documentation**, i.e. adds `sphinx.ext.intersphinx` configuration and converts existing external named references to explicit intersphinx (not fallback) roles across the docs.
  - `intersphinx-mapping` config added for for five external projects: upstream Ceph docs, Snapcraft docs, MicroCloud docs, Juju docs, and Ubuntu release notes.
  - Set `intersphinx_disabled_reftypes` to disable automatic fallback resolution; all intersphinx references are written explicitly as `:external+key:role:` (i.e. `:external+key:doc:` or `:external+key:ref:` roles, depending on whether it links to a whole page, identified by its file path - without the `.rst` extension, or it links to a specific section within a page, identified by an explicit `.. _label-name:` anchor placed before a heading in the source rST. Either using `std:doc` vs. `:std:ref` keys). 

## Type of change

## How has this been tested?

I've built the documentation locally and made sure that all changes are implemented as intended; I've also checked for spelling and done a link check, and they all pass. 

## Contributor checklist


docs/snap/explanation/index.rst
```diff
@@ -56,9 +58,9 @@ cryptographic technology used in MicroCeph.
    MicroCeph architecture <microceph-architecture>
    Security in MicroCeph <security/security-overview>
 
-If you have a specific goal, but are already familiar with MicroCeph, our :doc:`how-to guides <../how-to/index>` have more in-depth detail and instructions.
+If you have a specific goal, and are already familiar with MicroCeph, our :ref:`how-to guides <snap-how-to>` have more in-depth detail and instructions.
 
-Take a look at our :doc:`reference <../reference/index>` section when you need to know which MicroCeph commands to use.
+Take a look at our :ref:`reference section <reference>` when you need to know which MicroCeph commands to use.
```
sabaini: Hm, do we actually have a ref to `<reference>` section? I see labels being added elsewhere but not on reference/index.rst?

docs/snap/explanation/security/secure-deployment.rst
```diff
@@ -40,10 +41,10 @@ Auditing and centralized logging
 
 * Enable Auditing:  
 
-  * Configure Ceph logging levels via Ceph configuration options
-    (e.g., log_to_file \= true, debug_mon, debug_osd). Check MicroCeph
-    documentation for how to set these. Ceph logs are found in /var/snap/microceph/common/logs/ceph/.  
-  * microcephd logs to /var/log/syslog, see the MicroCeph documentation for details on setting log levels.  
+  * Configure Ceph logging levels via Ceph configuration options,
+    (e.g., ``log_to_file \= true``, ``debug_mon``, ``debug_osd``). See :ref:`changing the log level <change-log-level>`
```
sabaini: Nit: think inside the backticks you'd need to remove the backslash, as the backticks do literal rendering

docs/snap/how-to/configure-network-keys.rst
```diff
@@ -1,12 +1,13 @@
...
+
+Configuring cluster network
 ============================
 
 If you configure a cluster network, OSDs will route heartbeat, object replication and recovery traffic over the cluster network. This may improve performance compared to using a single network. 
 
 The MicroCeph cluster configuration CLI supports setting, getting, resetting and listing supported config keys mentioned below.
 
-.. list-table:: Supported Config Keys
+.. list-table:: Supported config Keys
```
sabaini: Nit: should probably be "Supported config keys"

sabaini (changes requested):
cheers, good stuff @skoech 
left some nits inline

skoech:
Ack, thanks @sabaini.

docs/snap/explanation/index.rst
skoech: I see that it's there locally. Must have added it a while back when testing the :ref: role functionality.

docs/snap/explanation/security/secure-deployment.rst
skoech: Ah, good catch. Fixed.

docs/snap/explanation/index.rst
skoech: I see that we have it already. Must have added it a while back when testing the :ref: role functionality.

skoech:
@sabaini, I've updated the description to reflect the new intersphinx changes.

docs/snap/explanation/cluster-maintenance.rst
```diff
@@ -1,17 +1,15 @@
...
---------
-
 Cluster maintenance is important for keeping the Ceph Storage Cluster at a healthy state.
 
 MicroCeph provides a simple and consistent workflow to support maintenance activity. Before
 executing any high-risk maintenance operations on a node, operators are strongly recommended to
 enable maintenance mode to minimise the impact and ensure system stability. For more information on how
-to enable maintenance mode in MicroCeph, please refer to :doc:`Perform cluster
-maintenance</snap/how-to/perform-cluster-maintenance>`.
+to enable maintenance mode in MicroCeph, please refer to our:ref:`guide on performing cluster
```
sabaini: nit: add whitespace: `our :ref:` otherwise this might not render right

docs/snap/tutorial/get-started.rst
```diff
@@ -327,7 +329,7 @@ Next steps
 
 You have deployed a healthy Ceph cluster on a single-node and enabled RGW on it. Even better, you have consumed the storage in that cluster by creating a bucket and storing an image object in it. Curious to see what else you can do with MicroCeph?
 
-See our :doc:`how-to guides <../how-to/index>`, packed with instructions to help you achieve specific goals with MicroCeph.
+See our :ref:`how-to guides <snap-how-to>`, packed with instructions to help you achieve specific goals with MicroCeph.
 
-Or, explore our :doc:`Explanation <../explanation/index>` and
-:doc:`Reference <../reference/index>` sections for additional information and quick references.
+Or, explore our :ref:`Explanation <snap-explanation>` and
+:ref:`Reference <snap-reference>` sections for additional information and quick references.
```
sabaini, line 335: Hm, `snap-reference` seems to imply a target for the generic reference page but actually this target is now in the commands/index.rst -- is that intended?

sabaini (changes requested):
Heya @skoech two nits but otherwise lgtm 🙏

docs/snap/tutorial/get-started.rst
skoech: Ah, right, @sabaini. I've renamed the CLI commands page label as `snap-cli-reference` and the reference landing page with `snap-reference`. (Also rephrased sentence in `explanation/index.rst` to correctly point to `reference/index.rst` with `snap-reference`.)

sabaini (approved):
lgtm!

PR #730, opened 2026-05-08
ci: add auto doc tag workflow

UtkarshBhatthere:
## Summary
- Adds `.github/workflows/auto-doc-tag.yml`. Cron every 6h plus `workflow_dispatch` (with `dry-run` input).
- Bumps `vX.Y.Z+doc` based on commits since the last `+doc` tag: `^feat(scope)?:` → minor, else patch.
- Scope-gated: only fires when `docs/` has commits since the previous tag. No-op otherwise.
- Idempotent: skips if HEAD already has a `v*+doc` tag.
- Triggers Read the Docs build via the existing tag-push webhook.

## Why
Today's stable docs publish path is manual (tag + RTD dashboard activate). This automates the tag step so each docs change on `main` produces a fresh stable version on RTD without operator intervention.

## Tag scheme
- Format: `v<major>.<minor>.<patch>+doc`
- First tag (already pushed manually): `v20.1.0+doc` → `0007d2c`
- Bump rule: `feat:` since last tag → minor; otherwise patch. No automatic major bumps.

## Test plan
- [ ] Run via `workflow_dispatch` with `dry-run=true` after merge; confirm summary shows expected next tag and no push.
- [ ] Make a trivial `docs/` commit on main; trigger workflow manually with `dry-run=false`; confirm tag pushed and RTD build kicks off.
- [ ] Trigger again immediately on same HEAD; confirm idempotency guard skips.
- [ ] Land a `feat:` commit and a `docs:` commit; confirm minor bump.

### DIFF
--- a/.github/workflows/auto-doc-tag.yml
+++ b/.github/workflows/auto-doc-tag.yml
@@ -1,0 +1,120 @@
+name: Auto Doc Tag
+
+# Cron every 6h. Inspect commits on main since latest vX.Y.Z+doc tag.
+# If any commit since tag touches docs/, create new tag at main HEAD:
+#   feat: in any subject since tag -> bump minor
+#   otherwise                      -> bump patch
+# No docs/ changes since tag -> exit.
+
+on:
+  schedule:
+    - cron: "0 */6 * * *"
+  workflow_dispatch:
+    inputs:
+      dry-run:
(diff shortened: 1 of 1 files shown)


UtkarshBhatthere:
converting to draft as it needs changes.

Issue #610, opened 2025-08-14
Troubles creating subvolume group in a cephfs mirroring environment


UtkarshBhatthere:
I have tried to reproduce this in a fresh and basic environment but it worked fine, so it probably has to do with cephfs-mirroring being enabled on the volume. I will add more information if this occurs again.

## Reprooduce
1. Install 2 node (primary-secondary) clusters.
2. Create volume "vol" on both sites
3. Enable `cephfs-mirror` daemon primary site.
4. Enable "mirroring" mgr module and enable snapshot mirroring on "vol"
5. Mount "vol" on primary and create multiple directories.
6. Enable snapshot mirroring for the said directories.
7. Create a subvolume group for volume "vol"
--> FAILURE
```
# ceph fs subvolumegroup create vol openstack
Error EINVAL: Traceback (most recent call last):
  File "/usr/share/ceph/mgr/mgr_module.py", line 1908, in _handle_command
    return self.handle_command(inbuf, cmd)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/ceph/mgr/volumes/module.py", line 653, in handle_command
    return handler(inbuf, cmd)
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/share/ceph/mgr/volumes/module.py", line 38, in wrap
    return f(self, inbuf, cmd)
           ^^^^^^^^^^^^^^^^^^^
  File "/usr/share/ceph/mgr/volumes/module.py", line 687, in _cmd_fs_subvolumegroup_create
    return self.vc.create_subvolume_group(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/ceph/mgr/volumes/fs/volume.py", line 1018, in create_subvolume_group
    set_group_attrs(fs_handle, group.path, attrs)
  File "/usr/share/ceph/mgr/volumes/fs/operations/group.py", line 190, in set_group_attrs
    pool = get_ancestor_xattr(fs, path, "ceph.dir.layout.pool")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/share/ceph/mgr/volumes/fs/fs_util.py", line 199, in get_ancestor_xattr
    return fs.getxattr(path, attr).decode('utf-8')
           ^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #732, opened 2026-05-13
ci: Add robot framework structure

johnramsden:
# Description

Migrate MicroCeph's CI test suite from bash + GitHub Actions to the Robot Framework. The new tests must be runnable locally without any GitHub Actions dependency.

See for context:

* [MAPPING.md](https://github.com/user-attachments/files/28408456/MAPPING.md)
* [MIGRATION-LINE-BY-LINE.md](https://github.com/user-attachments/files/28408457/MIGRATION-LINE-BY-LINE.md)

## Type of change

## Contributor checklist

Please check that you have:


sabaini:
Thanks @johnramsden, good stuff! I'm still reading through the PR, its big :-) One high-level comment is that I'd be wary of mixing too much bash into the robot test cases, that subverts the nice readability properties of robot tests

So e.g. having `Run In Container` can be ok for small bits but I wonder if `Run In Container    node-wrk0    microceph.ceph -s | grep "mon: 1 daemons, quorum node-wrk0"` wouldn't be better with a cmd `Should Have One Mon` or equivalent. Similarly could replace some of the poll loops with `Poll For Ceph Health` (for example) cmd.

johnramsden:
@sabaini I completely agree. I tried to architect it in a way where re-used behavior was put into modules, however the migration was large and some of the resulting tests ended up less readable than I would have hoped. 

I can take another pass at it and try to make some of the more complex tests more readable

tests/robot/resources/streaming_process.py
```diff
@@ -0,0 +1,49 @@
...
+    # Robot Framework redirects sys.stdout to its log buffer; write to the
+    # original stdout so lines appear on the console as they are produced.
+    console = sys.__stdout__ or sys.stdout
+    lines = []
+    try:
+        for line in proc.stdout:
+            console.write(line)
+            console.flush()
+            lines.append(line)
+        proc.wait(timeout=timeout_int)
```
sabaini, line 78: afaik proc.stdout would only finish if the underlying process finished, so to this .wait(timeout) doesn't really protect from a hanging subprocess. not a blocker but maybe worth mentioning in the comments

tests/robot/resources/microceph_harness.resource
```diff
@@ -0,0 +1,1121 @@
...
+    ...    msg=Expected failure but command succeeded: ${bash_cmd}
+    RETURN    ${result}
+
+Run In Container
+    [Documentation]    Runs cmd inside an inner LXD container via the outer VM.
+    ...    bash -eo pipefail: mirrors set -e semantics so any failing command or pipe stage
+    ...    inside the container fails the keyword immediately.
+    [Arguments]    ${container}    ${cmd}    ${timeout}=300
+    Log To Console    [${container}] ${cmd[:80]}
+    Run In VM And Check    lxc exec ${container} -- bash -eo pipefail -c "${cmd}"    ${timeout}
```
sabaini: bit wary of the nested quoting -- what if `$cmd` already has double quotes? suggest to write a temp. file script and execute it

tests/robot/robot.py
```diff
@@ -0,0 +1,65 @@
...
+
+    # Resolve the target path (always required by robot).
+    here = os.path.dirname(__file__)
+    if args.all:
+        target = here
+    elif args.test_suite:
+        target = os.path.join(here, args.test_suite)
+    else:
+        print(
+            "WARNING: no --test-suite or --all given; running stub_test.robot (no real tests).",
```
sabaini: nit: suggest to err on the side of safety and run a comprehensive test suite by default instead of stub_test (e.g. run `--all`)

tests/robot/README.md
```diff
@@ -0,0 +1,50 @@
+# MicroCeph Robot Framework Tests
```
sabaini, line 1: eventually could link to this README.md from top level AGENTS.md :-)

sabaini (changes requested):
Hey @johnramsden some more comments. Yes, factoring out more robot keywords would be nice!

johnramsden:
Ready for review again. All tests are passing locally, additionally made some changes to clean tests up and make them more modular.

tests/robot/wiping-test/wiping_tests.robot
```diff
@@ -0,0 +1,41 @@
...
+    [Documentation]    Full pristine-check cycle inside the outer VM via
+    ...    actionutils.sh verify_pristine_check: creates loop-backed dirty disks,
+    ...    removes and reinstalls microceph, verifies dirty disks are rejected without
+    ...    --wipe, then adds them successfully with --wipe.
+    ...
+    ...    Runs as a single command so snap remove/reinstall are not subject to
+    ...    per-step Robot timeouts; streaming shows progress in real time.
+    [Tags]    disk-management    pristine
+    ${rc}    ${out}=    Run Streaming Process
+    ...    lxc exec ${OUTER_VM} -- bash -c "~/actionutils.sh verify_pristine_check"
```
sabaini: hm, does this work with XTRACE? aiui this would become then `bash -x lxc exec ... ` 
would probably need to be with bash -c -x

tests/robot/resources/microceph_harness.resource
```diff
@@ -0,0 +1,1353 @@
...
+    Run In VM And Check    sudo snap start microceph    60
+    Wait For RGW    1
+    Run In VM And Check    grep -q "run dir = /var/snap/microceph/current/run" /var/snap/microceph/current/conf/radosgw.conf    10
+
+# ---------------------------------------------------------------------------
+# Teardown
+# ---------------------------------------------------------------------------
+
+Collect MicroCeph Diagnostics
+    [Documentation]    Collects diagnostics from the outer VM; errors are ignored.
```
sabaini: for multi node tests we'd want diagnostics from within the nodes too

sabaini (changes requested):
Some drive-by comments. 
Also tests.yml has some duplication going on, maybe we could factor out some of the checkout/lxd setup/pip steps?

tests/robot/resources/microceph_harness.resource
johnramsden: https://github.com/canonical/microceph/pull/732/commits/7392a90ae70c038a09612ba1ee61d8a9219f96a0

tests/robot/wiping-test/wiping_tests.robot
johnramsden: https://github.com/canonical/microceph/pull/732/commits/5f1026d4b30a55f00b5f9fb3517c27607942fcca

.github/workflows/checks.yml
```diff
@@ -0,0 +1,80 @@
...
+          sudo apt-get update
+          sudo apt-get install -y libdqlite1.17-dev
+
+      - name: Install Go (1.25)
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.25.x
+
+      - name: Install Robot Framework
+        run: pip3 install -r tests/robot/requirements.txt
```
sabaini: nit: it would be great if we could avoid installing into the system python esp. if we want to have these tests run on a local dev machine too. maybe use a local venv managed via uv?

tests/robot/rbd-replication-test/rbd_replication_tests.robot
```diff
@@ -0,0 +1,191 @@
...
+            IF    ${i} == 119
+                Fail    Timed out waiting for ${pool} mirror health on ${node} (last=${health})
+            END
+            Sleep    5s
+        END
+    END
+
+Disable RBD Mirroring
+    [Documentation]    Disables mirroring on pool_two images and both pools from siteb.
+    ...    Mirror health is UNKNOWN after failover; the disable ops work regardless so we skip the health wait.
```
sabaini: iiuc it checks for `OK` currently does this not work anymore?

tests/robot/cephfs-replication-test/cephfs_replication_tests.robot
```diff
@@ -0,0 +1,155 @@
...
+    Run In Container    node-wrk2    sudo microceph.ceph fs volume create vol    120
+    Run In Container    node-wrk0    sudo microceph replication enable cephfs --volume vol --dir-path /dir1/ --remote siteb    120
+    Run In Container    node-wrk0    sudo microceph replication enable cephfs --volume vol --dir-path /dir2/ --remote siteb    120
+    Log To Console    [cephfs] Installing ceph-common and mounting primary filesystem...
+    Mount CephFS From Container    node-wrk0    /mnt/primary    vol
+    Run In VM And Check    sudo mkdir -p /mnt/primary/dir1 /mnt/primary/dir2    10
+    Run In VM And Check    echo ${STR1} | sudo tee /mnt/primary/dir1/test_file    10
+    Run In VM And Check    echo ${STR2} | sudo tee /mnt/primary/dir2/test_file    10
+
+Verify CephFS Mirror List Output
```
sabaini, line 33: hm, this is a qite long function and feel readability suffers 
i wonder if this could be moved to python for better readability?

tests/robot/cephfs-replication-test/cephfs_replication_tests.robot
```diff
@@ -0,0 +1,155 @@
...
+    [Documentation]    Creates subvolumes, adds them to the mirror list, and verifies the list output.
+    ...    Both subvolumegroup and subvolume creation are attempted but not required — some Ceph
+    ...    builds have a Python binding bug (TypeError in make_ex / Charset.__init__) that causes
+    ...    these operations to fail even when the MDS is healthy. The list output will still be
+    ...    non-empty from the directory paths configured in Configure CephFS Mirroring.
```
sabaini: if you have a bug / tracker issue for these type errors at hand pls put them here for future tracking

tests/robot/resources/microceph_harness.resource
```diff
@@ -0,0 +1,1383 @@
...
+    IF    not ${restored}    Fail    Not all services re-enabled after 30s (expected ${count_before})
+    # Verify each previously enabled+active service is back, by name.
+    FOR    ${svc}    IN    @{services_before}
+        ${svc_state}=    Run In VM    snap services ${svc}    30
+        Should Contain    ${svc_state.stdout}    enabled    msg=Service ${svc} not enabled after re-enable
+        Should Contain    ${svc_state.stdout}    active    msg=Service ${svc} not active after re-enable
+    END
+    Verify Cluster Health
+
+Test SSL Certificate Rotation
```
sabaini, line 884: from a length/readability perspective think this might be better in python

sabaini (changes requested):
Hey some more review comments/questions

As a general observation I love robot if tests are a sequence of "do this, then check, then do that, ..."
I love it less with more complex logic (eg nested loops, data manipulation), wonder if more of these should be moved to python

tests/robot/cephfs-replication-test/cephfs_replication_tests.robot
johnramsden: #610, will add comment

tests/robot/rbd-replication-test/rbd_replication_tests.robot
johnramsden: Thanks this was a good catch, the issue was an incorrectly quoted grep

tests/robot/resources/microceph_harness.resource
johnramsden: Test SSL Certificate Rotation? It's mostly a list of commands run with no loops or logic. I think this is allright personally.

If you mean 'Test Snap Disable Enable' can you elaborate on what you do not like about it? Problem is if we move it to python completely then when we are calling external functions we have to use `BuiltIn().run_keyword(...)`

ie: 

```python
bi.run_keyword("Run In VM And Check", "sudo snap disable microceph", 60) 
```

Which can end up being a fair bit of boilerplate. 

We can do so, but I'm more curious what exactly you dislike here. 

I could extract out part of it such as the awk... (https://github.com/canonical/microceph/pull/732/commits/071a71c8a1d4e0363350b1c5416a301c3e83d9e6)?

tests/robot/cephfs-replication-test/cephfs_replication_tests.robot
johnramsden: https://github.com/canonical/microceph/pull/732/commits/0356f3011d5e25b7045b9ab2787c325d30b58fc9

tests/robot/cephfs-replication-test/cephfs_replication_tests.robot
johnramsden: https://github.com/canonical/microceph/pull/732/commits/0356f3011d5e25b7045b9ab2787c325d30b58fc9

tests/robot/rbd-replication-test/rbd_replication_tests.robot
johnramsden: https://github.com/canonical/microceph/pull/732/commits/073ec0c979b1583886c8c389034fd0659ed7e91a

.github/workflows/tests.yml
```diff
@@ -1203,259 +173,38 @@ jobs:
...
-    runs-on: ubuntu-24.04
-    needs: build-microceph
-    steps:
-      - name: Download snap
-        uses: actions/download-artifact@v4
-        with:
-          name: snaps
-          path: /home/runner
+      - name: Run Robot suite
+        run: tox -e robot -- --test-suite ${{ matrix.suite }} --snap-path '/home/runner/*.snap' ${{ matrix.args }}
```
UtkarshBhatthere, line 199: **HIGH - two Robot suites are never executed in CI.**

This job always pins to one named subdirectory via `--test-suite ${{ matrix.suite }}`, and the full-tree fallback in `tests/robot/robot.py:43-47` only fires when `--test-suite` is *omitted* (no job omits it). The matrix (lines 57-163) contains `multi-node-tests` and `multi-node-tests-with-custom-microceph-ip`, but **neither the bare `single-node` nor the bare `multi-node` directory is in the matrix**, so `tests/robot/single-node/basic_tests.robot` and `tests/robot/multi-node/basic_tests.robot` never run.

Those suites contain distinct scenarios (`waitready` lifecycle: must fail pre-bootstrap then succeed post-bootstrap; the Orchestrator module; a bare 4-node bootstrap asserting `osd: 3 osds` at `multi-node/basic_tests.robot:28`). As wired they contribute zero coverage.

**Fix:** add `suite: single-node` / `suite: multi-node` matrix entries, or delete both directories if they are superseded by `single-system-tests` / `multi-node-tests`.

.github/workflows/tests.yml
```diff
@@ -47,1150 +47,120 @@ jobs:
...
-          ~/actionutils.sh install_and_bootstrap_microceph
-
-      - name: Add loopback file OSDs
-        run: |
-          set -uex
-          sudo microceph disk add loop,1G,3
-          ~/actionutils.sh wait_for_osds 3
-          sudo microceph.ceph -s
+  robot-tests:
+    name: ${{ matrix.name }}
```
UtkarshBhatthere, line 51: **HIGH - the required status-check name `Tests / DSL functional tests` no longer exists.**

The matrix refactor replaced the original `dsl-functional-tests` job (`name: DSL functional tests` on `main`, reported by GitHub as `Tests / DSL functional tests`) with a single `robot-tests` job using `name: ${{ matrix.name }}`. The new check names are `Tests / DSL baseline tests`, `Tests / DSL WAL-DB validation tests`, etc. The old exact string disappears.

If branch protection requires `Tests / DSL functional tests`, merges become either unguarded (the gate silently never reports) or unblockable. The DSL tests still *run* (now more granular) -- only the **gate name** breaks.

**Fix:** audit the branch-protection required-status-check list and replace the old name with the new per-matrix check names as part of merging.

tests/robot/multi-node-tests/multi_node_tests.robot
```diff
@@ -0,0 +1,207 @@
...
+    Verify Ceph Config Has Public Network
+
+Enable Services On Head Node For
+    [Documentation]    Enables mon/mds/mgr on a target node, running commands from node-wrk0.
+    [Arguments]    ${node}
+    Log To Console    [cluster] Enabling mon/mds/mgr on ${node} via node-wrk0...
+    Run In Container    node-wrk0    microceph enable mon --target ${node}    120
+    Run In Container    node-wrk0    microceph enable mds --target ${node}    120
+    Run In Container    node-wrk0    microceph enable mgr --target ${node}    120
+    FOR    ${i}    IN RANGE    8
```
UtkarshBhatthere, line 24: **MEDIUM - poll loop has no `Fail` after the loop.**

This `FOR ... IN RANGE 8` loop `BREAK`s when the node joins the mon list but does nothing if it never does -- it just falls through to `microceph.ceph -s` on line 29. This is **not** a false-green (the subsequent `enable rgw --target` will fail), but a slow `node-wrk1` join surfaces as a confusing downstream RGW error instead of a clear message.

**Fix:** mirror `Wait For N Nodes In Cluster` (`microceph_harness.resource:1185`) by adding `Fail    ${node} did not appear in mon list after 16 s` after the `END`. (Same pattern recurs in `Remove Node Head Node` at lines 90-94.)

tests/robot/resources/microceph_harness.resource
```diff
@@ -0,0 +1,1387 @@
...
+    ${result}=    Run In VM    lxc exec ${node} -- sh -c "sudo microceph.rbd mirror pool status ${pool} --verbose | sed -n 's/^health: //p' | head -1"    30
+    ${health}=    Set Variable    ${result.stdout.strip()}
+    IF    "${health}" == "${EMPTY}"    RETURN    UNKNOWN
+    RETURN    ${health}
+
+Wait For CephFS Replication List Non Empty
+    [Documentation]    Polls until the CephFS replication list for ${vol} on ${node} is non-empty.
+    [Arguments]    ${node}    ${vol}    ${attempts}=50
+    FOR    ${i}    IN RANGE    ${attempts}
+        ${empty}=    Run In VM    lxc exec ${node} -- sh -c "sudo microceph replication list cephfs --json | jq '.${vol} | . == {}'"    30
```
UtkarshBhatthere: **MEDIUM - `jq '.${vol} | . == {}'` returns `false` for an absent key, causing an early success-`RETURN`.**

The loop `RETURN`s as soon as stdout is `"false"` (line 1222). But `jq '.${vol} | . == {}'` evaluates to `false` in **two** cases: the key exists and is a non-empty object (intended), *and* the key is absent -- because `null == {}` is `false`. So a missing or misspelled `${vol}` returns immediately as if replication succeeded, rather than polling and eventually failing.

The keyword does have a timeout `Fail` (line 1223), so a genuinely-empty list still fails correctly -- the defect is specifically the absent-key path returning success.

**Fix:** make the predicate explicit, e.g. `jq -e '(.${vol} // empty) | (. != null and . != {})'` and gate the `RETURN` on rc==0 (or `jq '.${vol} | type == "object" and length > 0'`).

UtkarshBhatthere (commented):
Automated multi-agent review (35 raw findings, adversarially verified -> 17 confirmed; LOW/nit items omitted here). Overall this is a solid migration and in several places more defensive than the bash it replaces, with **no confirmed false-greens**. Two HIGH and two MEDIUM findings below are worth addressing before merge.

_Generated with Claude Code._

.github/workflows/tests.yml
johnramsden: @UtkarshBhatthere I don't seem to have access to check this, do you?

johnramsden: I believe this basically just requires checking the current check rules and making sure they're correct before we merge

.github/workflows/tests.yml
johnramsden: Dropped (https://github.com/canonical/microceph/pull/732/commits/afd5ada8ede341203570b6551a78bbcf03d494a2) legacy

tests/robot/multi-node-tests/multi_node_tests.robot
johnramsden: Fail on last https://github.com/canonical/microceph/pull/732/commits/2eb8587f9f03edfb8b80cb867ebe38f6f88144d0

tests/robot/resources/microceph_harness.resource
johnramsden: Etracted this into python and handled the condition https://github.com/canonical/microceph/pull/732/commits/3a7147179cc619c50ad5f2e88e8a4eb65a27f7ef

tests/robot/resources/microceph_harness.resource
sabaini: Yes, apologies, meant `Test Snap Disable Enable` above. I think it's doing quite a bit of logic and robot isn't great at that imho.

Thinking of something along these lines (pseudo code b/c i don't actually know robot, not sure if possible):

```
Test Snap Disable Enable
    [Documentation]    Tests that snap disable/enable re-enables all services.
    ...    Records the names of services that were enabled+active before disable and verifies
    ...    each one by name is restored after re-enable (not just the aggregate count), and
    ...    fails on timeout. Mirrors bash test_snap_disable_enable + check_snap_service_active_enabled.
    Log To Console    [snap] Testing snap disable/enable service restoration...

    ${services_before}=    Get Enabled Active MicroCeph Snap Services

    Disable MicroCeph Snap
    Enable MicroCeph Snap

    Wait Until MicroCeph Snap Services Restored    ${services_before}    timeout=30s
    Verify Cluster Health
```


I wonder if it would help if those infra commands like `Run In VM And Check` etc. were pushed to Python. IMHO the strength of robot is it gives a nice UI, otoh the `Run ...` family of commands is fairly low level, moving them to Python wouldn't take away from the UI benefits Robot has

sabaini (approved):
Hey @johnramsden, as discussed I'm +1 on this -- there's some opportunities to refactor such that lower level test helpers are more pythonic but these can go into a separate PR, and the current PR is already a big improvement over the current bash spaghetti

PR #742, opened 2026-05-27
chore(security): bump Go to 1.25.10, go-jose v4.1.4, add govulncheck CI

### LINKED ISSUE #745: Patch Go stdlib + go-jose CVEs reachable from microceph (Go 1.25.7 → 1.25.10, go-jose v4.1.3 → v4.1.4)
## Problem

`govulncheck ./...` against `microceph/go.mod` (Go 1.25.7, go-jose v4.1.3) reports **13 called-path vulnerabilities** — 10 HIGH, 2 MEDIUM, 1 LOW — across the Go standard library and `github.com/go-jose/go-jose/v4`.

### Go stdlib (fixed in 1.25.10)

Affecting `net`, `crypto/x509`, `crypto/tls`, `net/http` (http2), `archive/tar`, `os`, `net/url`. Reachable from microceph code paths (CLI flag parsing, snap-side HTTP, TLS handshake to the dqlite/LXD bus).

### go-jose (fixed in v4.1.4)

- **GO-2026-4945 / CVE-2026-34986** — JWE decryption panic in `github.com/go-jose/go-jose/v4`. Called via the LXD / dqlite auth path.

### Remaining (not addressed here)

4 called-path findings remain after the bumps — all in `github.com/canonical/lxd`, no upstream fix released yet:

| ID | Module |
|----|--------|
| GO-2026-4595 | canonical/lxd |
| GO-2025-4121 | canonical/lxd |
| GO-2025-4003 | canonical/lxd |
| GO-2025-3999 | canonical/lxd |

These need an upstream LXD release; separate follow-up.

## Fix

PR #742 — bumps the Go directive to 1.25.10 and `go-jose/v4` to v4.1.4, and adds a govulncheck CI job (PR-diff gate + daily cron on `main`) so net-new called-path findings get caught early.

After the bumps: **13 → 4 called-path findings** (the 4 LXD-only ones above).

UtkarshBhatthere:
## Summary

- Bumps Go directive `microceph/go.mod` from **1.25.7 → 1.25.10**, clearing 8 Go stdlib vulnerabilities (`net`, `crypto/x509`, `crypto/tls`, `net/http` http2, `archive/tar`, `os`, `net/url`) reachable from microceph code paths.
- Upgrades `github.com/go-jose/go-jose/v4` from **v4.1.3 → v4.1.4**, clearing GO-2026-4945 / CVE-2026-34986 (JWE decryption panic).
- Wires govulncheck into CI via a **reusable composite action** + **two thin workflows** (PR-diff gating, daily scheduled full scan).

## govulncheck delta

Before: **13 called-path vulnerabilities** (10 HIGH, 2 MEDIUM/MODERATE, 1 LOW).
After: **4 called-path vulnerabilities**, all in `github.com/canonical/lxd` (no upstream fix available):

| ID | Module | Status |
|----|--------|--------|
| GO-2026-4595 | canonical/lxd | no fix upstream |
| GO-2025-4121 | canonical/lxd | no fix upstream |
| GO-2025-4003 | canonical/lxd | no fix upstream |
| GO-2025-3999 | canonical/lxd | no fix upstream |

These four require an upstream LXD release. Filing/tracking is a follow-up.

## CI design

`.github/actions/govulncheck` — composite action. Installs Go + libdqlite (cgo), runs `govulncheck -format json ./...`, emits the JSON output and a count of called-path findings. `fail-on-vuln` input lets callers either gate (cron) or collect output silently (PR diff).

`.github/workflows/vuln-scan-pr.yml` — PR-time job. Paths-filtered on `microceph/go.{mod,sum}` and the action itself: GitHub skips the workflow entirely when no dep change is in the PR. When it does run, it scans both PR HEAD and base ref via the composite action and diffs the called-path OSV ID sets — failing **only on net-new findings**. Existing unfixed LXD vulns therefore don't block unrelated PRs.

`.github/workflows/vuln-scan-cron.yml` — daily 06:00 UTC scan of the full module on `main`, plus `workflow_dispatch`. On any called-path finding it opens (or comments on) a single tracking issue titled `[vuln-scan] govulncheck found called-path vulnerabilities` and auto-closes it when the next run is clean.

The earlier ad-hoc `vuln-scan` job that lived in `tests.yml` is removed in favour of the dedicated workflows above.

## Test plan
- [x] `govulncheck ./...` rerun locally — 13 → 4 called vulns, all remaining are LXD-only with no fix
- [x] `go build ./...` clean
- [x] `go test -short ./...` passes (api, ceph, cmd, common, dsl, logger all OK)
- [x] jq extractor validated against real `govulncheck -format json` output (returns exactly the 4 LXD IDs, no false positives from module-only finding summaries)
- [ ] CI `vuln-scan-pr` job: paths-filter triggers it (go.{mod,sum} touched in this PR); diff vs main should report only the unchanged 4 LXD vulns and pass (no net-new)
- [ ] CI `vuln-scan-cron` job: `workflow_dispatch` trigger from `main` after merge should fail and open the tracking issue
- [ ] CI `static-checks` + `unit-tests` green on this PR

### DIFF
--- a/.github/actions/govulncheck/action.yml
+++ b/.github/actions/govulncheck/action.yml
@@ -1,0 +1,102 @@
+name: govulncheck
+description: Run govulncheck against a Go module and emit JSON results.
+
+inputs:
+  working-directory:
+    description: Directory containing the Go module to scan.
+    required: false
+    default: microceph
+  go-version:
+    description: Go toolchain version to install (passed to actions/setup-go).
+    required: false
+    default: 1.25.x
+  output-file:
+    description: Path (relative to working-directory) for the JSON scan output.
+    required: false
+    default: govulncheck.json
+  fail-on-vuln:
+    description: |
+      When 'true', fail the step if any called-path vulnerabilities are found.
+      When 'false', always succeed; the caller is expected to consume output-file.
+    required: false
+    default: "true"
+
+outputs:
+  output-file:
+    description: Absolute path to the JSON scan output produced by this step.
+    value: ${{ steps.scan.outputs.output-file }}
+  called-count:
+    description: Number of called-path Go vulnerability IDs found.
+    value: ${{ steps.scan.outputs.called-count }}
+  called-ids:
+    description: Newline-separated list of called-path Go vulnerability IDs.
+    value: ${{ steps.scan.outputs.called-ids }}
+
+runs:
+  using: composite
+  steps:
+    - name: Install cgo dependencies (libdqlite)
+      shell: bash
+      run: |
+        sudo add-apt-repository ppa:dqlite/dev -y --no-update
+        sudo apt-get update
+        sudo apt-get install -y libdqlite1.17-dev
+
+    - name: Install Go
+      uses: actions/setup-go@v5
+      with:
+        go-version: ${{ inputs.go-version }}
+
+    - name: Install govulncheck
+      shell: bash
+      run: go install golang.org/x/vuln/cmd/govulncheck@latest
+
+    - name: Run govulncheck
+      id: scan
+      shell: bash
+      working-directory: ${{ inputs.working-directory }}
+      env:
+        OUTPUT_FILE: ${{ inputs.output-file }}
+        FAIL_ON_VULN: ${{ inputs.fail-on-vuln }}
+      run: |
+        set -uo pipefail
+
+        # govulncheck exits 3 when vulnerabilities are found in called code,
+        # 0 when clean, non-zero/non-3 on tooling failure.
+        govulncheck -format json ./... > "${OUTPUT_FILE}"
+        gv_rc=$?
+
+        if [ "${gv_rc}" -ne 0 ] && [ "${gv_rc}" -ne 3 ]; then
+          echo "::error::govulncheck failed with exit code ${gv_rc}"
+          exit "${gv_rc}"
+        fi
+
+        # govulncheck -format json emits NDJSON (one JSON object per line).
+        # A finding has a 'called' trace when at least one trace frame carries
+        # a 'function' field — module-only summary findings are not reachable.
+        called_ids=$(jq -rs '
+          [ .[]
+            | select(.finding? and (.finding.trace[]?.function != null))
+            | .finding.osv
+          ] | unique[]
+        ' "${OUTPUT_FILE}" 2>/dev/null || true)
+        called=$(printf "%s\n" "${called_ids}" | grep -c . || true)
+
+        {
+          echo "output-file=${PWD}/${OUTPUT_FILE}"
+          echo "called-count=${called}"
+          echo "called-ids<<EOF"
+          printf "%s\n" "${called_ids}"
+          echo "EOF"
+        } >> "${GITHUB_OUTPUT}"
+
+        echo "govulncheck called-path vulnerabilities: ${called}"
+        if [ "${called}" -gt 0 ]; then
+          echo "${called_ids}"
+        fi
+
+        if [ "${FAIL_ON_VULN}" = "true" ] && [ "${called}" -gt 0 ]; then
+          echo "::error::govulncheck found ${called} called-path vulnerabilities"
+          exit 1
+        fi
+        exit 0
--- a/.github/actions/govulncheck/action.yml
+++ b/.github/actions/govulncheck/action.yml
@@ -79,7 +79,7 @@ runs:
             | select(.finding? and (.finding.trace[]?.function != null))
             | .finding.osv
           ] | unique[]
-        ' "${OUTPUT_FILE}" 2>/dev/null || true)
+        ' "${OUTPUT_FILE}")
         called=$(printf "%s\n" "${called_ids}" | grep -c . || true)
 
         {
--- a/.github/workflows/vuln-scan-cron.yml
+++ b/.github/workflows/vuln-scan-cron.yml
@@ -21,17 +21,15 @@ jobs:
 
       - name: Run govulncheck
         id: scan
-        continue-on-error: true
         uses: ./.github/actions/govulncheck
         with:
           working-directory: microceph
           output-file: govulncheck.json
-          fail-on-vuln: "true"
+          fail-on-vuln: "false"
 
       - name: Render summary
         id: render
         env:
-          OUTCOME: ${{ steps.scan.outcome }}
           CALLED_COUNT: ${{ steps.scan.outputs.called-count }}
           CALLED_IDS: ${{ steps.scan.outputs.called-ids }}
         run: |
@@ -58,7 +56,7 @@ jobs:
           cat "${body_file}" >> "${GITHUB_STEP_SUMMARY}"
 
       - name: Open or update tracking issue
-        if: steps.scan.outcome == 'failure'
+        if: steps.scan.outputs.called-count != '0' && steps.scan.outputs.called-count != ''
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           BODY_FILE: ${{ steps.render.outputs.body-file }}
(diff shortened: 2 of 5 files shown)


.github/workflows/vuln-scan-pr.yml
```diff
@@ -0,0 +1,91 @@
+name: Vuln scan (PR diff)
+
+on:
+  pull_request:
+    paths:
+      - microceph/go.mod
+      - microceph/go.sum
+      - .github/actions/govulncheck/**
+      - .github/workflows/vuln-scan-pr.yml
```
sabaini, line 10: Hm, we are only running this for dep changes. But the check will flag on vulns in called paths -- if a microceph/*.go source changes this could as well introduce new vulns if they're exercising new paths. That would result in introducing new vulns even if the deps don't actually change, right? Suggest to also trigger on source changes besides dep changes to address this

.github/workflows/vuln-scan-pr.yml
```diff
@@ -0,0 +1,91 @@
...
+    steps:
+      - name: Checkout PR head
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Scan PR head
+        id: head-scan
+        uses: ./.github/actions/govulncheck
```
sabaini, line 29: Paranoia nit: we're checking with scanner coming from the checkout we're about to scan. Suggest to use a scanner from a base checkout not from HEAD so we're checking with an already-merged scanner

.github/actions/govulncheck/action.yml
```diff
@@ -0,0 +1,102 @@
...
+
+        # govulncheck -format json emits NDJSON (one JSON object per line).
+        # A finding has a 'called' trace when at least one trace frame carries
+        # a 'function' field — module-only summary findings are not reachable.
+        called_ids=$(jq -rs '
+          [ .[]
+            | select(.finding? and (.finding.trace[]?.function != null))
+            | .finding.osv
+          ] | unique[]
+        ' "${OUTPUT_FILE}" 2>/dev/null || true)
```
sabaini: Nit: `|| true` makes us fail silent, which could result in falsely marking a PR as clean

.github/workflows/vuln-scan-cron.yml
```diff
@@ -0,0 +1,96 @@
...
+          fetch-depth: 0
+
+      - name: Run govulncheck
+        id: scan
+        continue-on-error: true
+        uses: ./.github/actions/govulncheck
+        with:
+          working-directory: microceph
+          output-file: govulncheck.json
+          fail-on-vuln: "true"
```
sabaini: Iiuc we'd mark a PR as vulnerable also in case of infra issues or tool faults?

sabaini (changes requested):
Hey, some comments inline.

.github/workflows/vuln-scan-cron.yml
UtkarshBhatthere: fail-on-vuln: basically defines whether the composite action should FAIL (from a github semantic) when vulns are found againt the go deps. In case of infra issues the CI test would come up failed which would happen irrespective of vuln check.

.github/workflows/vuln-scan-pr.yml
UtkarshBhatthere: by `scanner` do you mean the govulncheck composite action ?

.github/workflows/vuln-scan-pr.yml
UtkarshBhatthere: fair enough, let me add that.

.github/actions/govulncheck/action.yml
UtkarshBhatthere: let me refactor the checks so that the logical safety net is traceable.

.github/workflows/vuln-scan-pr.yml
sabaini: Yes that guy

.github/workflows/vuln-scan-cron.yml
sabaini: I see that we're exiting with an error if infra fails -- iiuc this would make as marking something as vulnerable then?

UtkarshBhatthere: The only time we report vulnerabilities is when a cron job finds them, scans exisiting issues and does not find any existing ones against the vuln. This will not give a false positive as if infra fails the found vulns be none.

For PR runs, if there is an infra issue we will see a failing test, this may look like a found vulnerability (false positive) but that's the same case for any CI test failure caused due to infra issues.

.github/workflows/vuln-scan-pr.yml
UtkarshBhatthere: added all source changes to the trigger.

.github/workflows/vuln-scan-cron.yml
sabaini: I'm wondering if it would make sense to separate scan status from vuln status. E.g. could set `fail-on-vuln: false` and only update the report issue if we have `called-count > 0`

UtkarshBhatthere: ack, that's better

.github/workflows/vuln-scan-pr.yml
```diff
@@ -0,0 +1,102 @@
...
+          fi
+
+          echo "### New called-path vulnerabilities" >> "${GITHUB_STEP_SUMMARY}"
+          echo '```' >> "${GITHUB_STEP_SUMMARY}"
+          echo "${new_ids}" >> "${GITHUB_STEP_SUMMARY}"
+          echo '```' >> "${GITHUB_STEP_SUMMARY}"
+
+          body="## govulncheck: new called-path vulnerabilities
+
+This PR introduces new called-path vulnerabilities vs base (\`${{ github.event.pull_request.base.ref }}\`):
```
sabaini: Uhm, is this valid yaml?

.github/workflows/vuln-scan-pr.yml
```diff
@@ -0,0 +1,102 @@
...
+
+\`\`\`
+${new_ids}
+\`\`\`
+
+Workflow run: ${GITHUB_SERVER_URL}/${GITHUB_REPOSITORY}/actions/runs/${GITHUB_RUN_ID}"
+
+          gh pr comment ${{ github.event.pull_request.number }} --body "${body}"
```
sabaini: Hm, I think we'd want to fail this workflow job if found issues -- we're adding a comment but otherwise return success now, right?

sabaini (commented):
Hey, 2 questions left

.github/workflows/vuln-scan-pr.yml
UtkarshBhatthere: Moved the `${{ }}` expressions into named env vars (`BASE_REF`, `PR_NUMBER`) — keeps GHA template syntax out of multiline bash strings entirely.

.github/workflows/vuln-scan-pr.yml
UtkarshBhatthere: Good catch — added `exit 1` after posting the comment so the job fails when new vulns are found.

.github/workflows/vuln-scan-cron.yml
```diff
@@ -0,0 +1,88 @@
...
+          if [ -n "${existing}" ]; then
+            echo "Updating issue #${existing}"
+            gh issue comment "${existing}" --body-file "${BODY_FILE}"
+          else
+            echo "Opening new issue"
+            gh issue create --title "${TITLE}" --body-file "${BODY_FILE}" --label security
+          fi
+
+      - name: Close tracking issue when clean
+        if: steps.scan.outputs.called-count == '0' || steps.scan.outputs.called-count == ''
```
sabaini, line 78: Nit: think it would be safer to only close if `called-count == '0'` -- closing on empty output could accidentally close if a tool error creeps in

sabaini (approved):
Hey @UtkarshBhatthere , one final nit but nothing to block over

.github/workflows/vuln-scan-cron.yml
UtkarshBhatthere: if we do that won't we be back to square one where an infra issue may cause the test to be red which was something you didn't want.

UtkarshBhatthere: I'll get this one through, based on the discussions we can remodify the deemed behavior again.

PR #748, opened 2026-05-29
fix: purgeOSD retries on any transient failure, not just EBUSY exit code

UtkarshBhatthere:
## Summary

- `purgeOSD` had a dead retry loop: it checked `syscall.Errno(exitError.ExitCode()) != syscall.EBUSY` to decide whether to retry, but `ceph osd purge` exits with code 1 on failure, not the errno value 16 (EBUSY). Every failure hit the "unexpected exit error" break and retried zero times.
- Fix: retry on any failure from `doPurge`, matching the existing `wipeDevice` pattern. The OSD process may still be shutting down when purge is attempted, causing transient errors.
- Add `purgeRetrySleepFunc` injectable variable (per AGENTS.md `Func` suffix convention) to allow test-time sleep suppression.
- Add `TestPurgeOSD` covering success-on-first-try, transient-failure-then-success, and retry exhaustion.

## Test Plan

- [x] `go test ./ceph/... -run TestOSD -count=1` passes
- [x] `TestPurgeOSD` covers all three cases
- [ ] Integration: `test_dsl_api_disk_hurl` OSD purge race no longer flaky

### DIFF
--- a/microceph/ceph/osd.go
+++ b/microceph/ceph/osd.go
@@ -1894,26 +1894,14 @@ func (m *OSDManager) purgeOSD(osd int64) error {
 	for i := 0; i < retries; i++ {
 		err = m.doPurge(osd)
 		if err == nil {
-			// Success: break the retry loop
 			break
 		}
-		// we're getting a RunError from common.ProcessExec.RunCommand, and it
-		// wraps the original exit error if there's one
-		exitError, ok := err.(shared.RunError).Unwrap().(*exec.ExitError)
-		if !ok {
-			// not an exit error, abort and bubble up the error
-			logger.Warnf("Purge failed with non-exit error: %v", err)
-			break
-		}
-		if syscall.Errno(exitError.ExitCode()) != syscall.EBUSY {
-			// not a busy error, abort and bubble up the error
-			logger.Warnf("Purge failed with unexpected exit error: %v", exitError)
-			break
-		}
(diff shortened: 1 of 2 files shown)


lmlg (approved):
LGTM

microceph/ceph/osd.go
```diff
@@ -1894,26 +1894,14 @@ func (m *OSDManager) purgeOSD(osd int64) error {
...
-		}
-		// purge failed with EBUSY - retry after a delay, and make delay exponential
-		logger.Infof("Purge failed %v, retrying in %v", err, backoff)
+		// Retry on any failure: the OSD process may still be shutting down
+		// when purge is attempted, causing transient errors. ceph CLI exits
+		// with 1 (not errno 16) in these cases.
+		logger.Infof("Purge attempt %d failed: %v, retrying in %v", i+1, err, backoff)
 		backoff = time.Duration(math.Pow(2, float64(i))) * time.Millisecond * 100
-		time.Sleep(backoff)
+		purgeRetrySleepFunc(backoff)
```
johnramsden, line 1904: Nit: Avoid last sleep (long with backoff)

ie:

https://github.com/canonical/microceph/blob/4aee44e035453a6e92bf6dce668a6bb414a23b6f/microceph/ceph/osd.go#L928-L931

Otherwise lgtm

PR #750, opened 2026-05-29
ceph: treat already-removed partition entry as cleaned on WAL/DB cleanup

### LINKED ISSUE #749: Flaky: WAL/DB partition cleanup fails with 'partx -d ... error deleting partition' on shared-device OSD removal
## Summary

`microceph disk remove` can intermittently fail when removing an OSD whose generated WAL/DB partitions live on a **device shared with another OSD**. The failure surfaces as:

```
Error: failed to remove disk: failed to clean generated WAL/DB partitions for osd.1:
  failed to remove kernel partition entry 1 on /dev/.../...-db1:
  Failed to run: partx -d --nr 1:1 /dev/.../...-db1: exit status 1
  (partx: /dev/.../...-db1: error deleting partition 1)
```

It is timing-dependent (flaky): the same test passes on some runs and fails on others.

Observed in CI, e.g. the `DSL WAL-DB cleanup tests` job, case
`rm2: test_dsl_remove_one_of_two_osds_only_cleans_its_partitions`

Attempt 1,2 fail, Attempt 3 pass:
* https://github.com/canonical/microceph/actions/runs/26655407817/job/78564273941).

## Root cause

`deletePartition` in `microceph/ceph/osd_waldb.go` (~L565) cleans a partition in two steps:

1. `sfdisk --delete <dev> <n>` to remove the partition from the GPT.
2. `partx -d --nr <n>:<n> <dev>` to drop the kernel partition entry.

Unlike `createPartition` (which uses `sfdisk --append --force --no-reread` and then explicitly refreshes with `partx -u`), the delete path runs **plain `sfdisk --delete` without `--no-reread`**. So `sfdisk` *also* asks the kernel to re-read the partition table after the edit. That makes two things race to remove the same kernel entry:

- **sfdisk's implicit re-read wins** -> kernel entry `n` is already gone when `partx -d` runs -> `partx -d` exits 1 (`error deleting partition`) -> whole remove fails.
- **`partx -d` wins** -> it removes the entry and exits 0 -> remove succeeds.

Two factors flip the outcome between runs:

1. **Disk-busy state.** `sfdisk`'s re-read is a whole-disk `BLKRRPART`, which fails if any partition on the disk is held open. In the shared-device case the sibling partition still belongs to a running OSD, so whether the re-read succeeds depends on whether that partition is held at that instant.
2. **udev/settle timing** between the re-read and the immediately-following `partx -d`.

This is why only the shared-device case (`rm2`) flakes while single-OSD cases (`rm1`, `rmr`) pass -- only the shared case has a sibling partition whose busy/idle state makes sfdisk's re-read flip between succeeding and failing.

The `sfdisk --delete` step already has idempotency handling (if the partition disappears despite an error, it's treated as cleaned), but the `partx -d` step does not -- its error is returned verbatim.

## Proposed fix

Make the `partx -d` step idempotent the same way the `sfdisk` step already is: if `partx -d` errors but the partition no longer resolves, the kernel entry is already gone (the desired end state), so treat it as cleaned rather than failing the remove.

```go
_, err = m.runner.RunCommand("partx", "-d", "--nr", partitionRange, parentPath)
if err != nil {
    _, resolveErr := m.resolvePartitionStablePath(parentPath, partition)
    if resolveErr != nil {
        logger.Infof("Kernel partition entry %d on %s already removed, treating as cleaned", partition, parentPath)
        return nil
    }
    return fmt.Errorf("failed to remove kernel partition entry %d on %s: %w", partition, parentPath, err)
}
```

A unit test covering "`partx -d` errors but the partition is already absent" should be added to `microceph/ceph/osd_waldb_test.go`.

## Notes

- Introduced with the WAL/DB DSL feature (#699); not specific to any particular caller.
- The Robot Framework CI migration started exercising this path in a way that makes the flake visible.

johnramsden:
# Description

deletePartition removes a generated WAL/DB partition in two steps: "sfdisk --delete" then "partx -d". Because the sfdisk call does not pass --no-reread, the kernel can re-read the partition table and drop the entry before "partx -d" runs, so partx exits non-zero with "error deleting partition". That redundant failure propagated up and intermittently failed "microceph disk remove" when generated partitions shared a WAL/DB carrier (DSL rm2 case).

Treat a "partx -d" failure as success when the partition no longer resolves, mirroring the existing idempotency handling on the preceding "sfdisk --delete" step. Add TestDeletePartitionTreatsAlreadyRemovedKernelEntryAsCleaned, which reproduces the race deterministically.

Fixes: #749

## Type of change

## How has this been tested?

Added unit test exercising regression

## Contributor checklist

Please check that you have:

### DIFF
--- a/microceph/ceph/osd_waldb.go
+++ b/microceph/ceph/osd_waldb.go
@@ -582,6 +582,11 @@ func (m *OSDManager) deletePartition(parentPath string, partition uint64) error 
 	partitionRange := fmt.Sprintf("%d:%d", partition, partition)
 	_, err = m.runner.RunCommand("partx", "-d", "--nr", partitionRange, parentPath)
 	if err != nil {
+		_, resolveErr := m.resolvePartitionStablePath(parentPath, partition)
+		if resolveErr != nil {
+			logger.Infof("Kernel partition entry %d on %s already removed, treating as cleaned", partition, parentPath)
+			return nil
+		}
 		return fmt.Errorf("failed to remove kernel partition entry %d on %s: %w", partition, parentPath, err)
 	}
 
--- a/microceph/ceph/osd_waldb_test.go
+++ b/microceph/ceph/osd_waldb_test.go
@@ -2,6 +2,7 @@ package ceph
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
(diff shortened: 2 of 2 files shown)


UtkarshBhatthere (approved):
LGTM, thanks @johnramsden

PR #751, opened 2026-05-29
fix: support local tentacle dashboard build

jamesbeedy:
## Summary
- add the dashboard crypto patch file used by local tentacle builds
- switch the snap build to `go mod download`
- relax the local-build-only snapcraft path so patched local builds can complete

## Why
We need tentacle for dashboard OAuth2 support, and the local snap build path needs to reliably produce a patched artifact for cluster bring-up.


UtkarshBhatthere:
@jamesbeedy 2 concerns about the change.

1.) we can always commit the patch in the git tree (even for local builds) so I'd prefer to keeping the check.

2.) In commit messages, we've agreed to steer to using "Assisted-by:" rather than "Co-Authored-by:" as both portray 2 different kinds of sign-offs.

Overall the patch lgtm, the 2 missing deb packages were on my list as well.

PR #753, opened 2026-06-02
feat: Implement a security event logging module

lmlg:
This PR implements a module for security event logging, closely modeled after the one in the ceph-charms repository.

### DIFF
--- a/microceph/logger/selog.go
+++ b/microceph/logger/selog.go
@@ -1,0 +1,141 @@
+// Copyright 2026 Canonical Ltd
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package logger
+
+import (
+	"encoding/json"
+	"fmt"
+	"strings"
+	"time"
+)
+
+// Parameters for generating a security log.
+type LogParams struct {
+	Level  string
+	Msg    string
+	AppID  string
+	Event  string
+	Detail string
+}
+
+type LogCallback func(msg string)
+
+var (
+    // Overridable callback that outputs a log string.
+	logCallbackFunc LogCallback = defaultLogOutput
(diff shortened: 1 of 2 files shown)


microceph/logger/selog.go
```diff
@@ -0,0 +1,141 @@
...
+
+var (
+    // Overridable callback that outputs a log string.
+	logCallbackFunc LogCallback = defaultLogOutput
+	defaults = make(map[string]string)
+)
+
+func defaultLogOutput(msg string) {
+	Warn(msg)
+}
```
sabaini, line 48: So on the default log cb implementation we're squashing all severities into `Warn` -- it would be good to have some reasoning why the squash and why `Warn` (afaict the ceph-charms version inherits a default of INFO from charmhelpers), either as source or reference docs

microceph/logger/selog.go
```diff
@@ -0,0 +1,141 @@
...
+	prev := logCallbackFunc
+	logCallbackFunc = fn
+	return prev
+}
+
+func RegisterDefaults(dfls map[string]string) map[string]string {
+	prev := defaults
+	defaults = dfls
+	return prev
+}
```
sabaini, line 64: Hm, this overwrites a global dict, I believe this could cause races for concurrent logging and defaults registration. Suggest to protect this with a mutex similar to how we do it in logger.go

sabaini (changes requested):
Hey @lmlg two minor nits inline 
It would also be good to have some general docs for this module, maybe even could just refer to the spec so that future travellers have some context

microceph/logger/selog.go
```diff
@@ -0,0 +1,150 @@
...
+	levelUpper := strings.ToUpper(level)
+	if levelUpper != "INFO" && levelUpper != "WARN" && levelUpper != "ERROR" {
+		return "", fmt.Errorf("level must be one of INFO, WARN, ERROR")
+	}
+
+	nowUTC := time.Now().UTC().Format("2006-01-02T15:04:05.000000-07:00")
+	obj := securityLog{
+		Level:       levelUpper,
+		Msg:         msg,
+		Type:        "security",
```
UtkarshBhatthere, line 94: Is "security" an expected type value ? Can there be any other values ?

lmlg: It's always set to "security". See: https://github.com/canonical/livepatch-client/blob/develop/docs/explanations/Security-Event-Logging.md

sabaini (approved):
changes lgtm, think it would be good to mention the spec and/or jira this relates to for context either in the source or the PR but not a blocker

sabaini:
FTR., spec CE115

PR #758, opened 2026-06-08
fix: check remote paths (CVE-2026-10720)

sabaini:
Fix CVE-2026-10720:

MicroCeph versions from the squid and tentacle track are vulnerable to a path traversal issue in the remote-import API. Holders of a trusted cluster mTLS certificate (such as enrolled cluster members) or join token can manipulate files in an imported remote cluster within the /var/snap/microceph confinement. This would allow daemon disruption and pollution of the cluster state.

CVSS 3.1

CVSS:3.1/AV:N/AC:L/PR:H/UI:N/S:U/C:N/I:L/A:L (Medium, 4.4)

### DIFF
--- a/microceph/api/cluster.go
+++ b/microceph/api/cluster.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
-	"regexp"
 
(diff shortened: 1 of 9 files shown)


lmlg (approved):
lgtm, good stuff.

PR #772, opened 2026-06-24
fix: refresh radosgw.conf mon host from UpdateConfig

### LINKED ISSUE #766: radosgw.conf mon host list is never refreshed (can be permanently stale)
# Issue report

Reproducible on current `main`. Reported downstream in Sunbeam:
https://bugs.launchpad.net/snap-openstack/+bug/2095567

1. Deploy a multi-node cluster and enable RGW (e.g. 3 monitors).
2. Note that nodes record only the monitors that existed at join time
   (node 2 learns mon 1; node 3 learns mon 1+2).
3. Inspect `radosgw.conf` on a later-joined node.

`radosgw.conf` is written once at RGW enable time and never refreshed, so its
`mon host` list can be permanently stale (e.g. only 2 of 3 monitors):

```
mon host = 192.168.123.11,192.168.123.12
```

Unlike `ceph.conf` -- which the `Start()` background goroutine re-renders every
minute via `UpdateConfig()` (`ceph/start.go`, `ceph/config.go`) -- the RGW config
is written exactly once by `EnableRGW()` (`ceph/rgw.go`) and has no refresh path.

`radosgw.conf`'s `mon host` list should be kept in sync with the current
monitors, the same way `ceph.conf` is.

Not applicable.

## Additional comments.

`mon host` is only a bootstrap hint; once radosgw reaches any one listed monitor
it pulls the full monmap. So this does not break HA in normal operation -- it
only bites when *every* listed monitor is down at radosgw startup while an
unlisted one is up.

Suggested fix: keep `radosgw.conf`'s `mon host` line in sync from the same
refresh path that updates `ceph.conf`. Because the RGW port/SSL frontend settings
are not persisted (and so cannot be re-rendered from scratch), the line can be
rewritten in place to preserve those settings.

sabaini:
# Description

radosgw.conf's "mon host" line was written once at RGW enable time by EnableRGW() and never refreshed, while ceph.conf is re-rendered every minute by the Start() goroutine via UpdateConfig(). As monitors join the cluster, radosgw.conf's mon host list goes permanently stale (e.g. only 2 of 3 monitors), so radosgw can fail to bootstrap when every listed monitor is down at startup while an unlisted one is up. (Downstream: LP snap-openstack bug 2095567.)

Keep radosgw.conf's mon host in sync from the same refresh path that updates ceph.conf: add updateRadosGWMonHost(), which reuses the existing atomic in-place line rewriter fixConfigLine (sibling to fixRadosGWRunDir) to rewrite only the "mon host =" line. This preserves the unpersisted RGW frontend port/SSL settings, which cannot be re-rendered from scratch. UpdateConfig() invokes it right after formatting the monitor addresses, so both files share one source of truth, and warns-and-continues on error so a radosgw.conf write failure cannot block the ceph.conf refresh. A missing file (RGW disabled) is a no-op; an already-current line is left untouched (idempotent).

Also switch UpdateConfig() to the injectable fetchConfigDb seam (already used by backwardCompatPubnet) so the refresh path is testable without a real database.

Fixes #766

Assisted-by: pi:z-ai/glm-5.2

## Type of change

## How has this been tested?

included tests

## Contributor checklist

Please check that you have:


microceph/ceph/config.go
```diff
@@ -403,6 +403,16 @@ func UpdateConfig(ctx context.Context, s interfaces.StateInterface) error {
 	// Ensure that IPv6 addresses have square brackets around them (if IPv6 is used).
 	monitorAddresses = formatIPv6(monitorAddresses)
 
+	// Keep radosgw.conf's `mon host` line in sync with the current monitors.
+	// radosgw.conf is written once at RGW enable time (EnableRGW) and is never
+	// re-rendered from scratch, because the RGW frontend port/SSL settings are
+	// not persisted. We therefore rewrite only the mon host line in place,
+	// preserving the other settings. A missing file (RGW disabled) is a no-op;
+	// a failure here must not block the ceph.conf refresh below.
+	if err := updateRadosGWMonHost(confPath, monitorAddresses); err != nil {
```
johnramsden: nit: Avoid one line assign/test per CONTRIBUTING.md

microceph/ceph/start.go
```diff
@@ -194,6 +195,41 @@ func fixRadosGWRunDir(confFile, correctRunDir string) error {
...
+
+	// Normalize the monitor order before comparing: getMonitorsFromConfig
+	// iterates a map (randomized order in Go), and the no-change check below
+	// is an exact byte comparison. Sorting the slice avoids a spurious rewrite
+	// (and the tmp+rename dance) on every UpdateConfig tick when the monitor
+	// set is stable but the iteration order differs.
+	sorted := append([]string(nil), monitors...)
+	sort.Strings(sorted)
+
+	return fixConfigLine(filepath.Join(confDir, "radosgw.conf"), func(line string) (string, bool) {
```
johnramsden: Previously this was only used for correcting run directory lines, so the function logs: 

> migration: fixed stale run dir in %s 

Suggest updating this log message to be more generic

microceph/ceph/config.go
johnramsden: If we run this function here since it is in a go routine happening simultaneously with `migrateStaleRunDir`, I think we could end up with a race condition where they are both editing simultaneously. The rename is atomic but they could end up ovewriting each other's temporary file. 

https://github.com/canonical/microceph/blob/2ff778adea9b2ca4b26f009420b3a4611f1c8f5d/microceph/ceph/start.go#L419 

Potential Solutions: Mutex, or synchronize the writes.

johnramsden (changes requested):
Overall looks good. I think there is one small race condition that should be investigated, and a nit.

microceph/ceph/config.go
sabaini: added a mutex

microceph/ceph/start.go
```diff
@@ -282,7 +294,7 @@ func fixConfigLine(confFile string, fn func(string) (string, bool)) error {
 		os.Remove(tmpFile)
 		return fmt.Errorf("failed to replace %s: %w", confFile, err)
 	}
-	logger.Infof("migration: fixed stale run dir in %s", confFile)
```
johnramsden, line 249: This line was being grepped for in NFS tests, so this needs to be resolved in the test as well. 

https://github.com/canonical/microceph/blob/c87edf7941d34d0f4245de65c9b75e58036898c7/tests/robot/nfs-test/nfs_tests.robot#L84

Suggest either changing the grep, or instead of logging in the function returning a boolean if the line is changed and logging a more relevant message.

johnramsden (changes requested):

johnramsden (approved):
Looks good!

PR #799, opened 2026-07-10
test(robot): trace every command to the console, set -x style

### LINKED ISSUE #798: Robot does not output sufficient logs during runs
This is in part because logs are sent to info which records only in the final html file rather than to the console

johnramsden:
# Description

Fixes #798
Fixes CEPH-1899

Robot's `--console verbose` prints one line per test case and nothing per keyword, and the harness sent command output to `logger.info` (log.html only). Watching a live job you saw phase banners and `| PASS |` but no cluster state; diagnosing a failure meant downloading the artifact. The pre-Robot bash CI didn't have this — it ran under `set -x`.

Restore that at the one layer every command funnels through, rather than converting call sites by hand. `run_in_vm` / `run_in_container` / `exec_in_container` / `run_in_container_unchecked` now echo `+ <command>` and its stdout to the console (nested `lxc exec` shown as its logical command, not the raw argv). The `*_and_check` / `*_must_fail` wrappers inherit it, so all ~135 harness and ~394 suite call sites are covered — no per-suite edits, nothing that can be silent.

Before — console:

```
Test Add WAL DB OSD :: ... | PASS |
```

After — console:

```
+ sudo microceph disk add /dev/sdia --wal-device /dev/sdib --db-device /dev/sdic
+-----------+---------+
|   PATH    | STATUS  |
+-----------+---------+
| /dev/sdia | Success |
+-----------+---------+
+ sudo microceph.ceph -s
  cluster:
    health: HEALTH_OK
  ...
```

- **Poll loops / bulk getters** pass `quiet=True` so the console stays readable (output still in log.html); the ~13 pollers already print their own found-value summary.
- The 80-char command truncation is dropped (full command shown).
- **Failure diagnostics**: `_poll_until` takes a callable `fail_msg` so pollers fold in the last observed value (`last saw 1 in`) and wrap `on_fail` in try/except so a raising diagnostic can't hide the real message; `on_fail` dumps are non-raising and now visible; teardown diagnostics stream to the console on failure only.

No suite files change.

## Type of change

## How has this been tested?

- `pytest tests/robot/resources/test_harness_helpers.py` — 114 pass (7 new: `_log_exec` console/quiet routing, `_poll_until` failure semantics).
- Runtime testing

## Contributor checklist

Please check that you have:

### DIFF
--- a/tests/robot/resources/microceph_harness.py
+++ b/tests/robot/resources/microceph_harness.py
@@ -174,6 +174,16 @@ class microceph_harness:
         """
         return self._coerce_xtrace(BuiltIn().get_variable_value("${XTRACE}", False))
 
+    def _suite_failed(self):
+        """Returns True when the current suite has already failed (read during teardown).
+
+        Robot exposes ${SUITE STATUS} as PASS/FAIL during suite teardown. Teardown
+        diagnostics stream to the console only on failure, so a green run does not flood
+        the console with a 200-line snap-logs dump.
+        """
+        status = BuiltIn().get_variable_value("${SUITE STATUS}", "")
+        return str(status).upper() == "FAIL"
+
     # -----------------------------------------------------------------------
     # Host dependency checking
     # -----------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


tests/robot/resources/microceph_harness.py
```diff
@@ -238,7 +273,6 @@ def run_in_container(self, container, cmd, timeout=300):
         bash -eo pipefail: mirrors set -e semantics so any failing command or pipe stage
         inside the container fails the keyword immediately.
         """
-        logger.console(f"[{container}] {cmd[:80]}")
```
sabaini, line 241: hm, think this could come in handy in cases where the cmd hangs so you see where it hangs
poss. could do similar loggin in exec... and run_in_container_unchecked() below

sabaini (approved):
Hey, one small suggestion but nothing to block over

PR #807, opened 2026-07-22
snap: rebase onto core26 (resolute)

UtkarshBhatthere:
- Rebase MicroCeph onto core26 for Resolute Ceph 20 compatibility.
- Move snap runtime/tooling to Python 3.14 and updated packaging flow (dqlite 1.18.7, pip-based microceph-orch install).
- Update monitor handling for Ceph 20 and remove obsolete Ceph manager compatibility assumptions.
- Align integration tests with Ubuntu 26.04 and `ceph-common` from `ppa:lmlogiudice/ceph-stonking-tentacle` so NFS/CephFS clients match snap Ceph versions.
- Update validation with Go unit tests, robot unit-tests, and mgr remote-regression coverage.


UtkarshBhatthere:
## CI triage: all four matrix failures reviewed

| Suite | Verdict |
|---|---|
| NFS multinode (both runs) | Real bugs, see below |
| Messenger v2 (one run) | Runner infra flake: LXD `forkfile.sock` dial failure pushing the test script into the VM; same-SHA twin passed |
| Deferred-Ceph multi (one run) | Flake-shaped: microcluster forward `context canceled` during ceph-only bootstrap; twin and all other multi-node suites passed |

### NFS multinode

Two independent bugs:

1. **ganesha 6.x rejects the quoted RecoveryBackend token.** 4.x accepted
   `RecoveryBackend = 'rados_cluster'`; 6.5 logs `Unknown token
   ('rados_cluster')` and silently falls back to filesystem v4 recovery,
   breaking clustered grace. Fixed in `fd84435` (unquoted; verified on a
   live 3-node core26 cluster: `rados_cluster` backend loads and reads the
   RADOS recovery db).

2. **resolute's ceph 20.2.0 libcephfs segfaults in `MDSMap::decode`.**
   Reproduced with a ~20-line C libcephfs client (`ceph_create` +
   `ceph_conf_read_file` + `ceph_mount`) run against the snap's own
   cluster, no ganesha involved:

   ```
   #0 MDSMap::decode(ceph::buffer::v15_2_0::list::iterator_impl<true>&)  libceph-common.so.2
   #1 libcephfs.so.2
   #3 Messenger::ms_deliver_dispatch(...)
   ```

   The crash needs a filesystem in the FSMap and appears tied to standby
   MDS entries, which is why single-node NFS (1 MDS, no standby) passes
   while multinode (3 MDS) crash-loops ganesha the moment the CephFS
   export loads. **Multinode NFS will stay red on this PR until the ceph
   side is fixed.**

   The tentacle-rc PPA's 20.2.1 (resolute) fixes the decode crash — the
   same reproducer mounts cleanly and the full multinode NFS flow
   (enable, export, cross-node mount, write) passes — but its ceph-mgr
   breaks every cross-module `remote()` call (`nfs export *`,
   `orch status`, pg_autoscaler → `HEALTH_ERR`, all failing with an empty
   "Remote method threw exception"), which would break more suites than
   it fixes. So this PR stays on the archive 20.2.0 and the path forward
   is getting a fixed 20.2.1 into the resolute archive (or fixing the
   PPA mgr build).

Also in `fd84435`: the robot teardown diagnostics now tail
`ganesha.log` — it logs to a file, so the journal-only dump made these
failures undiagnosable from CI output.

UtkarshBhatthere:
## Round 3 results + the multinode NFS crash explained

The ganesha.log teardown diagnostics paid off. CI's actual fatal on the
multinode inner nodes:

```
MAIN :FATAL :Failed to PR_SET_IO_FLUSHER with EPERM.
Take a look at config option allow_set_io_flusher_fail
```

ganesha 6.x makes a `prctl(PR_SET_IO_FLUSHER)` call at startup (4.x did
not) that needs CAP_SYS_RESOURCE, i.e. the `process-control` interface.
The single-node suite connects the full interface set; the multinode
inner nodes only connect the minimal one - hence single-node green,
multinode crash-looping. `2827de2` fixes both sides:

- `Allow_Set_Io_Flusher_Fail = true` in the generated ganesha.conf, so
  ganesha degrades gracefully wherever the capability is absent
  (verified: with process-control deliberately disconnected the old
  build dies FATAL, the new build logs `WARN ... ignoring` and the
  service stays active);
- `process-control` added to the tests' minimal interface set.

Round 3 scorecard (fd84435): everything green except NFS, including
messenger-v2 and deferred-ceph-multi (confirming those were flakes) -
one DSL WAL-DB consistency failure was an osd-purge-vs-osd-down timing
race whose same-SHA twin passed.

The NFS suites this round also confirmed the ceph 20.2.0 client-stack
instability is broader than MDSMap::decode: single-node NFS failed with
a `ganesha-rados-grace` **segfault** during the grace-db add (a plain
librados tool, no ganesha involved; prior rounds passed with identical
binaries). And a virgin single-node cluster on the PPA's 20.2.1
confirmed its mgr cross-module remote breakage is inherent, not
cluster-state: `fs volume create`, `nfs export create` and
pg_autoscaler fail from first boot.

So the ceph-side matrix stands: archive 20.2.0 = crashy client libs /
healthy mgr; PPA 20.2.1 = healthy client libs / broken mgr remote.
NFS needs both halves, so the suites stay at the mercy of 20.2.0's
nondeterministic client crashes until a fixed tentacle build lands in
resolute. Expectation for this round: multinode NFS gets past enable
now, and either passes or dies at export-load/mount on the libcephfs
decode bug.

UtkarshBhatthere:
## Root cause of the 20.2.1 (tentacle-rc PPA) mgr `remote()` breakage

Pinpointed to a mangled hunk in the PPA packaging, not upstream code.

Upstream `src/mgr` is essentially unchanged between v20.2.0 and
v20.2.1. The difference is that the PPA's 20.2.1 packaging carries
`debian/patches/cephadm-pyo3-fix.patch` (a backport of Samuel Just's
main-interpreter rework, needed because PyO3 modules broke cephadm on
resolute - LP: #2150762): mgr modules now load into the main
interpreter by default and `dispatch_remote` passes results between
modules as pickled bytes. The resolute archive's 20.2.0 does not carry
this patch, which is why its mgr works.

The backported hunk in `ActivePyModule::dispatch_remote` inverted the
success/error branches:

```c
  if (ret == nullptr) {
    std::string caller = "ActivePyModule::dispatch_remote "s + method;
    *err = handle_pyerror(true, get_name(), caller);
  } else {
    return std::nullopt;   // <- success path returns the error
  }                        //    sentinel with *err still empty
  dout(20) << "Success calling '" << method << "'" << dendl;
  // ... pickle.dumps(ret) - only reachable when ret == nullptr
```

Upstream main has the correct flow (`if (ret == nullptr) { *err = ...;
return std::nullopt; }`, then the success path pickles the result). So
on the PPA build every successful cross-module `remote()` call returns
the error sentinel with an empty message - exactly the observed
`RuntimeError: Remote method threw exception: ` from `nfs export *`,
`orch status`, `fs volume create`, and pg_autoscaler, and exactly the
debug-20 log signature (the `Calling <module>.<method>...` line
appears, the `Success calling` line never does).

Fix: correct that hunk to match upstream and respin the PPA. With that,
20.2.1 should be healthy on both axes (its client libs already fix the
`MDSMap::decode` segfaults), and this PR can switch the core26 snap to
it, unblocking NFS entirely.

tests/robot/resources/microceph_harness.py
johnramsden: Swap to core26 (caching)

tests/robot/cephfs-replication-test/cephfs_replication_tests.robot
```diff
@@ -10,12 +10,30 @@ Suite Teardown  Teardown MicroCeph Environment
...
 *** Keywords ***
 CephFS Replication Suite Setup
     Provision Multinode VM    microceph-cfsrep-vm    ${OUTER_VM_DISK}    public
+    Verify Resolute Outer VM
+
+Verify Resolute Outer VM
+    [Documentation]    Ensures the external CephFS mount client runs on the PPA's target release.
+    Run In VM And Check    grep -Fx 'VERSION_ID="26.04"' /etc/os-release    30
+
+Install Ceph Client From PPA
```
johnramsden: Duplicated in nfs tests, suggest moving to harness

tests/robot/cephfs-replication-test/cephfs_replication_tests.robot
```diff
@@ -10,12 +10,30 @@ Suite Teardown  Teardown MicroCeph Environment
...
+    [Documentation]    Installs the CephFS mount client from the same Resolute PPA as the snap.
+    Run In VM And Check    sudo env DEBIAN_FRONTEND=noninteractive apt-get install -y software-properties-common    300
+    Run In VM And Check    sudo add-apt-repository --yes ${CEPH_PPA}    120
+    Run In VM And Check    sudo apt-get update    120
+    Run In VM And Check    sudo env DEBIAN_FRONTEND=noninteractive apt-get install -y ceph-common    300
+    ${package}=    Run In VM And Check    dpkg-query -W ceph-common    30
+    Should Contain    ${package.stdout}    ~ppa
+    ${policy}=    Run In VM And Check    apt-cache policy ceph-common    30
+    Should Contain    ${policy.stdout}    lmlogiudice/ceph-stonking-tentacle
```
johnramsden: Nit: Could be var

tests/scripts/test_ceph_mgr_notify_patch.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+# Verify the staged Ceph manager workaround applies to the packaged module.
```
johnramsden: Observation: I think this is completely synthetic since you are testing a file you always create exactly the same in the test. Should this be testing patching an existing file as a regression test?

johnramsden (approved):
This looks good to me. Mostly a few non blocking nits. Nice work.

lmlg (approved):
lgtm, great work

PR #809, opened 2026-07-23
feat: add radosgw-admin multisite read wrappers

johnramsden:
# Description

Add plumbing for RGW multisite replication
* radosgw-admin exec wrapper with a remote-cluster variant and typed
* JSON-backed wrappers for realm get, zonegroup get, zone get, metadata sync status, per-source data sync status, and the mdlog/datalog head markers.

Failing commands yield zero-value results with a nil error, mirroring the RBD wrapper contract, so the upcoming replication handler can treat an unconfigured gateway as ordinary disabled state.

Fixtures are captured from a live two-site squid deployment, covering master, secondary and unconfigured views.

Nothing calls this code yet; the replication handler arrives in a follow-up.

## Type of change

## How has this been tested?

Mocked unit tests exercise new functions

## Contributor checklist

Please check that you have:


microceph/ceph/test_assets/rgw_zone_get.json
```diff
@@ -0,0 +1,43 @@
...
+    "user_swift_pool": "sitea.rgw.meta:users.swift",
+    "user_uid_pool": "sitea.rgw.meta:users.uid",
+    "otp_pool": "sitea.rgw.otp",
+    "notif_pool": "sitea.rgw.log:notif",
+    "topics_pool": "sitea.rgw.meta:topics",
+    "account_pool": "sitea.rgw.meta:accounts",
+    "group_pool": "sitea.rgw.meta:groups",
+    "system_key": {
+        "access_key": "VDS6K1QMOGKYIJA5WTQD",
+        "secret_key": "cE3yAkkOisWAwBj7fYBErJ5RedXeeAJX7JXZOERe"
```
sabaini, line 24: Just for avoidance of doubt -- these are from a test system presumably? Otherwise wouldn't want to commit those :-D

microceph/ceph/rgw_multisite.go
```diff
@@ -0,0 +1,342 @@
...
+			verdict.FullSyncShards++
+			continue
+		}
+		if shard.Key < len(masterLog) && masterLog[shard.Key].Marker > shard.Val.Marker {
+			verdict.BehindShards = append(verdict.BehindShards, shard.Key)
+		}
+	}
+
+	verdict.CaughtUp = len(verdict.BehindShards) == 0 && verdict.FullSyncShards == 0
+	return verdict
```
sabaini, line 420: one worry around the CaughtUp calc 
from radosgw-admin they're calculating this including the NumShards

```
total_behind = shards_behind.size() + (sync_status.sync_info.num_shards - num_inc);
```

https://github.com/ceph/ceph/blob/df272aef3800f7b748004d57635007fc694c4950/src/rgw/rgw_admin.cc#L2282

i wonder if we should include NumShards too? the worry is that if partial/initializing shards could be reported as caught up

microceph/ceph/rgw_multisite.go
```diff
@@ -0,0 +1,342 @@
...
+			verdict.FullSyncShards++
+			continue
+		}
+		if shard.Key < len(sourceLog) && sourceLog[shard.Key].Marker > shard.Val.Marker {
+			verdict.BehindShards = append(verdict.BehindShards, shard.Key)
+		}
+	}
+
+	verdict.CaughtUp = len(verdict.BehindShards) == 0 && verdict.FullSyncShards == 0
+	return verdict
```
sabaini, line 455: similar to metadata verdict above

microceph/ceph/rgw_multisite.go
```diff
@@ -0,0 +1,342 @@
...
+		logger.Warnf("REPRGW: failed data sync status operation for source(%s): %v", sourceZone, err)
+		return RgwDataSyncStatus{}, nil
+	}
+
+	err = json.Unmarshal([]byte(output), &envelope)
+	if err != nil {
+		return RgwDataSyncStatus{}, fmt.Errorf("cannot unmarshal data sync status output: %w", err)
+	}
+
+	return RgwDataSyncStatus{Info: envelope.SyncStatus.Info, Markers: envelope.SyncStatus.Markers}, nil
```
sabaini, line 283: suggest to check invariants before returning as a safety, later on we do indexing on keys e.g.
eg. NumShards >= 0, keys between 0 and NumShards
if NumShards==0 i think there must not be markers?

microceph/ceph/rgw_multisite.go
```diff
@@ -0,0 +1,342 @@
...
+		logger.Warnf("REPRGW: failed metadata sync status operation: %v", err)
+		return RgwMetadataSyncStatus{}, nil
+	}
+
+	err = json.Unmarshal([]byte(output), &envelope)
+	if err != nil {
+		return RgwMetadataSyncStatus{}, fmt.Errorf("cannot unmarshal metadata sync status output: %w", err)
+	}
+
+	return RgwMetadataSyncStatus{Info: envelope.SyncStatus.Info, Markers: envelope.SyncStatus.Markers}, nil
```
sabaini, line 257: suggest invariant checking as per comment for GetRgwDataSyncStatus() below

sabaini (changes requested):
Hey, thanks for this @johnramsden left a few nits / questions

microceph/ceph/test_assets/rgw_zone_get.json
johnramsden: Yup!

microceph/ceph/rgw_multisite.go
```diff
@@ -0,0 +1,436 @@
...
+		return verdict
+	}
+
+	numIncremental := 0
+	for _, shard := range local.Markers {
+		if shard.Val.State != RgwMetadataSyncStateIncremental {
+			continue
+		}
+		numIncremental++
+		if shard.Key >= len(masterLog) {
```
sabaini: nit: maybe add an explicit assert for a neg. shard.Key, for future travellers

sabaini: similar for data below

microceph/ceph/rgw_multisite.go
```diff
@@ -0,0 +1,436 @@
...
+			logger.Warnf("REPRGW: metadata shard %d is missing from the peer mdlog (peer reported %d shard(s)); its sync state cannot be confirmed", shard.Key, len(masterLog))
+		} else if masterLog[shard.Key].Marker > shard.Val.Marker {
```
sabaini, line 409: nit: maybe add a note that we indeed want a lexicographical comparison to match upstream (a future traveller might be tempted to "fix" the comparison thereby breaking it) ie. that markers are zero padded fixed width so lexicographical is ok; similar for data below

sabaini (approved):
Heya @johnramsden lgtm added two tiny nits but no blockers

PR #814, opened 2026-08-10
feat: role managed RGW placement

sabaini:
# Description

## Summary

 Adds declarative, role-managed RGW placement to MicroCeph: a charm places RGW on a member by PUT /1.0/placement with an rgw object carrying enabled/port/ssl_port/ssl_certificate/ssl_private_key, and the daemon atomically enables the gateway with that frontend config. No separate enable rgw step, no out-of-band TLS handling. Observed frontend state (ports + a TLS flag) is reported back via GET /1.0/placement, sourced from a new dqlite table so it's available even when the member is down.

The placement rgw field moves from a bare boolean to an object. The charm was never shipped with the bool form, so the wire-type break needs no legacy shim; clients gate on the advertised placement-rgw capability marker.

## What changes

Wire contract — MemberPlacement.Rgw changes from *bool to *RgwPlacement{Enabled, Port, SSLPort, SSLCertificate, SSLPrivateKey}. A bare-bool body is rejected as a client error. The placement-rgw capability marker is advertised in CapabilitiesSupported.

Placement engine — a new applyRgwPlacement reconcile pass runs after the control pass: add-before-remove (a migration keeps a gateway serving), sorted for determinism, no keep-one invariant (scale-to-zero is allowed), and only removes RGW where it is observed. Short-circuits when no member has a non-nil rgw so control-only applies never touch RGW. Production dispatch reuses the existing services/rgw PUT/DELETE transport (proxied to the target member).

Member side — applyRGWFrontend renders radosgw.conf from the database and compares only the rgw frontends line (+ SSL files) against the on-disk state, so restarts happen only on a real frontend/TLS change — not on mon host/run dir churn from the periodic in-place writers (the idempotency bug that would have restarted RGW on every reconcile in multi-mon/IPv6 clusters). effectiveRGWPorts centralizes the default-port-80 rule and forces ssl_port=0 for plaintext so observed state never reports a stray SSL port.

DisableRGW is idempotent.

Secrets posture — SSL key material travels over the authenticated API to the member that needs it, then is dropped before the policy is stored (PolicyForStorage strips at PUT time; redactStoredPolicy is defense-in-depth at GET time). Key material is never persisted in dqlite. Malformed base64 TLS maps to HTTP 400.

Observed state — a dedicated rgw_frontends table (one row per member, FK CASCADE to core_cluster_members) records the observed frontend; GET /placement reads it in the same transaction, so a down member doesn't block observed-frontend reporting.

Tests — unit tests for the types, DB accessors, member-side idempotency (incl. a multi-monitor reorder regression and plaintext ssl_port), engine ordering, and the secret-strip/redact helpers; a new rgw-placement-tests Robot suite drives the full feature end-to-end (capability → enable via object → observed frontend + no secret leak → scale-to-zero); pure Python parsers for the integration suite assert observed reporting and the no-leak posture.

## Commits, reviewable by area

 1. docs(agents) — conf-file/dqlite-source-of-truth invariant
 2. feat(api) — object-shaped rgw placement types + placement-rgw capability
 3. feat(db) — rgw_frontends table + accessors
 4. feat(ceph) — strip/redact RGW TLS + populate observed frontend (GET path)
 5. feat(rgw) — idempotent member-side enable/disable + observed frontend record
 6. feat(placement) — RGW reconcile pass + prod wiring (SET path)
 7. feat(api) — strip/redact in store/get + 400 on bad TLS
 8. docs — shipped design note
 9. test(robot) — functional suite + harness parsers

## Breaking change

The placement rgw field type changes (*bool → *RgwPlacement). No shipped charm uses the bool form; gate on placement-rgw before sending the object payload.


microceph/api/placement.go
```diff
@@ -176,8 +177,11 @@ func cmdPlacementPut(s mcTypes.State, r *http.Request) mcTypes.Response {
 		logger.Warnf("failed to clear placement refusal: %v", clearErr)
 	}
 
-	// Persist the policy only after successful application.
-	err = ceph.StorePlacementPolicyFunc(ctx, interfaces.CephState{State: s}, policy)
+	// Persist the policy only after successful application. Strip RGW SSL key
+	// material before storage (CE142 Option B secrets posture): the cert/key
```
johnramsden, line 181: Nit: I think we could just remove the mention of option B. Correct me if I'm wrong but I don't think we ever actually mentioned what option b is. It's possible to tell through deduction what that decision was, but it might be a little confusing if someone tries to figure out what this means.

This appears in various places across the PR

microceph/ceph/placement.go
```diff
@@ -341,6 +373,205 @@ func prodRemoveControlService(ctx context.Context, s interfaces.StateInterface,
...
+	}
+
+	var desiredEnable []string
+	desiredDisable := make(map[string]bool)
+	for memberName, mp := range policy.Members {
+		if mp.Rgw == nil {
+			continue // omitted: untouched
+		}
+		if mp.Rgw.Enabled {
+			desiredEnable = append(desiredEnable, memberName)
```
johnramsden, line 410: I believe we will get a failure here because we're not checking if it has already been enabled via the observedState. This isn't idempotent so we will get an error at

`genericHospitalityCheck("rgw")`

Which errors when it's already active.

I think the same problem occurs for changing the port. Since we go through this Hospitality check won't it always fail when you try to change the port?

I believe we need to allow it to not fail if it is already enabled but support modification of ports and other configuration that needs to be able to change, basically we need to allow it to get to the new code that you've written that will allow modification. It's my understanding that it might not be up to get there right now..

johnramsden (changes requested):
Right now the robot tests don't actually run, we need to add the new suite to `.github/workflows/tests.yml`

microceph/api/types/placement.go
```diff
@@ -7,15 +7,30 @@ type NFSPlacement struct {
...
 // (leave untouched). This is the generic, non-OS106 payload consumed by the
 // snap placement engine (CE142).
 type MemberPlacement struct {
 	// Control governs MON, MGR, and MDS placement. nil means untouched.
 	Control *bool `json:"control,omitempty" yaml:"control,omitempty"`
-	// Rgw governs RGW placement. nil means untouched.
-	Rgw *bool `json:"rgw,omitempty" yaml:"rgw,omitempty"`
+	// Rgw governs RGW placement and frontend config. nil means untouched; a
+	// non-nil value with Enabled false means remove RGW from the member.
+	Rgw *RgwPlacement `json:"rgw,omitempty" yaml:"rgw,omitempty"`
```
johnramsden, line 33: **Blocking: a policy stored by the already-shipped `declarative-placement` feature makes `GET /1.0/placement` return 500 after upgrade.**

`MemberPlacement.Rgw` was `*bool` at the merge base (`b8f8772`), and the PUT path already persisted whatever it accepted. After this change `ceph/placement.go:609` does a bare `json.Unmarshal` of the stored `PolicyJSON` into the new struct; `"rgw":true` fails to decode, the error propagates out of `GetPlacementStatus`, and `cmdPlacementGet` turns it into `InternalError`.

The whole status endpoint is then 500 until someone issues a corrective PUT - which is exactly the endpoint a charm polls to decide what to do next. "No shipped charm uses the bool form" covers the charm, but not a cluster whose policy was applied by hand or by a test.

A legacy-bool fallback in a custom `UnmarshalJSON`, or a one-time rewrite of stored `PolicyJSON` in `schemaUpdate10`, closes it.

microceph/ceph/rgw.go
```diff
@@ -1,16 +1,225 @@
...
+		curCert, e1 := os.ReadFile(certPath)
+		curKey, e2 := os.ReadFile(keyPath)
+		if os.IsNotExist(e1) || os.IsNotExist(e2) || !bytes.Equal(curCert, decodedCert) || !bytes.Equal(curKey, decodedKey) {
+			sslChanged = true
+		}
+	} else if fileExists(certPath) || fileExists(keyPath) {
+		sslChanged = true
+	}
+
+	if !frontendChanged && !sslChanged {
```
johnramsden, line 151: **Blocking: a stopped-but-correctly-configured RGW can never be recovered through placement.**

This is the mirror of the idempotency issue already raised, with the opposite trigger, so it needs its own fix. When the unit is *not* active - snapd gave up after repeated failures, an operator `snap stop`, or a prior enable that failed after writing the config - the sequence is:

1. `genericHospitalityCheck` passes (service is not active).
2. `ServiceInit` reaches `applyRGWFrontend`; the on-disk files already match the desired render, so this early return fires and returns `(false, nil)` without ever calling `startRGWFunc`/`restartRGWFunc`.
3. `genericPostPlacementCheck` fails on `snapCheckActive`, so the PUT returns `rgw service unable to sustain on host`.

Re-applying the identical policy reproduces the same no-op forever. Restart decisions are derived purely from file contents; liveness needs to be an input too - treat "service not active" as `changed` regardless of the file diff.

microceph/ceph/rgw.go
```diff
@@ -1,16 +1,225 @@
...
 )
 
+// effectiveRGWPorts applies the default-port-80 rule for a plaintext frontend
+// (no SSL material) and returns the port/sslPort that will actually be rendered
+// to radosgw.conf and recorded as observed state. When SSL is not configured
+// sslPort is forced to 0 so the observed frontend never reports a stray SSL
+// port (e.g. the enable-rgw CLI's --ssl-port default of 443) for a plaintext
+// gateway. Centralizing the rule keeps the on-disk render and the rgw_frontends
+// DB record consistent, so observed state matches what is running.
+func effectiveRGWPorts(port, sslPort int, sslCert, sslKey string) (int, int) {
```
johnramsden, line 34: **`effectiveRGWPorts` never defaults `ssl_port`, so a TLS placement without an explicit port binds to an ephemeral port.**

The default-port rule only runs in the plaintext branch; the TLS branch returns `port, sslPort` untouched. `ssl_port` is `omitempty`, so `{"enabled":true,"ssl_certificate":"...","ssl_private_key":"..."}` yields `sslPort == 0`. Rendering the actual template from `configwriter.go` with that data gives:

```
rgw frontends = beast ssl_port=0 ssl_certificate=/c ssl_private_key=/k
```

No plain port clause at all, and beast binds TLS to an OS-assigned port. It is then recorded in `rgw_frontends` as `ssl_port: 0`, so `GET /placement` cannot tell the operator where the gateway actually is.

The `enable rgw` CLI hides this behind a cobra default of 443; the placement path has no equivalent. Default to 443 when TLS is configured and `ssl_port` is 0, or reject it as `ErrRgwFrontendInvalid`.

microceph/ceph/rgw.go
```diff
@@ -1,16 +1,225 @@
...
+	pathConsts := constants.GetPathConst()
+
+	certPath := filepath.Join(pathConsts.SSLFilesPath, "server.crt")
+	keyPath := filepath.Join(pathConsts.SSLFilesPath, "server.key")
+
+	var decodedCert, decodedKey []byte
+	sslCertificatePath := ""
+	sslPrivateKeyPath := ""
+
+	sslConfigured := sslCert != "" && sslKey != ""
```
johnramsden, line 79: **Half a TLS pair silently downgrades the gateway to plaintext.**

`sslConfigured := sslCert != "" && sslKey != ""` treats "one set" identically to "neither set". A PUT carrying a certificate but no key - a certificates relation that is only half-ready is the obvious way to get there - returns 200, discards the certificate, and brings RGW up on plain port 80 via the default-port rule.

Related: because the fields are plain `omitempty` strings with no sentinel, "leave TLS as it is" and "go plaintext" are the same wire value. A client that reads `GET`, edits one field and PUTs it back therefore drops TLS with no signal. Rejecting a half-pair handles the first case and makes the second much harder to hit by accident.

microceph/ceph/rgw.go
```diff
@@ -1,16 +1,225 @@
...
+		}
+		decodedKey, err = base64.StdEncoding.DecodeString(sslKey)
+		if err != nil {
+			return false, fmt.Errorf("%w: failed to decode SSL private key: %v", ErrRgwFrontendInvalid, err)
+		}
+		sslCertificatePath = certPath
+		sslPrivateKeyPath = keyPath
+	}
+
+	port, sslPort = effectiveRGWPorts(port, sslPort, sslCert, sslKey)
```
johnramsden, line 94: **No range or collision validation on the ports before they are rendered.**

Nothing here or in the API layer checks that `port != sslPort`, or that either value is a valid TCP port. `{"port":8080,"ssl_port":8080,...}` renders

```
rgw frontends = beast port=8080 ssl_port=8080 ssl_certificate=... ssl_private_key=...
```

and beast fails its second `bind(2)`; a negative or >65535 port is written to `radosgw.conf` and into `rgw_frontends` unchallenged. The `ErrRgwFrontendInvalid` -> 400 pattern is already established in this PR for bad base64, so this would fit naturally alongside it (validating in `RgwServicePlacement.PopulateParams` would catch it before any DB write).

microceph/ceph/rgw.go
```diff
@@ -1,16 +1,225 @@
...
+			}
+		}
+	}
+
+	// Rewrite the config only when the frontend line changed (a cert rotation
+	// keeps the same paths, so the line is unchanged and only the SSL files are
+	// rewritten above). The full file is rendered so `mon host`/`run dir` stay
+	// present; UpdateConfig keeps them fresh thereafter.
+	if frontendChanged {
+		if err = rgwConf.WriteConfig(configs, 0644); err != nil {
```
johnramsden, line 177: **`radosgw.conf` is written here without the mutex that exists for that file.**

`radosgwConfMu` is defined at `ceph/start.go:197` precisely to serialise `updateRadosGWMonHost` (the per-minute tick) against `fixRadosGWRunDir`. `applyRGWFrontend` is a third writer of the same file, on an independent schedule, and doesn't take it.

The lost update is reachable: the ticker reads the old file, the placement PUT renders and renames the new one, then the ticker renames its patched copy of the *old* content back over it - silently reverting the port/TLS change the PUT just reported as successful. Taking `radosgwConfMu` around the read-compare-write section would close it.

microceph/ceph/placement.go
```diff
@@ -341,6 +373,205 @@ func prodRemoveControlService(ctx context.Context, s interfaces.StateInterface,
...
+		}
+	}
+	sort.Strings(desiredEnable)
+
+	// Add-before-remove: enable desired members first so a migration keeps a
+	// gateway serving while the old one is torn down.
+	for _, memberName := range desiredEnable {
+		err = enableRgwServiceFunc(ctx, s, memberName, *policy.Members[memberName].Rgw)
+		if err != nil {
+			return fmt.Errorf("failed to enable RGW on %s: %w", memberName, err)
```
johnramsden, line 422: **The RGW pass aborts the whole batch on the first per-member dispatch failure.**

This loop returns on the first error, before the disable loop below starts, and the disable loop does the same. One rebooting or unreachable member blocks every unrelated enable *and* the entire scale-down set in that PUT - and because iteration is deterministically sorted, every retry hits the same member first.

Since the policy is only stored on success, the declared intent also stays stale while the cluster has already partially moved. Accumulating per-member errors and continuing would match the partial-apply semantics the keep-one path already has.

microceph/ceph/placement.go
```diff
@@ -271,6 +293,16 @@ func ApplyPlacement(ctx context.Context, s interfaces.StateInterface, policy typ
 		}
 	}
 
+	// RGW pass (CE142 Option B): reconcile role-managed RGW placement after the
+	// control pass so control quorum is established first. RGW has no keep-one
+	// invariant (scale-to-zero is allowed); it follows add-before-remove so a
+	// migration keeps a gateway serving. An omitted rgw field (nil) leaves the
+	// member untouched; enabled:false removes RGW only where it is observed.
+	err = applyRgwPlacement(ctx, s, policy)
```
johnramsden, line 301: **A keep-one refusal is swallowed when the RGW pass also errors.**

This returns before the `refused` check at line 307, so when the control pass refused a removal *and* the RGW pass then fails, the control-service context is lost from both the response and the persisted `last_refusal`. The caller sees only the RGW error, and the policy is not stored as declared intent the way a plain keep-one refusal would be.

microceph/ceph/rgw_test.go
```diff
@@ -1,365 +1,303 @@
...
+	defer restorePaths()
+	_, restoreInj := setupRGWInjectables(t)
+	defer restoreInj()
+
+	changed, err := applyRGWFrontend(nil, 0, 443, "not-base64!!", b64(t, "key1"), []string{"mon1"})
+	require.Error(t, err)
+	assert.False(t, changed)
+	assert.Contains(t, err.Error(), "SSL certificate", "error must identify the bad material")
+	assert.ErrorIs(t, err, ErrRgwFrontendInvalid, "malformed TLS must map to a client-side error")
 }
```
johnramsden, line 303: **Test coverage regression in this rewrite.**

The file drops 313 lines and adds 251. Gone with them:

- `TestUpdateRGWCertificates` and `TestUpdateRGWCertificatesWhenRGWNotActive`. `UpdateRGWCertificates` is still live via `api/certificates.go`, and those were the only tests anywhere pinning `server.crt`/`server.key` to mode 0600.
- Direct coverage of `DisableRGW`, whose new `removeIgnoreMissing` idempotency is one of this PR's headline claims.
- `RestartRGW`, and `writeSSLFiles`' cleanup-on-failure path - the one bad-base64 test now fails inside `applyRGWFrontend` before `writeSSLFiles` is ever reached.

The new `applyRGWFrontend` tests are good; they just don't replace what was removed. The existing mocks/injectable-func pattern should make restoring these cheap.

microceph/database/schema.go
```diff
@@ -296,3 +297,29 @@ INSERT INTO placement_policy (id) VALUES (1);
...
+// schemaUpdate10 adds the rgw_frontends table (CE142 placement-rgw). It records
+// the observed RGW beast frontend (port, ssl_port, and a TLS on/off flag) for
+// each member hosting RGW, so GET /1.0/placement can report observed frontend
+// state from dqlite alongside the rest of its observed state, without a
+// per-member fan-out. Only ports and the TLS flag are stored — never cert/key
+// bytes, which remain on disk in server.crt/server.key (0600). The row is
+// cascade-deleted when the cluster member is removed, mirroring the services
+// and host_tags tables. At most one row per member is enforced by UNIQUE on
+// member_id.
+func schemaUpdate10(ctx context.Context, tx *sql.Tx) error {
```
johnramsden, line 310: **No backfill for RGW enabled before this schema update.**

`schemaUpdate10` creates `rgw_frontends` empty, and `RgwServicePlacement.DbUpdate` is the only path that ever writes a row. A member already running RGW from the CLI path therefore reports `rgw: true` with no `rgw_frontend` until it happens to be re-enabled, and `populateRGWFrontends` skips it silently - which on the wire is indistinguishable from "not TLS".

A backfill in this update (ports unknown, but the row could be seeded from the services table) or an explicit "unknown" signal in the observed output would avoid the ambiguity.

microceph/database/rgw_frontend.go
```diff
@@ -0,0 +1,83 @@
...
+		var f RgwFrontend
+		var sslInt int
+		err := rows.Scan(&f.Member, &f.Port, &f.SSLPort, &sslInt)
+		if err != nil {
+			return nil, fmt.Errorf("failed to scan rgw frontend: %w", err)
+		}
+		f.SSL = sslInt != 0
+		frontends = append(frontends, f)
+	}
+	if err := rows.Err(); err != nil {
```
johnramsden, line 67: Nit: one-line assign/test, which `AGENTS.md` forbids - `err := rows.Err()` then `if err != nil`. Same pattern at `ceph/rgw.go:421` (`if err := os.Symlink(...)`).

docs/snap/reference/declarative-placement.rst
```diff
@@ -0,0 +1,221 @@
...
+--------------
+
+.. list-table::
+   :header-rows: 1
+
+   * - Code
+     - Meaning
+   * - 400
+     - The request cannot be satisfied: Ceph is not bootstrapped, the policy
+       names an unknown member, the RGW frontend configuration is malformed, or
```
johnramsden, line 180: **The documented 400 for malformed TLS is a 500 in practice.**

`ErrRgwFrontendInvalid` is raised inside `applyRGWFrontend`, which runs on the *target member*. It is flattened twice before the handler can test it:

1. `ceph/services_placement.go:112` wraps with `%v`, not `%w`.
2. `client.SendServicePlacementReq` then crosses an HTTP boundary, which reconstructs a generic error regardless.

So `errors.Is` in `isClientSidePlacementError` cannot match, and `SmartError` falls back to 500. The test covering this (`api/placement_test.go:740`) stubs `ApplyPlacementFunc` and returns the wrapped sentinel directly, so it passes either way.

Fixing the `%v` and mapping `ErrRgwFrontendInvalid` to 400 on the receiving member's handler would make the doc true; a test that goes through the real dispatch would keep it that way.

johnramsden (commented):
## AI review

The shape of this is good: add-before-remove with sorted iteration, the frontend-line diff that stops mon-host churn from restarting RGW, strip-at-store plus redact-at-get for the key material, and observed state read from dqlite in the same transaction as the rest of `GetPlacementStatus`. `go build`, `go vet` and `go test ./ceph/... ./api/... ./database/...` all pass at `be69165`. The gaps are mostly at the edges the design didn't close: a wire-shape break with no upgrade path, several TLS/port misconfigurations that are accepted silently, and a reconcile pass that doesn't isolate per-member failure.

Findings are inline. Two more could not be anchored inline because they land in `microceph/ceph/services_placement.go`, which this PR does not touch:

**RGW private key and certificate are logged in cleartext** - `microceph/ceph/services_placement.go:44`

`logger.Debugf("Enabling %s service, payload: %v", payload.Name, payload.Payload)` prints the marshalled `RgwServicePlacement`, which includes `SSLPrivateKey`. The line is pre-existing, but this PR is what puts key material on an automated, repeating path through it - every enable, port change and cert rotation driven by the reconcile pass, rather than an occasional manual CLI call. It sits directly against the "never persisted, write-only" posture the PR documents. Redacting the payload for services that carry secrets seems in scope here.

**Pre-existing, but now on the rotation path** - `writeSSLFiles` renames cert and key as two independent operations, so a failure between them leaves a mismatched pair on disk that the next restart picks up. Cert rotation via placement exercises this far more often than the CLI did.

---

*Generated by an AI assistant (Claude Opus 5) using a multi-agent review pass; every finding was re-checked by hand against the code at `be69165` before posting. It deliberately excludes the three points already raised in review (the "Option B" naming, the suite not being wired into `tests.yml`, and the `genericHospitalityCheck` non-idempotency). Please verify before acting.*

tests/robot/rgw-placement-tests/rgw_placement_tests.robot
```diff
@@ -0,0 +1,108 @@
...
+Test Placement Reports Observed RGW Frontend And No Secret Leak
+    [Documentation]    GET /1.0/placement reports the observed rgw_frontend for the RGW
+    ...    member with port=8080 and ssl=false (ssl_port omitted for plaintext), and the
+    ...    stored policy carries no ssl_certificate / ssl_private_key material.
+    [Tags]    rgw    placement
+    ${placement}=    Get Placement Status JSON
+    # Observed frontend: port 8080, ssl false, no ssl_port (plaintext -> ssl_port=0 -> omitempty).
+    Should Contain    ${placement}    "rgw_frontend":{"port":8080,"ssl":false}    msg=observed rgw_frontend not reported as expected: ${placement}
+    # Defense in depth: no SSL key material anywhere in the placement body.
+    Should Not Contain    ${placement}    ssl_certificate    msg=ssl_certificate leaked into placement body
```
johnramsden, line 96: **This assertion never exercises the code path it claims to test.**

The enable PUT above sends `{"enabled":true,"port":8080}` - no TLS material at any point in the suite. So `Should Not Contain ... ssl_certificate` passes on a policy that never contained a secret, and would still pass with `PolicyForStorage` and `redactStoredPolicy` deleted outright.

Sending a real (throwaway) cert/key pair in the enable PUT and then asserting absence is what makes this a defence-in-depth check.

Also: `placement_status.member_rgw_frontend` and `placement_status.placement_leaks_rgw_secrets` are added in this PR with pytest coverage, but nothing outside those tests calls them - this suite uses raw-JSON substring matching instead, including `"rgw_frontend":{"port":8080,"ssl":false}`, which is sensitive to Go's struct field ordering. That is the opposite of the "fetch raw, decide in Python" rule the module's own docstring cites.

tests/robot/rgw-placement-tests/rgw_placement_tests.robot
```diff
@@ -0,0 +1,108 @@
...
+Test Enable RGW Via Placement Object
+    [Documentation]    PUT /1.0/placement with `rgw:{enabled:true,port:8080}` on the
+    ...    local member atomically enables RGW on port 8080 (engine -> member dispatch
+    ...    -> applyRGWFrontend render + start). The bare-bool form is rejected.
+    [Tags]    rgw    placement
+    ${hn}=    Get VM Hostname
+    # Bare-bool rgw must be rejected (HTTP 400) by the Option B parser.
+    ${bad}=    MicroCeph API Put    placement    {"mode":"reconcile","members":{"${hn}":{"rgw":true}}}    timeout=60
+    ${bad_code}=    Response Status Code    ${bad}
+    Run Keyword And Continue On Failure    Should Not Be Equal As Integers    ${bad_code}    200    msg=bare-bool rgw must be rejected, got ${bad}
```
johnramsden, line 78: Nit: this asserts only "not 200", so it would pass on a 500 or a 409 just as happily as on the documented 400. `Should Be Equal As Integers ${bad_code} 400` is what the docs promise.

Separately, on harness conventions in `AGENTS.md`: `Placement Has No RGW Frontend` hand-rolls a `FOR`/`Sleep` loop and `RGW Snap Service Is Not Active` hand-rolls an in-VM shell poll loop, where the rules call for `_poll_until` in Python for every poll loop.

microceph/api/types/placement.go
johnramsden: Do not address

johnramsden:
Overall looks pretty good to me. 

I left a couple comments earlier and then executed an AI review. My only main concern is https://github.com/canonical/microceph/pull/814#discussion_r3771312761, along with the tests not currently executing https://github.com/canonical/microceph/pull/814#pullrequestreview-4922234048. 

Feel free to just resolve/skip anything the AI came up with

PR #815, opened 2026-08-12
Implement the storage role interfaces

lmlg:
This changeset implements the necessary changes to support the storage role such that the necessary checks are performed before enrolling OSD's to make sure that the target node has the necessary capability.

### DIFF
--- a/microceph/ceph/osd.go
+++ b/microceph/ceph/osd.go
@@ -666,6 +666,11 @@ func (m *OSDManager) addLoopBackOSDs(ctx context.Context, spec string) error {
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


microceph/ceph/osd.go
sabaini, line 664: think we should check for role/placementeligibility here as well

sabaini, line 925: check for role/placementeligibility here as well

sabaini (changes requested):
Hey @lmlg thanks, lgtm in general.

Added two places where we want to check in addition
Also would suggest to add a robot test the check behaviour

sabaini (approved):
lgtm, thanks!

PR #817, opened 2026-08-14
fix: control role mon quorum preservation

sabaini:
- Remove the MON from the committed monmap before stopping its daemon.
- Verify the resulting monmap retains enough active monitors to form quorum.
- Make monitor removal idempotent and safe to retry after partial teardown failures.
- Serialize service placement and deletion with other service lifecycle operations.
- Reuse a common helper for reading MON quorum membership.
- Added unit/robot tests.

Assisted-by: pi:z-ai/glm-5.2, claude:opus-5

### DIFF
--- a/microceph/ceph/control_removal.go
+++ b/microceph/ceph/control_removal.go
@@ -6,12 +6,21 @@ import (
 )
 
 // controlDaemonRemovalVerifyTimeout bounds the post-eviction verification poll
-// for MGR/MDS map convergence. It comfortably exceeds the default Ceph beacon
-// grace (mon_mgr_beacon_grace, 30s) so that even when the active eviction
-// command is a partial no-op for a standby daemon, verification still succeeds
+// for MGR/MDS map convergence. It must comfortably exceed the worst-case
+// beacon-aging fallback so that even when the active eviction command is a
+// partial no-op (a standby daemon) or is rejected, verification still succeeds
 // via beacon aging within the bounded window rather than returning a spurious
-// error. It is a var (not a const) so unit tests can shrink it.
-var controlDaemonRemovalVerifyTimeout = 60 * time.Second
+// error.
+//
+// The window is sized with generous margin over that fallback, not merely equal
+// to it: the default grace periods are mon_mgr_beacon_grace (30s) and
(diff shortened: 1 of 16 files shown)


microceph/ceph/control_removal.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"context"
+	"time"
+)
+
+// controlDaemonRemovalVerifyTimeout bounds the post-eviction verification poll
+// for MGR/MDS map convergence. It comfortably exceeds the default Ceph beacon
+// grace (mon_mgr_beacon_grace, 30s) so that even when the active eviction
+// command is a partial no-op for a standby daemon, verification still succeeds
+// via beacon aging within the bounded window rather than returning a spurious
```
UtkarshBhatthere, line 12: The verification timeout is too short for the fallback path. `ceph mds fail` can reject a request when the cluster has MDS health warnings, and this command does not pass `--yes-i-really-mean-it`. In a two-MON cluster, removing one MON creates a new quorum; Ceph can then wait up to 60 seconds before removing a stopped MDS automatically. This timeout is also 60 seconds, so the request can fail just before the MDS is removed and skip cleanup. Please add some margin to this timeout.

tests/robot/resources/placement_status.py
```diff
@@ -89,11 +89,99 @@ def mon_count(raw):
...
+
+
+def control_service_presence(mon_raw, mgr_raw, mds_raw, member):
+    """Return explicit MON/MGR/MDS membership for *member* from Ceph JSON.
+
+    Invalid responses fail closed for each service. This is stricter than a
+    substring search over ``ceph -s``: a hostname appearing anywhere in status
+    does not prove that all three role-managed control services are present.
+    """
+    mgr = _parse(mgr_raw)
```
UtkarshBhatthere, line 155: Invalid JSON is treated as “service absent”. For the `expected=no` check, this poll can pass when `ceph` returned bad or empty output, rather than when the service was actually removed. Please make invalid JSON fail or retry instead of treating it as absence.

UtkarshBhatthere (approved):
Lgtm, thanks @sabaini

PR #819, opened 2026-08-18
ci: use larger runner for deferred multinode tests

sabaini:


johnramsden (approved):
LGTM, I believe RTD issue is due to migration.

PR #820, opened 2026-08-19
fix: make placement policy a full-replacement snapshot and move apply orchestration

sabaini:
# Description

1. Split ApplyPlacement into validate / store / reconcile

Orchestration used to live in the HTTP handler, which decided when to persist and special-cased ErrKeepOneInvariant. It's now ceph.ApplyPlacementPolicy driving three phases: ValidatePlacement (preconditions only — cert, pre-bootstrap guard, unknown members), storePlacementPolicy, reconcilePlacement (diff, adds before removals, keep-one safety).

Ordering: validate before persist, so an unappliable policy doesn't replace stored state; persist before reconcile, so a policy that fails to converge stays observable via GET /1.0/placement with last_refusal. This generalises the old keep-one special case to any mid-reconcile failure. Refusal writes moved into ceph with a detached 30s context, so a deadline-exceeded apply can still record why. StorePlacementPolicyFunc and SetPlacementRefusalFunc are unexported — the API layer can't bypass the sequence. The handler is now decode → lock → apply → map to 400/500.

2. Document the policy as a full-replacement snapshot

PUT overwrites wholesale; it never merges. Doc comments across types, api, ceph, and database now say so consistently, including that the empty-members waiting policy is an empty allow-list (not inert) and that DELETE returns eligibility to unmanaged. 

## Type of change

## How has this been tested?

unit tests

## Contributor checklist

Please check that you have:

### DIFF
--- a/microceph/api/placement.go
+++ b/microceph/api/placement.go
@@ -26,7 +26,9 @@ var placementCmd = mcTypes.Endpoint{
 	Delete: mcTypes.EndpointAction{Handler: cmdPlacementDelete, ProxyTarget: true},
 }
 
-// cmdPlacementGet returns the current placement status.
+// cmdPlacementGet returns the current placement status: the canonical desired
+// policy as stored (the complete snapshot from the last accepted PUT), the
+// observed placement, and the lifecycle/refusal state.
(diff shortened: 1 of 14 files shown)


lmlg (approved):
lgtm, good stuff

PR #821, opened 2026-08-22
feat: rgw replication status

johnramsden:
# Description

Part 02 of #809 

Add the plumbing and relevant API for a status subcommand. 

--- 

Note: We are pausing on this features development so we are not moving forward with creating the user facing part yet, but we are merging what we have to avoid having to come back to it in the future when drift has occurred. Everything here is read only, and will not be interacted with The intention of this PR is the place it in so that we can continue working in the future when we come back to the future.

As of this point we just have read wrappers to facilitate executing rgw commands, and the API for status.

This does not support bucket status, which will require more read wrappers.

Next steps when we pick this up again: 

- CLI command (separate branch)
- `list`
- DB table, write wrappers
- `enable` / `disable`
- sync policy, `configure`
- promote / failover, bucket status
- integration tests, docs

## Type of change

## How has this been tested?

* Unit tests

## Contributor checklist

Please check that you have:

### DIFF
--- a/microceph/ceph/replication_rgw.go
+++ b/microceph/ceph/replication_rgw.go
@@ -109,8 +109,11 @@ func (rh *RgwReplicationHandler) PreFill(ctx context.Context, request types.Repl
 
 // preFillSyncStatus reads this zone's own sync markers: its metadata progress,
 // and its data progress against every source zone it is configured to pull
-// from. All of it is local progress with no peer contact, so it is safe to
-// read before the state machine has decided anything.
+// from. The metadata read is purely local, but `data sync status` first
+// fetches the source zone's datalog shard info over HTTP: a source that
+// refuses connections fails the read immediately (surfaced per stream
+// below), while a blackholed one blocks it for radosgw-admin's full 300s
+// curl timeout.
 //
 // A read whose command failed outright is recorded per stream rather than
 // failing the whole request, mirroring how an unreadable peer log is already
--- a/microceph/ceph/replication_rgw.go
+++ b/microceph/ceph/replication_rgw.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"slices"
 
 	"github.com/canonical/microceph/microceph/api/types"
 	"github.com/canonical/microceph/microceph/database"
@@ -107,9 +108,9 @@ func (rh *RgwReplicationHandler) PreFill(ctx context.Context, request types.Repl
 }
 
 // preFillSyncStatus reads this zone's own sync markers: its metadata progress,
-// and its data progress against every other zone in the zonegroup. All of it
-// is local progress with no peer contact, so it is safe to read before the
-// state machine has decided anything.
+// and its data progress against every source zone it is configured to pull
+// from. All of it is local progress with no peer contact, so it is safe to
+// read before the state machine has decided anything.
 //
 // A read whose command failed outright is recorded per stream rather than
 // failing the whole request, mirroring how an unreadable peer log is already
@@ -135,7 +136,7 @@ func (rh *RgwReplicationHandler) preFillSyncStatus() error {
 
 	rh.DataSync = map[string]RgwDataSyncStatus{}
 	rh.DataSyncUnavailable = map[string]bool{}
-	for _, zone := range rh.peerZones() {
+	for _, zone := range rh.dataSyncSourceZones() {
 		status, err := GetRgwDataSyncStatus(zone.Name, "", "")
 		if err != nil {
 			if !errors.Is(err, ErrRgwSyncStatusUnreadable) {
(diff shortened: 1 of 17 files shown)


UtkarshBhatthere (changes requested):
I think these need fixing before merge:

**[P1] Use the realm metadata master, not the local zonegroup master**  
`isMasterZone()` only compares the local zone with the local zonegroup’s `master_zone`. In a realm with multiple zonegroups, the master of a non-master zonegroup will incorrectly skip metadata-sync status and be reported as `master`. The metadata source needs to be the master zone in the realm’s master zonegroup.

**[P2] Do not query every zonegroup peer as a data-sync source**  
The status code calls `data sync status --source-zone` for every peer zone. RGW can be configured with directional `sync_from` relationships or sync policies, so not every peer is necessarily a source for this zone. Non-source peers can therefore appear as unavailable or behind even though no data stream exists.

**[P2] Preserve failed local status reads**  
The RGW wrappers turn a failed local `radosgw-admin ... sync status` command into a zero value with no error. When the peer log is available, that zero value falls through to `behind`, even though no local sync state was read. This should be surfaced as an unavailable/error result instead.

Targeted check run successfully:

```text
go test -count=1 ./ceph ./api ./api/types
```

johnramsden:
Rebased to grab #825

UtkarshBhatthere (changes requested):
Some minor nits here and there.

microceph/ceph/replication_rgw.go
```diff
@@ -0,0 +1,537 @@
...
+
+	remote, ok := remotes[masterZone]
+	if !ok {
+		// Without a remote for the master cluster its metadata log cannot
+		// be read at all, which is not the same as being caught up.
+		logger.Warnf("REPRGW: no remote is imported for master zone %q, its metadata log cannot be read", masterZone)
+		return summariseRgwSyncVerdict(masterZone, "", rh.MetadataSync.Info, RgwSyncVerdict{PeerLogUnavailable: true})
+	}
+
+	masterLog, err := GetRgwMdlogStatus(remote.Name, remote.LocalName)
```
UtkarshBhatthere, line 280: **Read the master mdlog for the same period**

`GetRgwMdlogStatus()` does not pass a period, so the remote uses its current period. During a period update, the local zone may still be on the old period. This can compare markers from different periods and incorrectly report `caught-up` or `behind`.

Please pass the local sync period to the mdlog command, or compare the master’s period before comparing markers.

microceph/ceph/replication_rgw.go
```diff
@@ -0,0 +1,537 @@
...
+// unreadable rather than guessed at.
+func getRgwRemotesByZone(ctx context.Context, st interfaces.CephState) (map[string]types.RemoteRecord, error) {
+	records, err := database.GetRemoteDb(ctx, st.ClusterState(), "")
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch the imported remotes: %w", err)
+	}
+
+	remotes := make(map[string]types.RemoteRecord, len(records))
+	for _, record := range records {
+		remotes[record.Name] = record
```
UtkarshBhatthere, line 488: **Document or enforce the remote-to-zone naming requirement**

This lookup only works when the imported remote name exactly matches the RGW zone name. That requirement is not currently documented or enforced.

microceph/ceph/replication_rgw.go
```diff
@@ -0,0 +1,537 @@
...
+// check (RGWZone::syncs_from) before reporting a data stream.
+func (rh *RgwReplicationHandler) isDataSyncSource(peer RgwZoneGroupZone) bool {
+	local, ok := rh.localZoneGroupZone()
+	if !ok {
+		// An orphaned zone is already reported as disabled replication;
+		// hiding every peer here would dress that up as topology instead.
+		return true
+	}
+
+	return local.SyncFromAll || slices.Contains(local.SyncFrom, peer.Name)
```
UtkarshBhatthere, line 452: **Include sync policies when finding source zones**

This only checks `sync_from_all` and `sync_from`. RGW also uses the active sync policy to decide which zones are data sources.

With a directional policy, this code may query and report status for a stream that does not exist. Please use RGW’s effective source-zone configuration.

johnramsden:
@UtkarshBhatthere As I'm going through your reviews I am seeing a pattern in the areas you are finding issues. 

Context: Unfortunately Upstream does not support json output for sync status. As a result I essentially broke it up into many little calls that do support json. This translated to essentially reimplementing the upstream functionality on our end. Originally I did this because I did not want to introduce the potential drift in case the status output changes. 

I'm starting to think it introduces too much complexity on our end. Curious about your opinion here. I'm leaning toward restructuring some of the work I did in the previous PR, along with some of the work done here (this would address the majority of your reviews that you've brought up here by simply using what Upstream has). 

The downside we end up with here is we have to parse text output. But I think this is a reasonable downside to reduce complexity and maintenance. Based on a rough estimation I think this would remove almost 2,000 lines of code, and add back about 800 that is mostly simple text parsing. For what it's worth this is what the MGR dashboard does (text parsing) https://github.com/ceph/ceph/blob/35ae58662739d4fa9ffd16be32814ec8997b3a71/src/pybind/mgr/dashboard/services/rgw_client.py#L2843

PR #822, opened 2026-08-24
fix: tie monitor host config to member removal

### LINKED ISSUE #444: can not join additional node to cluster
# Issue report

18.2.4 reef (stable)

Use this section to describe the channel/revision which produces the unexpected behaviour.

I had to forcefully remove a node due to a hardware failure. Afterwards I wanted to join a new node with the previous node name and ip address, unfortunately it run in a timeout. Now I'm getting the following error: 
Error: failed to record mon db entries: failed to record mon host: This "config" entry already exists
I then tried to add a new node with a new ip address and name, unfortunately same result.

1. install new node with 'sudo snap install microceph --channel reef/stable'
2. create token on a cluster node with 'sudo microceph cluster add <node>'
3. join cluster on new node with 'sudo microceph cluster join <token>'
4. Error: failed to record mon db entries: failed to record mon host: This "config" entry already exists

Error: failed to record mon db entries: failed to record mon host: This "config" entry already exists
…

new node should have joined the cluster
…

If it’s considerably long, please paste to https://gist.github.com/ and insert the link here.

## Additional comments.
I've realized that in the ceph.conf entry 'mon host = ' only the ip of the new hosts are present, the ip's of the two currently operating nodes are missing
…

UtkarshBhatthere:
## Summary
- Tie member-named `mon.host.*` configuration to MicroCluster member deletion with a Dqlite schema trigger, so forced removal does not depend on an unreachable member running `PreRemove` or `PostRemove`.
- Upsert the joining monitor's config entry so an old same-name entry is repaired rather than causing a unique-key conflict.
- Render only active member-backed monitor entries while preserving numeric adopted-Ceph entries; remove the broad best-effort post-removal reconciliation hook.

Fixes #444

## Testing
- `make check-unit`
- `golangci-lint run --timeout 5m`
- `go build ./...`
- `git diff --check`

## Notes
- `make check-static` remains blocked locally because `shellcheck` is not installed; the direct `golangci-lint` invocation passed.
- The PR remains draft while CI runs.

### DIFF
--- a/microceph/ceph/services_placement_rgw.go
+++ b/microceph/ceph/services_placement_rgw.go
@@ -3,7 +3,6 @@ package ceph
 import (
 	"context"
 	"encoding/json"
-	"fmt"
 
 	"github.com/canonical/microceph/microceph/interfaces"
 )
@@ -30,18 +29,12 @@ func (rgw *RgwServicePlacement) HospitalityCheck(s interfaces.StateInterface) er
 }
 
 func (rgw *RgwServicePlacement) ServiceInit(ctx context.Context, s interfaces.StateInterface) error {
-	// fetch configs from db
-	config, err := GetConfigDb(ctx, s)
+	monitorAddresses, err := GetMonitorAddresses(ctx, s)
 	if err != nil {
-		return fmt.Errorf("failed to get config db: %w", err)
+		return err
 	}
 
-	activeMonitorMembers, err := getActiveMonitorMembersFunc(ctx, s)
-	if err != nil {
-		return fmt.Errorf("failed to get active monitor members: %w", err)
-	}
-
-	return EnableRGW(s, rgw.Port, rgw.SSLPort, rgw.SSLCertificate, rgw.SSLPrivateKey, getMonitorsFromConfig(config, activeMonitorMembers))
+	return EnableRGW(s, rgw.Port, rgw.SSLPort, rgw.SSLCertificate, rgw.SSLPrivateKey, monitorAddresses)
 }
 
 func (rgw *RgwServicePlacement) PostPlacementCheck(s interfaces.StateInterface) error {
--- a/microceph/ceph/services_placement_rgw.go
+++ b/microceph/ceph/services_placement_rgw.go
@@ -36,7 +36,12 @@ func (rgw *RgwServicePlacement) ServiceInit(ctx context.Context, s interfaces.St
 		return fmt.Errorf("failed to get config db: %w", err)
 	}
 
-	return EnableRGW(s, rgw.Port, rgw.SSLPort, rgw.SSLCertificate, rgw.SSLPrivateKey, getMonitorsFromConfig(config))
+	activeMonitorMembers, err := getActiveMonitorMembersFunc(ctx, s)
+	if err != nil {
+		return fmt.Errorf("failed to get active monitor members: %w", err)
+	}
+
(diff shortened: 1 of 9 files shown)


UtkarshBhatthere:
@sabaini i am under the impression that the failing test is a known trouble maker which I will be working on in a separate PR.

microceph/database/schema.go
```diff
@@ -297,3 +297,26 @@ INSERT INTO placement_policy (id) VALUES (1);
...
+// when MicroCluster removes that member. Numeric config suffixes are retained
+// because they are used for monitor entries adopted from an external Ceph
+// cluster rather than names of MicroCeph members.
+func schemaUpdate10(ctx context.Context, tx *sql.Tx) error {
+	tableName, err := getClusterTableName(ctx, tx)
+	if err != nil {
+		return err
+	}
+
+	stmt := fmt.Sprintf(`
```
johnramsden, line 312: I don't believe these triggers will ever run on clusters where members have already been removed forcefully. To deal with the existing stragglers I would recommend some sort of migration step in the schema update that cleans up those stragglers. 

This shouldn't cause any issues for joining, since you have the upsert mechanism. I'm just concerned it could be a footgun for anything consumes mons based on the database list (currently or in the future).

microceph/ceph/services_placement_rgw.go
```diff
@@ -36,7 +36,12 @@ func (rgw *RgwServicePlacement) ServiceInit(ctx context.Context, s interfaces.St
 		return fmt.Errorf("failed to get config db: %w", err)
 	}
 
-	return EnableRGW(s, rgw.Port, rgw.SSLPort, rgw.SSLCertificate, rgw.SSLPrivateKey, getMonitorsFromConfig(config))
+	activeMonitorMembers, err := getActiveMonitorMembersFunc(ctx, s)
+	if err != nil {
+		return fmt.Errorf("failed to get active monitor members: %w", err)
+	}
+
+	return EnableRGW(s, rgw.Port, rgw.SSLPort, rgw.SSLCertificate, rgw.SSLPrivateKey, getMonitorsFromConfig(config, activeMonitorMembers))
```
johnramsden: Not blocking observation: This was present prior to this PR but everywhere else that calls getMonitorsFromConfig also checks backwardCompatMonitors, since we're changing it here it might be worth checking them and erroring if we can't find any to avoid populating with empty monitors. 

https://github.com/canonical/microceph/blob/79f99b50a1abf05ca8abb463b0fcb3f749f13c89/microceph/ceph/config.go#L500-L507

johnramsden (changes requested):
(Not blocking) Suggest regression test: Could be worth having a regression test for the --force behavior to ensure the trigger continues to function. Otherwise potentially a change to the database could result in triggers not firing and mons accumulating again in the database.

johnramsden:
Overall looks really good. A couple not blocking nits, and one concern for existing cluster regarding cleanup.

microceph/database/schema.go
UtkarshBhatthere: Resolved in c41b2bbf2275cd1ae97639fd32053133c65ddd1f

microceph/ceph/services_placement_rgw.go
UtkarshBhatthere: Resolved in c41b2bbf2275cd1ae97639fd32053133c65ddd1f

PR #829, opened 2026-09-04
fix: Deferred placement mon removal flake

### LINKED ISSUE #828: Deferred placement mon removal flake
# Issue report

Title: Mon removal verify hangs when the removed node's `mon host` is stale, making the deferred placement CI job flaky

Snap built by CI from `main` at commit 79f99b50a1abf05ca8abb463b0fcb3f749f13c89 and later. The same failure shows on every PR branch that has run the `Deferred-Ceph lifecycle and placement multi node` job since https://github.com/canonical/microceph/pull/817 merged on 2026-08-18.

1. Bootstrap MicroCluster on two nodes with `--defer-ceph`, then run `microceph cluster bootstrap-ceph --target node-wrk1`.
2. Add control services to `node-wrk0` with `PUT /1.0/placement` and body `{"mode":"reconcile","members":{"node-wrk0":{"control":true}}}`.
3. Within about a minute, move control off `node-wrk1` with `PUT /1.0/placement` and body `{"mode":"reconcile","members":{"node-wrk0":{"control":true},"node-wrk1":{"control":false}}}`.

The Robot suite `tests/robot/deferred-ceph-multinode-tests/deferred_ceph_multinode_tests.robot` does exactly this. It fails about half the time in `Test Declarative Control Placement Migration Preserves Quorum`.

The migration request fails after about 32 seconds with:

```
failed to remove mon on node-wrk1: failed disabling service mon: Failed to send request to target "node-wrk1": failed to remove monitor membership "node-wrk1": monitor "node-wrk1" removal could not be verified: failed to read monitor map: Failed to run: ceph mon dump -f json: signal: killed
```

The mon was actually removed from the monmap. The Ceph cluster is left with one mon on `node-wrk0`. But the `mon` service row for `node-wrk1` stays in the database. The next placement request then reports mon on both nodes in its keep-one refusal.

The migration should succeed every time. When it works, it takes 6 to 9 seconds.

Failing run: https://github.com/canonical/microceph/actions/runs/33803231238/job/100807908215

Timeline from that run, all on `node-wrk1` unless noted, UTC:

```
20:45:47  node-wrk1 renders ceph.conf; only its own mon exists yet
20:46:17  node-wrk0 mon is written to the DB (on node-wrk0)
20:46:41  migration PUT is sent (on node-wrk0)
20:46:43  node-wrk1 mon is removed from the monmap and exits
20:46:47  node-wrk1 renders ceph.conf again, 4 s too late
20:47:13  ceph mon dump is killed after 30 s and the request fails
```

Other runs with the same failure on `main` (scheduled runs on the same commit): 98756928221, 99036958689, 99362024493. On the same commit, the scheduled runs on 2026-08-23 and 2026-08-30 passed. In every failing run, the ceph.conf render on `node-wrk1` came 1 to 4 s after its mon exited. In every passing run, it came before.

Full journals: (gist link)

## Additional comments.

Root cause. `DeleteService` for a mon is proxied to the node being removed. On that node, `ensureMonAbsent` in `microceph/ceph/monitor.go` runs `ceph mon rm <self>`. The local mon then exits and is not restarted. Right after that, `verifyMonAbsent` runs `ceph mon dump -f json` on the same node with a 30 s timeout. The `mon host` line in that node's `ceph.conf` is only refreshed by the background loop in `microceph/ceph/start.go`, which sleeps a full minute between checks. Nothing refreshes it when another node adds a mon. So if the loop has not run since the other mon was added, `mon host` lists only the mon that just exited. The `ceph` client has nothing to connect to and is killed at the deadline.

The race window is only a few seconds wide. The test reaches the removal 56 to 59 s after the post-bootstrap render, and the next render is 60 s after it. That is why the job looks like a coin flip.

Suggested fix. Call `updateConfigFunc` at the top of `DeleteService` in `microceph/ceph/services.go`, before `ensureMonAbsent`. This mirrors what `EnableService` already does in `microceph/ceph/services_placement.go`. The reconcile adds services before it removes any, so the DB already holds the new mon's address at that point. A stronger option is to take the other mons' addresses from the `ceph mon dump` that `ensureMonAbsent` already runs before the removal, and pass them with `-m` to both `ceph mon rm` and the verify.

Related: https://github.com/canonical/microceph/pull/817 added both the verify step and the test. https://github.com/canonical/microceph/pull/820 did not change the add-then-remove order.

johnramsden:
# Description

The source's ceph.conf is rewritten to list only its own MON, as the once-a-minute refresh leaves it after a recent MON add. The migration must then finish quickly and re-render mon host, which fails on every run until service teardown renders ceph.conf itself.

Fixes #828 

## Type of change

## How has this been tested?

Regression & unit test

## Contributor checklist

Please check that you have:

### DIFF
--- a/tests/robot/deferred-ceph-multinode-tests/deferred_ceph_multinode_tests.robot
+++ b/tests/robot/deferred-ceph-multinode-tests/deferred_ceph_multinode_tests.robot
@@ -32,6 +32,34 @@ Ceph Only Bootstrap Target And Verify
     Run In Container    node-wrk0    microceph.ceph -s    30
     Assert Member Has Control Services    ${target}    yes
 
+Get Public Network IP
+    [Documentation]    The container's address on the LXD public network, which is the
+    ...    address ceph.conf's mon host carries. `hostname -I` lists the management
+    ...    address first, so the public one is selected by subnet.
+    [Arguments]    ${container}
+    ${cidr}=    Get Public Network CIDR
+    ${res}=    Run In Container    ${container}    hostname -I    30
+    ${ip}=    Evaluate    next((a for a in """${res.stdout}""".split() if ipaddress.ip_address(a) in ipaddress.ip_network("""${cidr}""", strict=False)), "")    modules=ipaddress
+    Should Not Be Empty    ${ip}    msg=No address of ${container} on public network ${cidr}: ${res.stdout}
+    RETURN    ${ip}
+
+Pin Mon Host To Self
+    [Documentation]    Overwrite the container's ceph.conf `mon host` line with only its own
+    ...    address, reproducing a render from before a peer MON was added. The daemon's
+    ...    refresh loop only re-renders when the MON list changes, so once it has rendered
+    ...    the current list the pin holds until service teardown renders the file itself.
+    [Arguments]    ${container}
+    ${ip}=    Get Public Network IP    ${container}
+    Run In Container    ${container}    sed -i 's/^mon host = .*/mon host = ${ip}/' /var/snap/microceph/current/conf/ceph.conf    30
+    ${pinned}=    Get Ceph Conf Value    ${container}    mon host
+    Should Be Equal As Strings    ${pinned}    ${ip}    msg=Failed to pin mon host on ${container}: ${pinned}
+
+Assert Mon Host Includes
+    [Documentation]    The container's rendered ceph.conf `mon host` line names the given address.
+    [Arguments]    ${container}    ${peer_ip}
+    ${hosts}=    Get Ceph Conf Value    ${container}    mon host
+    Should Contain    ${hosts}    ${peer_ip}    msg=mon host on ${container} does not list ${peer_ip}: ${hosts}
+
 *** Test Cases ***
 Test Deferred Join Forms MicroCluster Without Ceph
     [Documentation]    `microceph cluster join --defer-ceph` joins MicroCluster but does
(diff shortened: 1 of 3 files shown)


tests/robot/deferred-ceph-multinode-tests/deferred_ceph_multinode_tests.robot
```diff
@@ -72,11 +100,26 @@ Test Declarative Control Placement Migration Preserves Quorum
...
+    # minute of the add and then stays idle until the MON list changes again.
+    Wait Until Keyword Succeeds    90s    5s    Assert Mon Host Includes    node-wrk1    ${dst_ip}
+    Pin Mon Host To Self    node-wrk1
+    ${start}=    Get Time    epoch
     ${resp}=    MicroCeph API Put In Container    node-wrk0    placement    ${policy}
     ${code}=    Response Status Code    ${resp}
     Should Be Equal As Integers    ${code}    200    msg=Control migration failed: ${resp}
+    ${end}=    Get Time    epoch
+    ${elapsed}=    Evaluate    ${end} - ${start}
+    Should Be True    ${elapsed} < 25    msg=Migration took ${elapsed}s: the MON removal verify ran against a stale mon host and timed out
```
sabaini: wonder if 25s for mon/mds/mgr removal might not be too tight for a potentially loaded CI system, i'd extend the grace period to >>120s the mgr/mds removal allows

sabaini (approved):
Hey @johnramsden one nit, but nothing to block over -- cheers!

sabaini (approved):
👍

Issue #836, opened 2026-09-16
Test harness preflight: probe Snap Store and archive reachability inside each LXD instance and fail fast with a labelled infra error


johnramsden:
# Issue report

Title: Test harness preflight: probe Snap Store and archive reachability inside each LXD instance and fail fast with a labelled infra error

- Affects: main, squid (same design, different implementation)
- Part of: https://github.com/canonical/microceph/issues/773
- Related: https://github.com/canonical/microceph/issues/838, https://github.com/canonical/microceph/issues/842, the ci-health dashboard change, to be filed in canonical/ceph-qa, https://github.com/canonical/microceph/issues/841
- Jobs: every suite that installs snaps or packages inside a fresh LXD VM or container

## What version of MicroCeph are you using ?

CI harness on `main` at `86c3250` and `squid` at `897bcdd9`.

## What are the steps to reproduce this issue ?

1. Snap Store or Ubuntu archive connectivity degrades from inside a freshly created LXD instance (2026-09-05 run 33938527624; Snap Store 408s on 2026-08-13, 08-20, 08-21, 09-11).
2. The suite's first `snap install` or `apt-get update` hangs until the harness timeout or errors, and setup failure cascades to every test.

## What happens (observed behaviour) ?

Nothing in the job distinguishes an unreachable store from a product or test defect; the failure text is a generic `Command timed out after 300s` or a snapd error buried in the log, and the job burns the full timeout before failing.

## What were you expecting to happen ?

The harness checks reachability from the instance that is about to install, fails within seconds when it is absent, and labels the failure so it can be recognised in the log, the run summary and by tooling.

## Relevant logs, error output, etc.

- https://github.com/canonical/microceph/actions/runs/33938527624 (run-wide stall), https://github.com/canonical/microceph/actions/runs/34553934187/job/103123918274 (Snap Store 408)
- main install points: `Setup LXD In VM` (`microceph_harness.resource`), `install_tools`, `build_base_lxd_image` (`microceph_harness.py`)
- squid install points: `actionutils.sh` lines 19, 28, 63-64 and the per-container installs at 504, 972, 985, 1025

## Additional comments.

[...]

PR #852, opened 2026-09-17
ci: gate the test matrix on a store/archive preflight probe

### LINKED ISSUE #841: Add a workflow-level preflight job so a gross Snap Store or archive outage skips the test matrix instead of failing every suite
# Issue report

Title: Add a workflow-level preflight job so a gross Snap Store or archive outage skips the test matrix instead of failing every suite

- Affects: main, squid (`tests.yml` on each branch)
- Part of: https://github.com/canonical/microceph/issues/773
- Related: https://github.com/canonical/microceph/issues/836 (complement; catches what this cannot), the ci-health dashboard change, to be filed in canonical/ceph-qa
- Jobs: all robot / multi-node test jobs

CI workflows on `main` at `86c3250` (`robot-tests` matrix, `needs: build-microceph`) and `squid` at `897bcdd9` (per-job shell steps).

1. The Snap Store or Ubuntu archive is down when a run starts.
2. Every test job starts, installs fail, and each job is recorded as a separate failure.

An outage produces one red job per suite (sixteen on 2026-09-05), each counted by the ci-health dashboard as a flake on that suite, with no single place that says the store was unreachable.

One clearly named job fails, the suites are not run, and the dashboard does not count them.

- https://github.com/canonical/microceph/actions/runs/33938527624
- Collector semantics: `collect.py` in canonical/ceph-qa counts only `success` and `failure` conclusions; `skipped` jobs are not counted

## Additional comments.

High-level approach. Add a `preflight` job in `tests.yml` that runs in parallel with `build-microceph` on a plain runner: curl the Snap Store info endpoint, `archive.ubuntu.com` and (main) the PPA with short timeouts and a few retries, emit `::error title=Infra::kind=preflight <message>` and a step-summary line on failure (same convention as the harness preflight). List it in `needs` of the matrix job (main) and of every test job (squid) so that when it fails the suites get conclusion `skipped`, the run shows one red job named for the cause, and the gate's failure count becomes the dashboard's outage metric. Limitation to state in the issue: the gate probes from the runner host minutes before the suites create their LXD instances, so it would not have caught the nested-instance stalls of 2026-09-05; the harness preflight covers that case.

johnramsden:
# Description

Add a workflow-level preflight job to `.github/workflows/tests.yml`, running alongside the snap build and required by the Robot test matrix. Persistent infrastructure failures are reported by the preflight job, while dependent suites are skipped instead of each recording a failure.

The job checks Snap Store, Ubuntu archive and Ceph PPA reachability with bounded exponential retries. It then downloads `hello-world` and verifies its assertions, fetches signed Ubuntu package indexes, and downloads `jq` and `s3cmd` without installing anything. Temporary APT state prevents cached files or installed package versions from affecting the check; host update hooks are disabled and temporary files are cleaned up on success and failure.

Failures produce an `Infra` annotation and matching step-summary message with the `kind=preflight` prefix. Native commands have explicit time limits and use tools already available on the Ubuntu runner.

This checks connectivity from the runner, not from the LXD instances created later. In-instance checks remain separate work tracked in https://github.com/canonical/microceph/issues/836. The native APT check covers Ubuntu repositories; the Ceph PPA retains its HTTP reachability probe.

Fixes https://github.com/canonical/microceph/issues/841

## Type of change

## How has this been tested?

Tested locally on Ubuntu 24.04 with curl 8.5.0, snap 2.76.3 and APT 2.8.3:

- Ran the complete preflight sequence against live services, including Snap payload/assertion retrieval and signed Ubuntu indexes followed by package downloads.
- Exercised curl recovery, persistent errors, connection refusals, stalled responses, exponential backoff and `Retry-After` handling.
- Injected Snap Store and APT index/package download failures; verified nonzero exits, matching annotations and summaries, and temporary-file cleanup.
- Exercised native-command timeout paths with shortened test deadlines, including cleanup of partial downloads.
- Checked legacy-format Ubuntu 22.04 sources using the local APT client; an installed-version negative control failed without status isolation and passed with it.
- Validated YAML and Bash syntax, and confirmed with `apt-config` that the inline configuration comments preserve the parsed settings and leave host update hooks disabled.

Verification used temporary smoke scripts; no permanent automated tests were added. Hosted GitHub Actions and the full Robot integration suite have not been run for this change.

## Contributor checklist

Documentation pages and release notes were not changed. Unchecked items are left for author review; the test item is unchecked because no permanent tests were added.

- [x] added or updated HTML meta descriptions for any new or modified documentation pages (see https://github.com/canonical/microceph/pull/643)


sabaini (approved):
thanks @johnramsden good addition!
i wonder if it would make sense to move the shell to a script e.g. actionutils.sh for easier reusability? not a blocker though

johnramsden:
@sabaini Moved to dedicated script

PR #853, opened 2026-09-17
ci: gate the squid test jobs on a store/archive preflight probe

johnramsden:
# Description

Add a workflow-level preflight job to `.github/workflows/tests.yml` on squid, running alongside the snap build and required by all 21 integration test jobs. Persistent infrastructure failures are reported by the preflight job, while dependent suites are skipped instead of each recording a failure; the static-check and unit-test jobs remain unchanged.

The job checks Snap Store and Ubuntu archive reachability with bounded exponential retries. It then downloads `hello-world` and verifies its assertions, fetches signed Ubuntu package indexes, and downloads `jq` and `s3cmd` without installing anything. Temporary APT state prevents cached files or installed package versions from affecting the check; host update hooks are disabled and temporary files are cleaned up on success and failure.

Failures produce an `Infra` annotation and matching step-summary message with the `kind=preflight` prefix. Native commands have explicit time limits and use tools already available on the Ubuntu runner.

This checks connectivity from the runner, not from the LXD instances created later. In-instance checks remain separate work tracked in https://github.com/canonical/microceph/issues/836. Unlike the main-branch counterpart, this workflow does not probe the Ceph PPA.

Fixes https://github.com/canonical/microceph/issues/841

## Type of change

## How has this been tested?

Tested the workflow scripts locally on Ubuntu 24.04 with curl 8.5.0, snap 2.76.3 and APT 2.8.3:

- Ran the Snap and APT download steps against live services, including Snap payload/assertion retrieval and signed Ubuntu indexes followed by package downloads.
- Exercised curl recovery, persistent errors, connection refusals, stalled responses, exponential backoff and `Retry-After` handling.
- Injected Snap Store and APT index/package download failures; verified nonzero exits, matching annotations and summaries, and temporary-file cleanup.
- Exercised native-command timeout paths with shortened test deadlines, including cleanup of partial downloads.
- Checked the legacy sources.list path against Ubuntu 22.04 repositories using the local APT client; an installed-version negative control failed without status isolation and passed with it.
- Validated YAML and Bash syntax, and confirmed with `apt-config` that the inline configuration comments preserve the parsed settings and leave host update hooks disabled.

Verification used temporary smoke scripts; no permanent automated tests were added. These checks were not an execution on an Ubuntu 22.04 runner; hosted GitHub Actions and the full integration suite have not been run for this change.

## Contributor checklist

Documentation and release notes were not changed. Unchecked items are left for author review; the test item is unchecked because no permanent tests were added.


sabaini (approved):
lgtm, thx!

PR #864, opened 2026-09-17
docs: add bcache to the custom wordlist

johnramsden:
# Description

The spelling check fails on "bcache" in the hyperconverged architecture page; it is the Linux block cache and is spelled correctly.

PR #865, opened 2026-09-17
docs: replace hardcoded readthedocs links with :ref: cross-references

johnramsden:
# Description

Two pages in the reference architecture linked to other MicroCeph pages by absolute readthedocs-hosted URL. The redirect to canonical.com/ceph/docs does not rewrite those paths, so both return 404 and the link check has failed on every run since 2026-08-12. Link to the existing page labels (release-notes, microceph-architecture) with :ref: instead, which keeps working when pages move or the domain changes.


skoech (approved):
LGTM, spellcheck failure fixed by #864. Thanks, @johnramsden!

PR #867, opened 2026-09-18
feat: Reimplement/role managed rgw

johnramsden:
# Description

Lets the declarative placement policy place RGW on members, with explicit TLS intent. Reimplements https://github.com/canonical/microceph/pull/814 on current main and resolves its review. Co-authored with Peter Sabaini.

Supersedes https://github.com/canonical/microceph/pull/814

### Changes from the original PR

- **Explicit intent.** `enabled` and `ssl` are required. TLS is never inferred from the certificate fields, so sending a GET response back, with its blanked certificate and key, cannot turn TLS off. A certificate without its key is a 400. Ports default to 80 and 443.
- **Safe member apply.** Idempotent, with a journal, a readiness check that pins the served certificate, and rollback. Each certificate and key pair lives in a directory named by its hash. A gateway already enabled by CLI is adopted by the first PUT that names its member. Only the frontends line of `radosgw.conf` is rewritten.
- **Reconciliation.** Every addition is attempted, failures are aggregated, and removals wait while any addition fails. A refusal to remove the last mon, mgr or mds stays in the response next to an operational failure. 400 and 500 survive the hop between members.
- **Capability gate.** `placement-rgw` is checked on each target member before an enable.
- **Tests.** Two Robot suites registered in CI replace 5 unregistered cases: 18 on one VM, and 9 across three VMs on the self-hosted xlarge runner. They use real certificates and a key-leak probe.
- Rebased onto https://github.com/canonical/microceph/pull/820.
- **No published docs page yet.** The feature is still being built, so the RGW contract lives in a `HACKING.md` section until it settles.

```mermaid
stateDiagram-v2
    [*] --> Validated: request passes checks
    Validated --> Published: old config journaled, TLS pair staged, config written
    Published --> Restarted: gateway started or restarted
    Restarted --> Ready: port answers and served certificate matches
    Ready --> Recorded: row saved in rgw_frontends
    Recorded --> [*]: journal cleared, old pair deleted, best effort
    Validated --> RolledBack: a file write fails
    Restarted --> RolledBack: not ready within 2 minutes
    Ready --> RolledBack: gateway stops, or database write fails
    RolledBack --> [*]: previous config and gateway restored
```

### Things to know

- A policy stored by current main with a boolean `rgw` cannot be decoded. `GET /1.0/placement` returns 500 until the next PUT or DELETE replaces it. This is left as is on purpose, see https://github.com/canonical/microceph/pull/814#discussion_r3771600272.
- The schema update means a partly refreshed cluster answers 503 until every member is refreshed.
- `ssl: true` without a certificate and key reuses the pair already on the member, and fails with 500 if there is none.
- Do not mix `microceph enable rgw` or `disable rgw` with a policy that declares `rgw` for that member. The next PUT reverts the manual change without warning.
- A first gateway start can use the whole 2 minute readiness wait. The client timeout is 5 minutes.

## Type of change

- Breaking change (fix or feature that would cause existing functionality to not work as expected): the `rgw` field of a stored placement policy changes from a boolean to an object, see "Things to know"

## How has this been tested?

- Go unit tests: 71 new or changed, covering validation, the member apply with rollback, TLS reuse and the flat-layout migration, cluster reconciliation, secret redaction and the `rgw_frontends` table. Fifteen planted bugs were each caught by at least one test.
- Robot Framework: `rgw-placement-tests` (18 cases, one VM) and `rgw-placement-multivm-tests` (9 cases, three VMs) are registered in `tests.yml`. Both pass on a snap built from this branch, run in LXD VMs.
- The existing Robot regression passed on an earlier build of this branch, except one `deferred-ceph-multinode-tests` case that fails the same way on main.

## Contributor checklist

Please check that you have:

- [x] checked and added or updated relevant documentation (an "RGW placement (in development)" section in `HACKING.md`; the public reference page is held back until the feature settles, so no documentation pages are added)


microceph/ceph/rgw.go
```diff
@@ -1,246 +1,704 @@
...
+	if changed || materialChanged || pending != nil {
+		rb = &rgwRollback{confPath: rgwConfPath(), prevConf: current, prevActive: wasActive}
+		if pending != nil {
+			rb.prevConf = []byte(*pending.PreviousConfig)
+			rb.prevActive = *pending.PreviousActive
+		}
+		previousCert, _ := rgwConfTLSPaths(rb.prevConf)
+		if previousCert != "" {
+			rb.prevGenDir = filepath.Dir(previousCert)
+		}
```
sabaini, line 448: i wonder what the interaction here is with a manual update of RGW tls certificates -- we document that we can restart rgw later, what happens if a manual restart is interleaved here, are we then rolling back to the wrong version? maybe worthwhile to gate on pending restarts, or at least document this should be avoided

sabaini (commented):
hey @johnramsden 
one worry wrt to the rgw conf is the interplay between role based rgw rollout/rollback and manual activation, eg. for tls material changes and things like deferred restart and rollback -- maybe also just some documentation update necessary to avoid surprises

microceph/ceph/rgw.go
johnramsden: I'm thinking we can simply check if the state is already represented correctly (someone already restarted the service). We can do this with a probe that tells us if it's already serving the new target, and then we can go ahead and adjust the rollback target. 

I'm going to be opening a new pull request with these stacked and I will incorporate it there.

johnramsden:
See https://github.com/canonical/microceph/pull/881

Issue #839, opened 2026-09-16
Upgrade tests break the OSD readiness poll on the first "3 up" reading and fail on one unguarded re-check


johnramsden:
# Issue report

Title: Upgrade tests break the OSD readiness poll on the first "3 up" reading and fail on one unguarded re-check

- Affects: main, squid
- Part of: https://github.com/canonical/microceph/issues/773
- Related: https://github.com/canonical/microceph/issues/686 (Reef, stalled on fix-or-remove), https://github.com/canonical/microceph/issues/712, https://github.com/canonical/microceph/pull/683 (`waitready --storage`)
- Jobs: squid Test quincy upgrades 6/29, squid Test reef upgrades 3/30, main Test reef upgrades 2/33 (renamed Test squid upgrades by PR 807)

## What version of MicroCeph are you using ?

CI builds from `squid` at `897bcdd9` (all nine squid failures) and from `main` at `79f99b5` (2026-08-30). The helper is unchanged at both branch tips.

## What are the steps to reproduce this issue ?

1. Run a multi-node upgrade job; `upgrade_multinode` (shell) or `upgrade_multi_node` (Python) installs the local snap on each node in turn.
2. After each install the helper polls `microceph.ceph osd status` until three OSDs are `exists,up`, breaks on the first success, then re-issues the same query once more with no retry to decide pass or fail.

## What happens (observed behaviour) ?

The poll succeeds on the first or second attempt (`Found 3 osd up`), and the lone re-query 1 to 7 seconds later sees a transient dip during the rolling restart and exits 1 with `Expected 3 OSD up, got 2`. The 200-second budget is never used. All nine squid failures and the 2026-08-30 main failure show this pair of lines.

## What were you expecting to happen ?

A transient dip immediately after a node's restart returns the check to waiting inside the existing budget instead of failing the job on one sample.

## Relevant logs, error output, etc.

- squid quincy: https://github.com/canonical/microceph/actions/runs/34301961720/job/102310763960 plus 2026-08-16, 08-20, 08-28, 08-29, 09-01
- squid reef: https://github.com/canonical/microceph/actions/runs/32546076430/job/96964728686, https://github.com/canonical/microceph/actions/runs/32802765072/job/97667634391, https://github.com/canonical/microceph/actions/runs/33461494602/job/99712592739
- main reef (Python harness): https://github.com/canonical/microceph/actions/runs/33287574075/job/99193448704
[...]

Issue #773, opened 2026-06-25
Reduce Flaky Tests


johnramsden:
# Master Issue for flaky test tracking

Flaky test reduction issues should tag fixes: #773, then re-open it.

PR #872, opened 2026-09-18
ci: require two consecutive OSD-up polls in the upgrade tests

johnramsden:
# Description

Part of #839
Part of #773

The upgrade readiness poll broke on the first `Found 3 osd up` reading and then re-issued the same query once more with no retry; a transient dip during the rolling restart (typically `Expected 3 OSD up, got 2`) failed the job on a single sample, and the poll's budget was never used.

Now require the expected count on two consecutive polls before trusting it

Because the new gate spends one poll of the original budget on confirmation, the Python harness's `attempts` goes from 36 to 37, and the shell loop's `seq 1 30` becomes `seq 1 31`, so a first success on what used to be the very last poll still passes.

The `actionutils.sh` `upgrade_multinode` copy has no caller in any workflow on the `main` branch today; it is changed only to stay in step with the `squid` copy, which does run there.

## Type of change

## How has this been tested?

- `pytest tests/robot/resources/test_harness_helpers.py`: four tests for the consecutive-poll helper `_advance_consecutive`, including one that drives the sequence `[True, False, True, True]` through it to prove a dip resets the gate.
- Robot dry-run (`--dryrun`) over `tests/robot/multi-node-tests` (9/9 passed) and `tests/robot/upgrade-squid-tests` (6/6 passed).
- Locally on nested LXD

## Contributor checklist


lmlg (approved):
lgtm

Issue #849, opened 2026-09-16
RBD pool disable rejected with "mirror peers still registered" immediately after peer removal


johnramsden:
# Issue report

Title: RBD pool disable rejected with "mirror peers still registered" immediately after peer removal

- Affects: main (observed); the code path is identical on squid
- Part of: https://github.com/canonical/microceph/issues/773
- Related: https://github.com/canonical/microceph/issues/729 (read-then-act without interlock; parent https://github.com/canonical/microceph/issues/726),
- Jobs: Test MicroCeph RBD Remote Replication features 5/37 (1 failure)

## What version of MicroCeph are you using ?

CI build from `main` at `de758ee` (2026-09-07). `rbd_mirror.go` is unchanged at `86c3250`.

## What are the steps to reproduce this issue ?

1. Two sites configured as mirror peers with two mirrored pools; disable the images, then disable `pool_two`, then `pool_one`.
2. `DisablePoolMirroring` calls `RemovePeer` and immediately runs the local `rbd mirror pool disable`.

## What happens (observed behaviour) ?

`pool_two` disables cleanly; `pool_one` fails with `rbd mirror pool disable pool_one: exit status 16 ... mirror peers still registered` although its images were already disabled. `getPeerUUID` returns only the first peer whose `RemoteName` matches and `RemovePeer` removes one id per site with no post-removal verification; whether a duplicate entry or an eventual-consistency gap caused it is not established.

## What were you expecting to happen ?

Peer removal is verified on both sites before the local pool disable is attempted, and the disable succeeds.

## Relevant logs, error output, etc.

- https://github.com/canonical/microceph/actions/runs/34075768213/job/101601800650 (09-07): job log lines 1666-1674, daemon journal lines 2274-2279
- Code at `86c3250`: `microceph/ceph/rbd_mirror.go` `getPeerUUID` (275-289), `RemovePeer` (292-319), `DisablePoolMirroring` (198-228; `RemovePeer` at 200, local disable at 207, remote retry at 213-221)

## Additional comments.

Proposed fix: after `RemovePeer`, re-fetch `GetRbdMirrorPoolInfo` on both sites and wait (bounded, same retry idiom as the remote disable) until no peer with that `RemoteName` remains, removing any second match, before calling the local pool disable. Narrow stopgap ahead of the #727/#729 redesign; add this evidence to #729.

Issue #837, opened 2026-09-16
RBD replication pool disable is killed by client and harness timeouts shorter than the daemon's retry budget


johnramsden:
# Issue report

Title: RBD replication pool disable is killed by client and harness timeouts shorter than the daemon's retry budget

- Affects: main, squid
- Part of: https://github.com/canonical/microceph/issues/773
- Related: https://github.com/canonical/microceph/pull/468 (added the server retry), https://github.com/canonical/microceph/pull/456 (added the 120-second client timeout)
- Jobs: main Test MicroCeph RBD Remote Replication features 5/37 (2 of the failures), squid same job 1/30

## What version of MicroCeph are you using ?

CI builds from `main` (`263f881`, `2ff1e99`) and `squid` (`897bcdd9`). `rbd_mirror.go`, `client/replication.go` and the Robot suite are unchanged at both tips.

## What are the steps to reproduce this issue ?

1. Configure two sites as RBD mirror peers, enable pool and image mirroring, disable the images.
2. Run `sudo microceph replication disable rbd pool_two` while the remote cluster has not yet observed the state change, so the remote pool-disable step retries.

## What happens (observed behaviour) ?

`DisablePoolMirroring` wraps the remote disable in `retry.Retry` with `strategy.Limit(10)` and `backoff.Linear(5s)`, a worst case of about 225 seconds of sleep. The Robot keyword kills the CLI at 60 seconds (main: rc 124, empty output), and the CLI's request context in `client/replication.go` expires at 120 seconds (squid: `context deadline exceeded` at 120.3 seconds). A slow but healthy disable is reported as a failure and the following remote removal is refused. `strategy.Delay(5)` in the same call is 5 nanoseconds.

## What were you expecting to happen ?

The client and the harness wait at least as long as the daemon may legitimately take, and the disable completes.

## Relevant logs, error output, etc.

- main: https://github.com/canonical/microceph/actions/runs/31765209153/job/94659712228 (08-14), https://github.com/canonical/microceph/actions/runs/32091780936/job/95575430244 (08-18)
- squid: https://github.com/canonical/microceph/actions/runs/33142327201/job/98755808713 (08-28)
[...]

PR #878, opened 2026-09-22
fix: make the rbd pool disable survive peer re-registration and slow retries

johnramsden:
# Description

Part of #849
Follow-up to https://github.com/canonical/microceph/issues/837
Part of #773

`microceph replication disable rbd <pool>` could fail with `mirror peers still registered` right after MicroCeph had removed the peers. The other site's rbd-mirror daemon pings the pool every 30 seconds and adds the peer back if it is missing, so a plain retry would fail the same way.

The pool disable now retries and removes any peer that came back before each attempt. The same branch fixes the product half of issue 837: the CLI timeout for replication requests was shorter than the time the daemon can spend retrying, and `strategy.Delay(5)` slept 5 nanoseconds instead of 5 seconds. One commit per issue.

Issue 849 also has a squid backport, so it stays open until both merge.

## Type of change

## How has this been tested?

- Unit tests for the retry and the timeouts.
- Reproduced locally on two LXD clusters with mirroring enabled; the failure disappears with the fix.
- Full suites run locally on nested LXD.

## Contributor checklist

Please check that you have:


UtkarshBhatthere (approved):
lgtm, thanks @johnramsden

